PR #15, opened 2021-09-26
Initial Rosaria implementation

Tsym-or-Tysm:
This is my first pass at the Rosaria implementation. I checked the implementation in sim and the following appear to work: Rosaria A2, A4, C2-C6.

The only one that I didn't check is C1 because it has an attack speed buff, and I see comments scattered throughout that indicate it's not worked out yet.

I also compared a few of the sim numbers versus the basic Rosaria sim made by Sitri [here](https://docs.google.com/spreadsheets/d/1jEuYsnexNbvEJ68Cr8-GuHA_MNqq7iZmS8qTBGpHvNg), and those all check out so there aren't any crazy mistakes that I can see.

I tried to keep the coding style consistent with the rest of the sim, but happy to accommodate any changes.


genshinsim:
should change burst to clone the snapshots:

```go
	c.AddTask(func() {
		dot = c.Snapshot(
			"Rites of Termination (DoT)",
			core.AttackTagElementalBurst,
			core.ICDTagNone,
			core.ICDGroupDefault,
			core.StrikeTypeDefault,
			core.Cryo,
			25,
			burstDot[c.TalentLvlBurst()],
		)
		dot.Targets = core.TargetAll
		c.applyC6(&dot)
		
		// dot every 2 second after lance lands
		for i := 120; i < dur; i += 120 {
			x := dot.Clone()
			c.QueueDmg(&x, 10+i)
		}
	}, "rosaria-snapshot", f-10)

	c.Core.Status.AddStatus("rosariaburst", dur)```
	
Also consider using the c.Stats function instead of accessing the stat directly from the snapshot on the crit share `crit_share := 0.15 * hit1.Stats[core.CR]`

Tsym-or-Tysm:
Made requested changes - confirmed that the C6 callback is still getting copied and applied correctly, and I also added a TODO note for optimizing her C4 as we discussed, regarding changing it to a subscription generated from her skill ability.

PR #24, opened 2021-09-29
Add instructor artifact set

Tsym-or-Tysm:


internal/artifacts/instructor/instructor.go
```diff
@@ -0,0 +1,60 @@
...
+			if s.ActiveChar != c.CharIndex() {
+				return false
+			}
+			// Source of elemental reaction must be the character with instructor
+			if ds.ActorIndex != c.CharIndex() {
+				return false
+			}
+
+			// Add 120 EM to all characters except the one with instructor
+			m := make([]float64, core.EndStatType)
```
genshinsim, line 40: these next 2 lines should be moved to outside the call back to above line 28 so that array is only initialized once. there's no problem sharing the same array across all instances since it's read only. saves some memory allocations

PR #26, opened 2021-10-02
Initial Xiao implementation

Tsym-or-Tysm:
Initial Xiao implementation.

Checked damage versus Xiao mains calculator and things look about right for some individual hits.


genshinsim (approved):

internal/characters/xiao/xiao.go
```diff
@@ -0,0 +1,128 @@
...
+// Xiao specific character implementation
+type char struct {
+	*character.Tmpl
+	eCharge      int
+	eChargeMax   int
+	eNextRecover int
+	eTickSrc     int
+	qStarted     int
+	a4Expiry     int
+	c6Check      []interface{} // int (source frame), string (snapshot ability name)
```
genshinsim, line 21: Shouldn't use interface here as that's actually slower. This should just be

`c6Src int`

internal/characters/xiao/xiao.go
```diff
@@ -0,0 +1,128 @@
...
+// Implements Xiao C6:
+// While under the effect of Bane of All Evil, hitting at least 2 opponents with Xiao's Plunge Attack will immediately grant him 1 charge of Lemniscatic Wind Cycling, and for the next 1s, he may use Lemniscatic Wind Cycling while ignoring its CD.
+// Adds an OnDamage event checker - if we record two or more instances of plunge damage, then activate C6
+func (c *char) c6() {
+	c.Core.Events.Subscribe(core.OnDamage, func(args ...interface{}) bool {
+		ds := args[1].(*core.Snapshot)
+		if ds.ActorIndex != c.Index {
+			return false
+		}
+		if !((ds.Abil == "High Plunge") || (ds.Abil == "Low Plunge")) {
```
genshinsim, line 86: Here can check for tag instead

```go
if ds.AttackTag != core.AttackTagPlunge {
   return false
}
if c.Core.Status.Duration("xiaoburst") == 0 {
    return false
}
```

since actor is already xiao and this is plunge then it should be fine. should also check if his burst is activated on the next line though?

internal/characters/xiao/xiao.go
genshinsim: also need a counter here
`c6Count int`

internal/characters/xiao/xiao.go
```diff
@@ -0,0 +1,128 @@
...
+			return false
+		}
+		// Prevents activation more than once per plunge damage instance
+		if (c.c6Activated == 1) && (c.c6Check[0].(int) == ds.SourceFrame) {
+			return false
+		}
+		if (c.c6Check[0].(int) == ds.SourceFrame) && (c.c6Check[1].(string) == ds.Abil) {
```
genshinsim, line 93: Here we can just check for the source frame

```go
if c.c6Src != ds.SourceFrame {
   c.c6Src = ds.SourceFrame
   c.c6Count = 1
   return false
}

//otherwise increment the counter
c.c6Count++

//since it'll only = 2 once, this should only activate once
if c.c6Count == 2 {
   //do activation stuff here
}

```

internal/characters/xiao/xiao.go
Tsym-or-Tysm: Can't use the plunge tag since you don't want the normal plunging hits to be counted towards the proc, as otherwise you would be able to proc C6 Xiao on single entities which we know isn't the case.

PR #34, opened 2021-10-17
Add crystallize to list of reactions that Fischl A4 work on

Tsym-or-Tysm:


genshinsim:
just and update on this one. intending on merge/fix this one after i refactor the reaction system. really the OnReaction hook needs to be redesigned so that it's called back at a more sensible point.

Tsym-or-Tysm:
Subsumed by #51 - Closed.

PR #38, opened 2021-10-18
Add Tartaglia

Tsym-or-Tysm:
Here are is my initial Tartaglia implementation. It's missing a few things, with the biggest being C4. Not sure what's the best way to integrate this with Vigne's changes - maybe we can take the best ideas from each and integrate?


Tsym-or-Tysm:
Elements combined with Vigne's merged Childe patch and merged into master.

PR #43, opened 2021-10-24
Add Childe

123prob7:
My childe implementation.


internal/characters/childe/childe.go
```diff
@@ -0,0 +1,143 @@
...
+	c.Core.Events.Subscribe(core.OnCharacterSwap, func(args ...interface{}) bool {
+		prev := args[0].(int)
+		if prev != c.CharIndex() {
+			return false
+		}
+
+		if c.Core.Status.Duration("childemelee") > 0 {
+			c.Core.Status.DeleteStatus("childemelee")
+			//TODO: preemptive cd doesnt seem fixed at 6s
+			newCD := float64(c.Core.F - c.eCast + 6*60)
```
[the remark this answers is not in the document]
Tsym-or-Tysm: The CD implementation in my version should be correct - right up above the linked location there's some documentation from messages in discord if you want to use those as a reference.

internal/characters/childe/stats.go
```diff
@@ -1 +1,205 @@
 package childe
```
Tsym-or-Tysm, line 1: Should probably format these?

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+								core.StrikeTypeDefault,
+								core.Hydro,
+								25,
+								rtSlash[c.TalentLvlSkill()],
+							)
+							d.Targets = core.TargetAll
+
+							c.Core.Combat.ApplyDamage(&d)
+							c.Core.Log.Debugw("Riptide Slash ticked", "frame", c.Core.F, "event", core.LogCharacterEvent, "dur",
+								c.Core.Status.Duration("childemelee"), "target", t.Index(), "fl", c.rtflashICD[t.Index()], "sl", c.rtslashICD[t.Index()], "rtExpiry", c.rtExpiry[t.Index()])
```
Tsym-or-Tysm, line 294: Why is the flash riptide ICD in this debug message? Also maybe we can rename "sl" to use the full name so it's more obvious to the end user what it means.

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+					}
+				}
+
+				// A4:Sword of TorrentsWhen Tartaglia is in Foul Legacy: Raging Tide’s Melee Stance,
+				// on dealing a CRIT hit, Normal and Charged Attacks apply the Riptide status effect to opponents.
+				if crit {
+					c.applyRT(t.Index())
+				}
+			}
+		case core.AttackTagExtra:
```
Tsym-or-Tysm, line 306: Maybe I'm missing something, but isn't this mostly just a repeat of the NormalAttack case? Can't we just have the above case match two values and mostly drop this? After moving the aim mode stuff up?

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+				}
+
+				// A4:Sword of TorrentsWhen Tartaglia is in Foul Legacy: Raging Tide’s Melee Stance,
+				// on dealing a CRIT hit, Normal and Charged Attacks apply the Riptide status effect to opponents.
+				if crit {
+					c.applyRT(t.Index())
+				}
+			} else {
+				// aim mode
+				if c.rtExpiry[t.Index()] > c.Core.F {
```
Tsym-or-Tysm, line 341: Same comments as Riptide Slash case.

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+								c.Core.Status.Duration("childemelee"), "target", t.Index(), "fl", c.rtflashICD[t.Index()], "sl", c.rtslashICD[t.Index()], "rtExpiry", c.rtExpiry[t.Index()])
+						}, "Riptide Flash", 5)
+						c.rtflashICD[t.Index()] = c.Core.F + 42 //0.7s icd
+					}
+				}
+
+				c.applyRT(t.Index())
+			}
+		case core.AttackTagElementalBurst:
+			if c.Core.Status.Duration("childemelee") > 0 {
```
Tsym-or-Tysm, line 373: Similar comments as Riptide Slash

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+		}
+		return false
+	}, "childe-riptide")
+}
+
+//Q: does all type of childe's riptide share the same icd of particle?
+//A: Hard 3s cooldown on particle generation. Particles can only be gained through riptide procs.
+//Q: do all riptide flash, burst, slash and blast share same icd of particle generation?
+//A: flash and slash share it iirc, burst doesn't have one, and I don't think blast gens energy
+func (c *char) rtParticleGen() {
```
Tsym-or-Tysm, line 420: Maybe I'm missing something, but I'm not sure why this function needs to needs to be a separate event subscription. Can't you just handle riptide energy procs after each riptide instance in the main riptide function or something?

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+	c.Core.Events.Subscribe(core.OnDamage, func(args ...interface{}) bool {
+		ds := args[1].(*core.Snapshot)
+		if ds.ActorIndex != c.CharIndex() {
+			return false
+		}
+		if ds.Abil != "Riptide Flash" && ds.Abil != "Riptide Slash" &&
+			ds.Abil != "C4 Riptide Flash" && ds.Abil != "C4 Riptide Slash" {
+			return false
+		}
+		if c.rtParticleICD > c.Core.F {
```
Tsym-or-Tysm, line 430: I think you're missing the equality condition between these two.

internal/characters/childe/abil.go
```diff
@@ -0,0 +1,513 @@
...
+		if c.rtParticleICD < c.Core.F {
+			c.Core.Log.Debugw("childe gen a hydro particle", "frame", c.Core.F, "event", core.LogCharacterEvent, "icd", c.rtParticleICD)
+			c.rtParticleICD = c.Core.F + 180 //3 sec
+			c.QueueParticle("tartaglia", 1, core.Hydro, 100)
+		}
+		return false
+	}, "childe-particle-gen")
+}
+
+func (c *char) c4TickFunc(t int) func() {
```
Tsym-or-Tysm, line 443: To be honest I have no idea how his C4 actually works... I'll have to review this later when I find time to look into it.

internal/characters/childe/childe.go
```diff
@@ -0,0 +1,143 @@
...
+	c.rtflashICD = make([]int, len(c.Core.Targets))
+	c.rtslashICD = make([]int, len(c.Core.Targets))
+	c.rtExpiry = make([]int, len(c.Core.Targets))
+
+	if c.Base.Cons >= 4 {
+		c.funcC4 = make([]bool, len(c.Core.Targets))
+	}
+}
+
+func (c *char) ActionStam(a core.ActionType, p map[string]int) float64 {
```
Tsym-or-Tysm, line 71: I think this needs a dash stamina case too

Tsym-or-Tysm (commented):
Overall the implementation looks solid - certainly catches a lot of edge cases better than my implementation!

I think the only big global comment that I have is that we should consider calling everything "Tartaglia" instead of "Childe", or at least all of the user facing elements (Status and debug messages). Tartaglia is his official name in game and is what the library uses, and you've also registered him under that name.

123prob7:
Have made changes through those comments. Some big one can be addressed is ditching riptide hook up and using the callback on each targets hit like Tsym's way. I removed c4 also. Since it got issues after tartaglia executed riptide blast in order to clear riptide status. Not sure how it works either. So its better to ditch it

internal/characters/tartaglia/abil.go
```diff
@@ -0,0 +1,411 @@
...
+				fmt.Sprintf("Normal %v", c.NormalCounter),
+				core.AttackTagNormal,
+				core.ICDTagNormalAttack,
+				core.ICDGroupDefault,
+				core.StrikeTypeSlash,
+				core.Hydro,
+				25,
+				mult[c.TalentLvlSkill()],
+			)
+			d.OnHitCallback = c.rtSlashCallback
```
Tsym-or-Tysm, line 66: This is going to seem somewhat cursed, but I realized in game that riptide applications from melee attacks always apply first, so if you apply riptide with a melee crit, then the riptide damage procs immediately. Should check that happens here.

Note that it doesn't work this way for CAs - if you apply riptide from CAs, then Riptide flash does not immediately proc.

internal/characters/tartaglia/abil.go
```diff
@@ -0,0 +1,411 @@
...
+		if hitWeakPoint != 0 {
+			d.HitWeakPoint = true
+		}
+		d.Targets = core.TargetAll
+		d.OnHitCallback = c.rtSlashCallback
+		c.QueueDmg(&d, f-meleeChargeDelayOffset[i])
+		// c.AddTask(func() {
+		// 	c.Core.Combat.ApplyDamage(&d)
+		// }, "tartaglia-charge-attack", f-meleeChargeDelayOffset[i])
+	}
```
Tsym-or-Tysm, line 150: This is a dynamic ability, so you need to use either QueueDmgDynamic, or AddTask to align the snapshot creation and damage instance.

internal/characters/tartaglia/abil.go
```diff
@@ -0,0 +1,411 @@
...
+			50,
+			mult,
+		)
+		d.Targets = core.TargetAll
+		if c.Core.Status.Duration("tartagliamelee") > 0 {
+			d.OnHitCallback = c.rtBlastCallback
+		}
+		c.Core.Combat.ApplyDamage(&d)
+		if c.Core.Status.Duration("tartagliamelee") > 0 {
+			if c.Base.Cons >= 4 {
```
Tsym-or-Tysm, line 255: Isn't this C6 not C4?

internal/characters/tartaglia/abil.go
```diff
@@ -0,0 +1,411 @@
...
+				}
+				//aim mode
+				//dont log if it just refreshes riptide status
+				if c.rtExpiry[t.Index()] <= c.Core.F {
+					c.Core.Log.Debugw("Tartaglia applied riptide", "frame", c.Core.F, "event", core.LogCharacterEvent, "target", t.Index(), "rtExpiry", c.Core.F+rtA1)
+				}
+
+				c.rtExpiry[t.Index()] = c.Core.F + rtA1
+			}
+		}
```
Tsym-or-Tysm, line 396: I think the if logic is somewhat confusingly laid out because you need to repeat the "AttackTag" if condition in the ranged section. I think it'd be clearer if you moved the melee vs. ranged logic as the top layer, then the attack tag logic can be the inner layer.

internal/characters/tartaglia/abil.go
```diff
@@ -0,0 +1,411 @@
...
+
+		if ds.AttackTag == core.AttackTagElementalBurst {
+			if c.Core.Status.Duration("tartagliamelee") == 0 {
```
Tsym-or-Tysm, line 399: Personal preference, but I don't like nesting ifs when it's not needed. Can we change this to an && condition? If the long line bothers you then just have the condition spread across 2 lines.

Tsym-or-Tysm (commented):
Some initial comments here - I'll do some more testing with this implementation later.

internal/characters/tartaglia/tartaglia.go
```diff
@@ -0,0 +1,114 @@
...
+func (c *char) onExitField() {
+	c.Core.Events.Subscribe(core.OnCharacterSwap, func(args ...interface{}) bool {
+		if c.Core.Status.Duration("tartagliamelee") > 0 {
+			c.onExitMeleeStance()
+		}
+		return false
+	}, "tartaglia-exit")
+}
+
+func (c *char) onDefeatTargets() {
```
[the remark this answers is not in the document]
123prob7: hmm I don't know what you meant. Could you elaborate?

internal/characters/tartaglia/tartaglia.go
```diff
@@ -0,0 +1,114 @@
...
+		c.mlBurstUsed = false
+		c.funcC4 = make([]bool, len(c.Core.Targets))
+	}
+
+	c.rtParticleICD = 0
+	c.rtFlashICD = make([]int, len(c.Core.Targets))
+	c.rtSlashICD = make([]int, len(c.Core.Targets))
+	c.rtExpiry = make([]int, len(c.Core.Targets))
+
+	c.Core.Flags.ChildeActive = true
```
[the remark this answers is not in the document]
123prob7: isn't this flag supposed to be for his A1 +1 level for AA every units in the comp?

internal/characters/tartaglia/tartaglia.go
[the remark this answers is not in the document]
Tsym-or-Tysm: Just like comments for the private functions. Like it would just be something like

`// Handles Childe riptide burst and C2 on death effects`

internal/characters/tartaglia/tartaglia.go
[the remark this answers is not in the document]
Tsym-or-Tysm: Ah you're right - forgot that the code handling this was already in the sim.

internal/characters/tartaglia/abil.go
123prob7: Cursed indeed. I didn't find a right way to get it done with current riptide slash callback. So I will let how it is rn

PR #59, opened 2021-11-07
Add temporary option for Guoba swirl onto Sucrose E

Tsym-or-Tysm:


Tsym-or-Tysm:
Un-needed with the refactor.

srliao:
implementation added in #122

PR #72, opened 2021-11-24
Add Husk of Opulent Dreams and Cinnabar Spindle

Tsym-or-Tysm:


internal/artifacts/huskofopulentdreams/huskofopulentdreams.go
```diff
@@ -0,0 +1,128 @@
...
+
+				lastStackGain = s.F
+
+				c.AddTask(gainStackOfffield(src), "husk-4pc-off-field-gain", 180)
+				c.AddTask(checkStackLoss, "husk-4pc-stack-loss-check", 360)
+			}
+		}
+
+		// Initiate off-field stacking if off-field at start of the sim
+		if s.ActiveChar != c.CharIndex() {
```
srliao, line 73: should be in an initialize event similar to this: https://github.com/genshinsim/gcsim/blob/eba053dec6d55ce604f12f9b0c0061f0f552d2b9/internal/weapons/claymore/spine/spine.go#L19

PR #79, opened 2021-12-01
Add qiqi

Tsym-or-Tysm:
- Change incoming healing bonus to be healed character specific
- Add Qiqi


internal/characters/qiqi/abil.go
```diff
@@ -0,0 +1,219 @@
...
+		if c.Core.Status.Duration("qiqiskill") == 0 {
+			return
+		}
+
+		// TODO: Not sure how this interacts with sac sword... Treat it as only one instance can be up at a time for now
+		if c.skillLastUsed > src {
+			return
+		}
+
+		// Create a temporary snapshot to get the atk mods included
```
srliao, line 170: Is there a reason why you didn't use this function to get ATKP and ATK instead of getting a full snapshot? https://github.com/genshinsim/gcsim/blob/0a1394e66d5b675ad6ba6ebbbe0dacf2612c87df/pkg/character/stats.go#L28

internal/characters/qiqi/qiqi.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	c.Tmpl = t
+	c.Energy = 80
+	c.EnergyMax = 80
+	c.Weapon.Class = core.WeaponClassSword
+	c.NormalHitNum = 5
+	c.BurstCon = 3
+	c.SkillCon = 5
+
+	c.talismanExpiry = make([]int, len(c.Core.Targets))
```
srliao, line 36: This should be in the character's init call similar to this:
https://github.com/genshinsim/gcsim/blob/0a1394e66d5b675ad6ba6ebbbe0dacf2612c87df/internal/characters/raiden/raiden.go#L48

make sure to call the template's init function

this is just in case for whatever reason the targets have not yet been initialized

internal/characters/qiqi/qiqi.go
```diff
@@ -0,0 +1,176 @@
...
+
+		if c.talismanExpiry[t.Index()] < c.Core.F {
+			return false
+		}
+		if c.talismanICDExpiry[t.Index()] >= c.Core.F {
+			return false
+		}
+
+		// Talisman healing is dynamic, so need to create a new snapshot instance for it
+		dsHeal := c.Snapshot(
```
srliao, line 60: see prev comment re using .Stat function

internal/characters/qiqi/qiqi.go
```diff
@@ -0,0 +1,176 @@
...
+		c.talismanICDExpiry[t.Index()] = c.Core.F + 60
+
+		c.Core.Log.Debugw("Qiqi Talisman Healing", "frame", c.Core.F, "event", core.LogCharacterEvent, "char", c.Index, "target", t.Index(), "healed_char", ds.ActorIndex, "talisman_expiry", c.talismanExpiry[t.Index()], "talisman_healing_icd", c.talismanICDExpiry[t.Index()], "healed_amt", healAmt)
+
+		return false
+	}, "talisman-heal-hook")
+}
+
+// Handles C2, A4, and skill NA/CA on hit hooks
+// Additionally handles burst Talisman hook - can't be done another way since Talisman is applied before the burst damage is dealt
```
srliao, line 81: rather than queueing damage in burst, why not just add a task that takes a snapshot, applies talisman, then call ApplyDamage? wouldn't that achieve the goal of applying talisman first before triggering damage?

Tsym-or-Tysm: Talisman is supposed to be applied on hit. I'm not a fan of manually applying it in the burst since that's more liable to break depending on how things interact in the future.

internal/characters/qiqi/qiqi.go
Tsym-or-Tysm: Wait does this need to be fixed in the Tartaglia code too then?

https://github.com/genshinsim/gcsim/blob/0a1394e66d5b675ad6ba6ebbbe0dacf2612c87df/internal/characters/tartaglia/tartaglia.go#L45

srliao: > 

ya.. i thought i commented on that one before

Tsym-or-Tysm: Well it got merged as-is so I used that as a reference haha. I'll change this in a bugfix PR.

PR #80, opened 2021-12-02
Implement Ocean-Hued Clam

Tsym-or-Tysm:


pkg/core/core.go
```diff
@@ -20,9 +20,10 @@ const (
 )
 
 type Flags struct {
-	DamageMode   bool
-	LogDebug     bool
-	ChildeActive bool
+	DamageMode    bool
+	LogDebug      bool // Used to determine logging level
+	ChildeActive  bool // Used for Childe +1 NA talent passive
+	OHCActiveChar int  // Shows which character currently has an active OHC proc. -1 = Non-active
```
srliao, line 26: Should ideally use custom flags for this instead of adding to the flags struct: https://github.com/genshinsim/gcsim/blob/0a1394e66d5b675ad6ba6ebbbe0dacf2612c87df/pkg/core/core.go#L221

Arguably I should have done the same for childe flag though

pkg/core/snapshot.go
```diff
@@ -45,6 +45,9 @@ type Snapshot struct {
 	ReactMult        float64 //reaction multiplier for melt/vape
 	ReactBonus       float64 //reaction bonus %+ such as witch; should be 0 and only affected by hooks
 
+	//Special OHC flag
+	IsOHCDamage bool
```
srliao, line 49: this is ok as a work around for now but with the reaction refactor there won't be a different apply damage function for reaction damage vs regular damage. i'll have to change this to be handled in the snapshot queued itself

srliao: on that note, snapshot will need an IgnoreDefense flag for the purpose of.. well.. ignoring defense for reaction damage

Tsym-or-Tysm: Yeah figured this would change with the refactor - not a big deal.

pkg/core/core.go
Tsym-or-Tysm: I can make the Childe change in another bugfix PR

PR #85, opened 2021-12-04
Fix geo resonance

123prob7:
Fixed geo resonance

- Buffs only affect on-field character
- 15% dmg bonus can't be snapshotted

### DIFF
--- a/resonance.go
+++ b/resonance.go
@@ -69,7 +69,8 @@ func (s *Simulation) initResonance(count map[core.EleType]int) {
 				})
 				s.C.Events.Subscribe(core.OnDamage, func(args ...interface{}) bool {
 					t := args[0].(core.Target)
-					if s.C.Shields.IsShielded() {
+					ds := args[1].(*core.Snapshot)
+					if s.C.Shields.IsShielded() && s.C.ActiveChar == ds.ActorIndex {
 						t.AddResMod("geo res", core.ResistMod{
 							Duration: 15 * 60,
 							Ele:      core.Geo,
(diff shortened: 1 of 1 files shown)


Tsym-or-Tysm:
The change so that geo resonance can't be snapshotted looks good.

For the changes so that buffs only affect the on-field character, I wonder if the better way to handle this is to change the IsShielded instead? I'm pretty sure that shields only work for the active character, so we may want to do something like this instead so we have a consistent interface, and we can fix all of the other instances where IsShielded occurs in the code base at the same time.

```go
func (s *ShieldCtrl) IsShielded(char int) bool { return len(s.shields) > 0 && char == s.C.ActiveChar }
```
Thoughts @srliao?

srliao:
> The change so that geo resonance can't be snapshotted looks good.
> 
> For the changes so that buffs only affect the on-field character, I wonder if the better way to handle this is to change the IsShielded instead? I'm pretty sure that shields only work for the active character, so we may want to do something like this instead so we have a consistent interface, and we can fix all of the other instances where IsShielded occurs in the code base at the same time.
> 
> ```go
> func (s *ShieldCtrl) IsShielded(char int) bool { return len(s.shields) > 0 && char == s.C.ActiveChar }
> ```
> 
> Thoughts @srliao?

Hmm how would this work for something like memory of dust which adds attack if shielded? I actually don't know if something like that works off field?

srliao:
I'm going to merge this for now and add a new issue to refactor IsShielded

PR #99, opened 2021-12-11
Energy calc mode

Kisa75:
This PR implements a new Run Option "Energy calc mode"

- Characters can use their burst regardless of their current energy
- When using a burst action, the current energy gets logged before it gets reset to 0
- At the end of the simulation, the logged amounts of energy will be used to calculate how much ER would have been required
- The calculated recommended ER per character will be printed in the Text Summary
- The Energy calc mode can be activated via the cmd line option "ercalc"


run.go
```diff
@@ -96,12 +96,18 @@ func (s *Simulation) AdvanceFrame() error {
...
 		//check if the current action is executable right now; if not then delay
-		delay := s.C.AnimationCancelDelay(s.queue[0].Typ)
+		actionType := s.queue[0].Typ
+		delay := s.C.AnimationCancelDelay(actionType)
 		if delay > 0 {
 			s.skip = delay
 			return nil
 		}
 
+		if s.C.Flags.EnergyCalcMode && actionType == core.ActionBurst {
```
srliao, line 106: Instead of modifying the simulation run function here, it's better to hook on to burst events. Something like the following code would work:

```go
		s.C.Events.Subscribe(core.PreBurst, func(args ...interface{}) bool {
			activeChar := s.C.Chars[s.C.ActiveChar]
			//log your stats here
		}, "reaction-log")
```

It's preferable not to modify the main code here and instead rely on adding in subscription to events if energy mode is enabled.

worker.go
```diff
@@ -16,6 +17,7 @@ import (
 
 	"github.com/genshinsim/gcsim/pkg/core"
 	"github.com/genshinsim/gcsim/pkg/parse"
+	_ "github.com/montanaflynn/stats"
```
srliao, line 20: this blank import shouldn't be needed

cmd/gcsim/main.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/genshinsim/gcsim/pkg/calcqueue"
 	"github.com/genshinsim/gcsim/pkg/core"
 	"github.com/genshinsim/gcsim/pkg/parse"
+	_ "github.com/montanaflynn/stats"
```
srliao, line 22: not sure this is needed here? stats should be calculated in the sim anyways not in the cmd line?

worker.go
```diff
@@ -462,6 +466,38 @@ func CollectResult(data []Stats, mode bool, chars []string, detailed bool) (resu
 		result.DPSByTarget[j] = dpsTargetRollup
 	}
 
+	// required ER
+
+	if erCalcMode {
```
srliao, line 471: So I was suggesting for this part, we do something like this:

```go
if erCalcMode {
    energyCount := make([][]float64, charCount)

    for i := 0; i < charCount; i++ {
        energyCount[i] = make([]float64, charMaxEnergy) //need to prob add a database for this or soemthing so we know what the max is?
    }

    for _, d := range data {
        for j, charEnergy := range d.EnergyWhenBurst {
            for _, v := range charEnergy {
                energyCount[j][v]++
            } 

        }
    }

    //so now you have basically distribution of various energy and their frequency. I would use this then to calculate
    //the min/max/std dev of raw energy received

}
```

benefit of this approach is now we are able to summarize what's the min/mean/max/std. dev. of raw energy received. it gives us a better picture of how volatile the energy generation of this team is

and then we can calculate the min and average ER threshhold.

worker.go
```diff
@@ -592,6 +628,18 @@ func (r *Result) PrettyPrint() string {
 			}
 		}
 	}
+
+	// Recommended ER, only in ER calc mode
+	if r.RequiredER[0] != 0 {
```
srliao, line 633: not sure why here you are checking if the first element does not equal 0?

shouldn't you be checking if ER mode?

worker.go
Kisa75: Removed the import

cmd/gcsim/main.go
Kisa75: Removed the import

worker.go
Kisa75: Fixed

run.go
Kisa75: Resolved

worker.go
Kisa75: Resolved, additional representation of the data will be added later on

PR #101, opened 2021-12-12
Barbaradev

HanSeoulOh:
Talents complete
Charged Attack complete
Constellations Complete
Ascensions Essentially complete
Elemental Skill Complete enough.. may be missing some weird mechanics, but is pretty complete I think. Doesn't have on particle gain duration extension yet..


HanSeoulOh:
Think this PR is ready for initial review if @srliao or @Tsym-or-Tysm wants to go over it..

PR #109, opened 2021-12-23
barbara implementation

HanSeoulOh:


internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+func (c *char) Skill(p map[string]int) (int, int) {
+
+	f, a := c.ActionFrames(core.ActionSkill, p)
+
+	//add field effect timer
+	//assumes a4
+	c.Core.Status.AddStatus("barbskill", 15*60)
+	//hook for buffs; active right away after cast
+
+	c.stacks = 0
```
srliao, line 108: what's this used for?

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+	heal := (bursthp[c.TalentLvlBurst()] + bursthpp[c.TalentLvlBurst()]*c.MaxHP()) * (1 + hpplus)
+	c.Core.Health.HealAll(c.Index, heal)
+
+	c.Energy = 0
+	c.SetCD(core.ActionBurst, 20*60)
+	return f, a //todo fix field cast time
+}
+
+//inspired from raiden
+func (c *char) onSkillStackCount(stats [core.EndStatType]float64, skillInitF int) {
```
srliao, line 179: this function doesn't need stats does it?

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+	} else {
+		c.SetCD(core.ActionSkill, 32*60)
+	}
+	return f, a //todo fix field cast time
+}
+
+func (c *char) barbaraHealTick(healAmt float64, skillInitF int) func() {
+	return func() {
+		//make sure it's not overwritten
+		if c.skillInitF != skillInitF {
```
srliao, line 147: you have to set c.skillInitF = c.Core.F in the skill itself else this will always be false

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+			return
+		}
+		//do nothing if buff expired
+		if c.Core.Status.Duration("barbskill") == 0 {
+			return
+		}
+		c.Core.Log.Debugw("barbara field ticking", "frame", c.Core.F, "event", core.LogCharacterEvent)
+		c.Core.Health.HealActive(c.Index, healAmt)
+
+		// tick per second
```
srliao, line 157: is it tick every second or tick every 5 second? i don't know actually. need to check

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+		//check for healing
+		if c.Core.Status.Duration("barbskill") > 0 {
+			//heal target
+			heal := (prochpp[c.TalentLvlSkill()] + prochp[c.TalentLvlSkill()])
+			c.Core.Health.HealAll(c.Index, 4*heal)
+			done = true
+		}
+
+	}
+	var cbenergy func(a core.AttackCB) = nil
```
srliao, line 77: you don't need to set = nil. go by default initializes this as nil

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+	energyCount := 0
```
srliao, line 78: this should prob be moved into the if statement so there's no wasted mem alloc if cons < 4?

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+	done := false
+	// Taken from Noelle code
+	cb := func(a core.AttackCB) {
+		if done { //why do we need this @srl
+			return
+		}
+		//check for healing
+		if c.Core.Status.Duration("barbskill") > 0 {
+			//heal target
+			heal := (prochpp[c.TalentLvlSkill()] + prochp[c.TalentLvlSkill()])
```
srliao, line 71: i think you forgot to multiply this by her current hp? or is it max?

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
```
srliao, line 32: i think you forgot to multiply this by her current hp? or is it max?

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// TODO: Not sure of snapshot timing
+	c.Core.Combat.QueueAttack(ai, core.NewDefCircHit(2, false, core.TargettableEnemy), 0, f, cb, cbenergy)
+
+	return f, a
+}
+
+// barbara skill - copied from bennett burst
+
+func (c *char) Skill(p map[string]int) (int, int) {
```
srliao, line 99: so she should have another on collision mechanic with her E that applies hydro to targets it touches similar to xingqiu

unfortunately i don't have the exact details on that mechanics. for now it's ok to leave it out i guess till i have more time to research it

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,197 @@
...
+
+	c.stacks = 0
+	ai := core.AttackInfo{
+		ActorIndex: c.Index,
+		Abil:       "Let the Show Begin♪",
+		AttackTag:  core.AttackTagElementalArt,
+		ICDTag:     core.ICDTagNone,
+		ICDGroup:   core.ICDGroupDefault,
+		Element:    core.Hydro,
+		Durability: 25, //TODO: what is 1A GU?
```
srliao, line 116: 1A is 25 so this is fine

internal/characters/barbara/barbara.go
```diff
@@ -0,0 +1,118 @@
...
+		c.c2()
+	}
+	if c.Base.Cons >= 6 {
+		c.c6()
+	}
+	return &c, nil
+}
+
+func (c *char) a1() {
+	c.Core.AddStamMod(func(a core.ActionType) (float64, bool) { // @srl does this activate for the active char?
```
srliao, line 50: stamina is on a per player basis so it doesnt matter what char is active. so this is fine

internal/characters/barbara/barbara.go
```diff
@@ -0,0 +1,118 @@
...
+		if c.Core.Status.Duration("barbskill") >= 0 {
+			return -0.12, false
+		}
+		return 0, false
+	})
+}
+
+func (c *char) c2() {
```
srliao, line 58: for this take a look at how raiden's E is implemented. you want to add this mod onto every single character

also by design the Amount function returns 2 values. first is the stat slice that has the changes to stat to be added to the snapshot. second is a bool stating if the buff applies or not

so if the bool is false, the first slice is ignored all together. so it's fine to just have val[core.HydroP] = 0.15 and never change that. and when it doesnt apply just return nil, false

internal/characters/barbara/abil.go
HanSeoulOh: nothing lol

internal/characters/barbara/abil.go
HanSeoulOh: Accidentally deleted it in `Skill`, it's back now

internal/characters/barbara/abil.go
HanSeoulOh: it's every 5 seconds starting at 0, so for 15 seconds, it ticks 4 times

internal/characters/barbara/abil.go
HanSeoulOh: maxhp... fixed

internal/characters/barbara/barbara.go
HanSeoulOh: do you mean the burst? I think I changed it to what it is correctly but will need another review

internal/characters/barbara/abil.go
HanSeoulOh: huh in a previous iteration my linter was giving me a hard time about it.. removed it now I guess

internal/characters/barbara/abil.go
HanSeoulOh: from Barbara mains:

Finding:
Barbara's melody loop ticks every 1.5 seconds and follows the ~universal rule of being able to apply its element (hydro) every 2.5s per target, and also each third hit within those 2.5s windows. The extra two damaging droplet strikes that occur immediately after activation count towards those three hits for an early, extra application of hydro, after which it will only have the capability to apply hydro every 3 seconds (every other tick).

HanSeoulOh: It looks like 4 hits at the beginning to me with only 2 of the hits registering doing damage and applying Hydro. https://streamable.com/bbesgl ... going to need instruction here, I might as well try to do something as close as I can.

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,196 @@
...
+
+	f, a := c.ActionFrames(core.ActionCharge, p)
+
+	ai := core.AttackInfo{
+		ActorIndex: c.Index,
+		Abil:       "Charge Attack",
+		AttackTag:  core.AttackTagExtra,
+		ICDTag:     core.ICDTagNone,
+		ICDGroup:   core.ICDGroupDefault,
+		StrikeType: core.StrikeTypeBlunt,
```
srliao, line 56: this shouldn't be blunt. i don't think barb attacks are blunt

internal/characters/barbara/abil.go
```diff
@@ -0,0 +1,196 @@
...
+		Element:    core.Hydro,
+		Durability: 25,
+		Mult:       charge[c.NormalCounter][c.TalentLvlAttack()],
```
srliao, line 59: why's there an array index for normalcounter here? her charge attack is only one attack right? shouldn't depend on the normal string?

internal/characters/barbara/abil.go
HanSeoulOh: fixed

internal/characters/barbara/abil.go
HanSeoulOh: came from copying yanfei as template kek

HanSeoulOh: fixed

PR #122, opened 2021-12-31
guoba swirl

### LINKED ISSUE #121: sucrose guoba swirl

srliao:
fix #121


srliao:
this is kinda ugly because guoba is literally on field as a target forever, instead of added and removed. however current system doesn't really play nice on removing because it'd mess up the stats

PR #166, opened 2022-01-13
hakushinring implementation

HanSeoulOh:


internal/weapons/catalyst/hakushin/hakushin.go
```diff
@@ -0,0 +1,47 @@
...
+
+	for _, char := range c.Chars {
+		char.AddMod(core.CharStatMod{
+			Key: "hakushin-passive",
+			Amount: func() ([]float64, bool) {
+				return m, expiry > c.F
+			},
+			Expiry: -1,
+		})
+	}
```
Tsym-or-Tysm, line 39: We should not be coding expiry like this anymore - duration should be tracked directly in the Expiry field to promote tracking.

internal/weapons/catalyst/hakushin/hakushin.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+
+	"github.com/genshinsim/gcsim/pkg/core"
+)
+
+func init() {
+	core.RegisterWeaponFunc("hakushin ring", weapon)
+	core.RegisterWeaponFunc("hakushin", weapon)
```
Tsym-or-Tysm, line 11: You should probably add the genshin-db name here...

internal/weapons/catalyst/hakushin/hakushin.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func weapon(char core.Character, c *core.Core, r int, param map[string]int) string {
+
+	expiry := 0
+	e := 10 + float64(r)*2.5
```
Tsym-or-Tysm, line 17: This seems wrong on two levels, first you don't divide this by 100 anywhere so this is going to be a 1000% buff...

Second the R1 buff is 10% not 12.5%

internal/weapons/catalyst/hakushin/hakushin.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+			m[ele] = e
+			expiry = c.F + 6*60
+			return false
+		}
+	}
+
+	for _, char := range c.Chars {
+		char.AddMod(core.CharStatMod{
+			Key: "hakushin-passive",
```
Tsym-or-Tysm, line 33: I think this will cause collisions in naming on buff refresh? Pretty sure you need a separate name for each element.

Tsym-or-Tysm (changes requested):
Did you test that this works as expected?

You're also missing a change to imports.go.

internal/weapons/catalyst/hakushin/hakushin.go
```diff
@@ -23,21 +23,20 @@ func weapon(char core.Character, c *core.Core, r int, param map[string]int) stri
 				return false
 			}
 			m[ele] = e
-			expiry = c.F + 6*60
+
+			for _, char := range c.Chars {
+				char.AddMod(core.CharStatMod{
+					Key: "hakushin-passive-" + key,
+					Amount: func() ([]float64, bool) {
+						return m, true
```
srliao, line 31: not sure if this is intentional or not (haven't looked into how this thing works) but because m is declared way at the top, you can get a situation where you're adding elemental dmg for multiple elements at the same time because you're not resetting the values

Tsym-or-Tysm:
Hmm I just pulled this and looked at it - something doesn't seem right here... I'm not sure how it's adding attack of all things.

image

Config: https://viewer.gcsim.app/share/bB4ZUB3xcrLnOAXqENh0C

HanSeoulOh:
> Hmm I just pulled this and looked at it - something doesn't seem right here... I'm not sure how it's adding attack of all things.
> 
> image
> 
> Config: https://viewer.gcsim.app/share/bB4ZUB3xcrLnOAXqENh0C

was not able to reproduce

Tsym-or-Tysm (approved):
Looks good to me to merge

PR #176, opened 2022-01-21
fix albedo burst cd, babara skill scaling

srliao:


srliao:
Forgot to add it in commit message... Barbara fix credit: Shizuka#7791

PR #181, opened 2022-01-22
Initial attempt at KQM substat optimization implementation

Tsym-or-Tysm:
Currently a completely beta/rough implementation, coded up as a proof of concept.

Performance is currently pretty solid among the configs that I've tested. [This sheet](https://docs.google.com/spreadsheets/d/1ThRAQSCQ0o0kzjyG6bFjYIR98yhGCk32GzlV53PwaQw/edit#gid=0) contains a sampling of the configs that I've tested - so far haven't encountered any issue where this implementation completely bombs out.

image

Tried a few external optimization packages to evaluate performance, but in the end things ended up too slow with difficult general parameter tuning so I manually implemented a lot of stuff which can be done better.

Please give comments/suggestions - I would be totally ok with changing large chunks of this if possible since it's so ugly...

### DIFF
--- a/cmd/gcsim/optimize_substats.go
+++ b/cmd/gcsim/optimize_substats.go
@@ -1,0 +1,628 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"math"
+	"os"
+	"path"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/genshinsim/gcsim"
+	"github.com/genshinsim/gcsim/pkg/core"
+	"github.com/genshinsim/gcsim/pkg/parse"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+var (
+	// Only includes damage related substats scaling. Ignores things like HP for Barbara
+	charRelevantSubstats = map[core.CharKey][]core.StatType{
+		core.Albedo:  {core.DEFP},
+		core.Hutao:   {core.HPP},
+		core.Kokomi:  {core.HPP},
+		core.Zhongli: {core.HPP},
+		core.Itto:    {core.DEFP},
+		core.Yunjin:  {core.DEFP},
+		core.Noelle:  {core.DEFP},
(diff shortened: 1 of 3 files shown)


HanSeoulOh:
well i need to use this so might as well review it

HanSeoulOh:
Ok there is a million outputs now, and it's not clear which one is which anymore. We may need clean it up a little but it is very low priority because there's only a handful of us using the CLI.

cmd/gcsim/optimize_substats.go
```diff
@@ -0,0 +1,673 @@
...
+	var data strings.Builder
+	var re = regexp.MustCompile(`(?m)^import "(.+)"$`)
+
+	// Regex to identify main stats based on flower. Check that characters all have one that we can recognize
+	var reMainstats = regexp.MustCompile(`(?m)^[A-z]+ add stats.* hp=(4780|3571).*\n`)
+	if len(reMainstats.FindAll(src, -1)) != 4 {
+		sugarLog.Error("Error: Could not identify valid main artifact stat rows for all characters based on flower HP values.")
+		sugarLog.Error("5* flowers must have 4780 HP, and 4* flowers must have 3571 HP.")
+		os.Exit(1)
+	}
```
HanSeoulOh, line 110:
```suggestion
	var reMainstats = regexp.MustCompile(`(?m)^[A-z]+ add stats.* hp=(4780|3571).*\n`)
	var numChars = regexp.MustCompile(`char`)

	if len(reMainstats.FindAll(src, -1)) != len(numChars.FindAll(src, -1)) {
		sugarLog.Error("Error: Could not identify valid main artifact stat rows for all characters based on flower HP values.")
		sugarLog.Error("5* flowers must have 4780 HP, and 4* flowers must have 3571 HP.")
		os.Exit(1)
	}

```

To allow for substat optimization for less than 4 chars

HanSeoulOh (changes requested):

Tsym-or-Tysm:
Note that there will still need to be cleanup done on this at some point - will re-open another issue for it.

PR #205, opened 2022-01-30
Refactor results/workers

srliao:
This PR is to split up the current `worker.go` into multiple reasonable packages in preparation for potentially moving in a webapp based on the following design


srliao:
will redo PR when next is ready to be merged

PR #206, opened 2022-01-30
Swap Delay and Action Delay

Tsym-or-Tysm:
- Add "defaults" keyword to config, with ability to set "swap_delay"
- Add "delay" parameter to allow for custom delays after actions

Works in configs like this, which adds a global 8 frame swap delay, and a 60 frame delay to Hu Tao CA on top of the standard animation frames.
```
defaults swap_delay=8;

hutao attack:2,charge[delay=60],dash
      +if=.status.paramita>60
      && .stam.hutao>180;
```

### DIFF
--- a/pkg/core/action.go
+++ b/pkg/core/action.go
@@ -393,6 +393,11 @@ func (a *ActionCtrl) execAction(n *ActionItem) (int, bool, error) {
 	return f, true, nil
 }
 
+// Each "abil" function returns two frames:
+// First int is the total animation frame, or the number of frames the ability should last.
+// Second int is the first frame after using an ability that the sim can start looking to queue the next action.
+// Supports user config overriding the total animation frames (second number)
+// TODO: Second int is basically completely unimplemented
 func (a *ActionCtrl) execActionItem(
 	n *ActionItem,
 	pre, post EventType,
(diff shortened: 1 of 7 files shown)


srliao:
@Tsym-or-Tysm just looking at alternative/cleaner way of implementing this. can we safely assuming that `delay` should only be used for actions that are NOT one of the following:

- swap
- dash
- jump

I think there's a good place in `core` to check for this. Just need to know if we should allow `delay` param on ALL action types? or can we exclude the above 3

Tsym-or-Tysm:
Hmmm IMO I'd prefer it to be available for all action types for consistency. Otherwise things get very confusing because now people have to guess at what actions have `delay` available and what actions don't, which can lead to a lot of additional questions (since it's clear by now that people don't always read up on the available documentation lol).

Tsym-or-Tysm:
@srliao I saw that you added swap delay in the `next` branch, but did the action delay make it in anywhere? Too many changes and I don't know if it wasn't mentioned in commit messages or something... If so then we can close this out.

srliao:
@Tsym-or-Tysm i forgot to push the changes.. it was on my local branch. See https://github.com/genshinsim/gcsim/blob/76aef4b45c7ca463824d3ad7d84dbd833db39264/pkg/simulation/run.go#L115

PR #208, opened 2022-01-30
Bugfixes

Tsym-or-Tysm:
Not yet done - going to add/fix later.


Tsym-or-Tysm:
I removed the proposed change to add panics in for now since it will block this patch for some time. Assuming no major concerns, this is ok to merge in now.

PR #214, opened 2022-01-30
Fixes / Kokomi's A4/Skill/Burst & Prototype Amber

imring:
- Fix bonus from Kokomi's A4;
- Fix heal bonus from Kokomi's skill/burst;
- Fix heal bonus from passive of Prototype Amber;


Tsym-or-Tysm (approved):
This all looks fine to me so we can merge this in. I may make a few more changes at a later date, but this aligns with what I was finding as well.

Thanks for the PR!

Issue #203, opened 2022-01-28
Rework All Mods to Switch to Using Expiry Rather than Custom Expiration Methods


Tsym-or-Tysm:
Will be far more useful for buff/debuff tracking moving forward.

We may also want to evaluate whether this leads to performance issues, as mods are constantly getting added/removed as opposed to being set once at the start. It's possible that we may want to move to some kind of other system? Not sure...

srliao:
i think all but a couple mods are fixed now?

srliao:
Searched for `modifier\.New(.+), -1` and only noted the following that are still using -1 as expiry (that's not permanent):

- [ ] Aloy's coil stack damage increase
- [ ] Ganyu's C4
- [ ] Heizou's declension stacks
- [ ] Itto's A1
- [ ] Mona's bubble damage bonus
- [ ] Mona's C4
- [ ] Razor attack speed on burst
- [ ] Razor's ER on sigil
- [ ] Sayu's C2
- [ ] Shenhe's C4
- [ ] Xinyan's A4
- [ ] Yoimiya A1
- [ ] Mappa Mare proc
- [ ] Lost Prayer stacks

k0l11:
new TODO (waiting for mod/stat array refactors to allow for easily adding new enemy modifiers, namely damage bonus and crit rate mods attached to enemies):
- [ ] ganyu c4 damage bonus
- [ ] mona burst damage bonus
- [ ] mona c4 crit rate
- [ ] toukabou shigure damage bonus

PR #227, opened 2022-02-03
Rewrite Calamity Queller

shizukayuki:


internal/weapons/spear/calamity/calamity.go
```diff
@@ -11,66 +11,56 @@ func init() {
...
+
+		if expiry < c.F {
+			skillInitF = c.F
+		}
+		expiry = c.F + 60*20
+
+		return false
+	}, fmt.Sprintf("calamity-queller-%v", char.Name()))
 
 	char.AddMod(core.CharStatMod{
```
srliao, line 44: it's probably best if this AddMod gets moved into the Subscribe event with a proper expiry set. We're trying to clean up a bit and avoid using -1 (which is just laziness on my part). See https://github.com/genshinsim/gcsim/issues/203

shizukayuki: i think the AddMod will have to stay because the Ele% is permanent buff.
unless i am missing something here or you meant the ATK buff...
i could try moving the ATK buff inside the event but don't know if i can do that while keeping stacks from previous cast.

srliao: Oh sorry you're right. Let me think about this for a bit

PR #231, opened 2022-02-04
Moonglow Implementation

imring:


internal/weapons/catalyst/moonglow/moonglow.go
```diff
@@ -0,0 +1,65 @@
...
+	char.AddMod(core.CharStatMod{
+		Key: "moonglow-heal-bonus",
+		Amount: func() ([]float64, bool) {
+			return mheal, true
+		},
+		Expiry: -1,
+	})
+
+	matk := make([]float64, core.EndStatType)
+	matk[core.ATK] = (0.0005 + float64(r)*0.0005) * char.MaxHP()
```
srliao, line 27: i don't think anything exists in game right now that modify maxhp in combat but to be future proof probably best moving this line into the Amount function?

i would also move `atk.Info.AttackTag == core.AttackTagNormal` into an if and return nil if not true so that way the maxhp calculation doesn't get called if not required (after moving it into the amount function that is)

internal/weapons/catalyst/moonglow/moonglow.go
```diff
@@ -0,0 +1,65 @@
...
+	char.AddPreDamageMod(core.PreDamageMod{
+		Key: "moonglow-na-bonus",
+		Amount: func(atk *core.AttackEvent, t core.Target) ([]float64, bool) {
+			return matk, (atk.Info.AttackTag == core.AttackTagNormal)
+		},
+		Expiry: -1,
+	})
+
+	icd, dur := -1, -1
+	c.Events.Subscribe(core.PostBurst, func(args ...interface{}) bool {
```
srliao, line 37: just as an FYI, PostBurst is called AFTER the animation frames. i'm not sure how moonglow works but if this is not desirable behavior then might want to hook onto PreBurst

srliao (changes requested):

internal/weapons/catalyst/moonglow/moonglow.go
imring: > PostBurst is called AFTER the animation frames
oh, i'll replace to PreBurst

PR #244, opened 2022-02-11
Implement Kazuha CA, Fix Beidou C6, Kokomi A4/Skill, Ganyu Bloom, Tartaglia A4, PaleFlame, NA/CA/Plunge snapshot timing

shizukayuki:
- Implement Kazuha CA
- Fix Beidou C6, now start right after burst initial hit
- Change Kokomi A4 to be dynamic
- Change Ganyu Bloom to include CA travel time
- Fix NA/CA/Plunge snapshot timing for most chars (expect catalyst and bow)
- Raiden polearm CA don't share ICD with NA
- Fix Electro Traveler skill particle (was Cryo)
- Change Geo Traveler A2 to scale off NormalAttackDMG%
- Allow Kokomi Skill to snapshot burst bonus if swapped out early (before 2nd tick)
- Change Tartaglia melee NA/CA to apply Riptide before checking for A4 proc
- Fix Pale Flame stacks


internal/characters/travelergeo/abil.go
```diff
@@ -21,22 +21,22 @@ func (c *char) Attack(p map[string]int) (int, int) {
...
 
-	c.Core.Combat.QueueAttack(ai, core.NewDefCircHit(0.3, false, core.TargettableEnemy), 0, f-1)
+	c.Core.Combat.QueueAttack(ai, core.NewDefCircHit(0.3, false, core.TargettableEnemy), f-1, f-1)
 
 	if c.NormalCounter == c.NormalHitNum-1 {
 		//add 60% as geo dmg
 		ai := core.AttackInfo{
 			ActorIndex: c.Index,
 			Abil:       "A2",
-			AttackTag:  core.AttackTagNone,
```
srliao, line 31: Are you sure about this one? I thought I was right on this one having no attack tag. There was a ticket in KQM on this I think?

srliao (commented):
Looked at most of it. The Kokomi changes I'm not sure about because I (still) don't know how she works.

The change from PostBurst to PreBurst I'm also not sure about because I don't know the actual timing of those things. Do we have evidence to support these changes?

Rest of the snapshot changes looks good

shizukayuki:
> The change from PostBurst to PreBurst I'm also not sure about because I don't know the actual timing of those things. Do we have evidence to support these changes?

maybe i should revert this change.. all i know is that the passive start on cast if the atk buff anim can be trusted.

> Looked at most of it. The Kokomi changes I'm not sure about because I (still) don't know how she works.

Kokomi NA/CA don't snapshot A4. it is possible to buff NA when doing `N1 > Q`
as for Kokomi skill it used to keep the burst dmg bonus in sim, when swapping out after 2nd tick.

the skill snapshot window start at 1st and end a bit before 2nd tick, and it only snapshot the burst dmg bonus.
i'm not entirely sure about the timing, so for now i coded it to allow for N1 before swapping.
and the snapshot only happens if you swap out while in that window and have burst active.

https://library.keqingmains.com/evidence/characters/hydro/kokomi#snapshot-burst-dmg-bonus-onto-bake-kurage-by-switching

internal/characters/travelergeo/abil.go
shizukayuki: its in KQM TCL

https://library.keqingmains.com/characters/geo/traveler-geo#frenzied-rockslide
> This damage bonus is also considered Normal Attack Damage that also deals Elemental DMG.

internal/characters/kokomi/abil.go
```diff
@@ -214,10 +217,6 @@ func (c *char) Burst(p map[string]int) (int, int) {
 	if c.Core.Status.Duration("kokomiskill") > 0 {
 		// +1 to avoid same frame expiry issues with skill tick
 		c.Core.Status.AddStatus("kokomiskill", 12*60+1)
-		c.skillLastUsed = c.Core.F - 1
-		d := c.createSkillSnapshot()
-		// Tick intervals stay the same if duration is refreshed
-		c.AddTask(c.skillTickTask(d, c.Core.F), "kokomi-e-ticks", 120-(c.Core.F-c.skillLastTick))
```
Tsym-or-Tysm, line 220: I don't think you can remove this since Kokomi's E gets re-snapshot when you cast her Q.

Tsym-or-Tysm (changes requested):

internal/characters/kokomi/abil.go
shizukayuki: Seems like the burst doesn't re-snapshot the skill, only extends the duration.

calc: https://i.imgur.com/ifTczDY.png
video: https://www.youtube.com/watch?v=vmdj37M2Dxw

Tsym-or-Tysm: Sounds good, then we can go ahead and merge this. Make sure to submit something to the KQM website feedback too so the library gets fixed before someone else comes in and asks about it at some point down the line lol

Tsym-or-Tysm (approved):

PR #248, opened 2022-02-12
Implement compound bow

jordanlovato:
WIP, not 100% sure how to test/verify yet

### DIFF
--- a/internal/weapons/bow/compound/compound.go
+++ b/internal/weapons/bow/compound/compound.go
@@ -1,0 +1,80 @@
+package generic
+
+import (
+	"fmt"
+
+	"github.com/genshinsim/gcsim/pkg/core"
+)
+
+func init() {
+	core.RegisterWeaponFunc("compoundbow", weapon)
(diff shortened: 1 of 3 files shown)


internal/weapons/bow/compound/compound.go
```diff
@@ -0,0 +1,80 @@
...
+
+		// Check if char is correct?
+		if atk.Info.ActorIndex != char.CharIndex() {
+			return false
+		}
+
+		// Only apply on normal or charged attacks
+		if (atk.Info.AttackTag != core.AttackTagNormal) || (atk.Info.AttackTag != core.AttackTagExtra) {
+			return false
+		}
```
HanSeoulOh, line 52: This part doesn't do what you think it does. The logic always evals to true, you need to use XOR here or change your boolean statement.

HanSeoulOh (changes requested):

internal/weapons/bow/compound/compound.go
jordanlovato: Oh derp, ty! Nice catch =]

PR #249, opened 2022-02-12
Jl 220212 add compound bow

jordanlovato:
Adds compound bow, feedback appreciated!


internal/weapons/bow/compound/compound.go
```diff
@@ -0,0 +1,79 @@
...
+
+	incAtk := .03 + float64(r)*0.01
+	incSpd := 0.009 + float64(r)*0.003
+
+	stacks := 0
+	maxStacks := 4
+	stackExpiry := 0.0
+	stackDuration := 60.0 * 6.0 // frames = 6s * 60 fps
+
+	cd := .3 * 60 // frames = 0.3s * 60fps
```
srliao, line 30: this should be hard coded to 18 frames. this way icd doesn't need to be a float and you don't need to incur the cost of casting it back and forth

internal/weapons/bow/compound/compound.go
```diff
@@ -0,0 +1,79 @@
...
+			return false
+		}
+
+		// Check if cd is up
+		if icd > float64(c.F) {
+			return false
+		}
+
+		// Assuming all stacks fall off after 6s
+		if float64(c.F) > stackExpiry {
```
srliao, line 62: this check i think needs to be in the amount call?

jordanlovato:
@srliao Updated, fixed the typecasting and move the stackExpiry check to the mod calc routine

Tsym-or-Tysm (changes requested):

internal/weapons/bow/compound/compound.go
```diff
@@ -0,0 +1,79 @@
...
+		Expiry: -1,
+	})
+
+	c.Events.Subscribe(core.OnDamage, func(args ...interface{}) bool {
+		atk := args[1].(*core.AttackEvent)
+
+		// Check if char is correct?
+		if atk.Info.ActorIndex != char.CharIndex() {
+			return false
+		}
```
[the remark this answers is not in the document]
srliao: Technically this check is correct but I guess it would only matter if we implement the frames for a normal attack or charge attack inappropriately....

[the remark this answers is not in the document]
Tsym-or-Tysm: I'd prefer to have the active character check always explicitly made. For example, Tartaglia Riptide burst is considered NA damage and can be proc'd from off field, so you can't always assume that NA/CA damage is purely from on field. If we ever get a character that works like that in the future it'd be nice if we didn't have to go back and add this back in for all the weapons...

internal/weapons/bow/compound/compound.go
```diff
@@ -72,6 +63,20 @@ func weapon(char core.Character, c *core.Core, r int, param map[string]int) stri
 		icd = c.F + cd
 		stackExpiry = c.F + stackDuration
 
+		char.AddMod(core.CharStatMod{
+			Key: "compoundbow",
+			Amount: func() ([]float64, bool) {
+				if c.F > stackExpiry {
+					stacks = 0
+				}
```
Tsym-or-Tysm, line 71: Hmmm this method of setting stacks to 0 seems unsafe to me actually, as I think the `Amount` function isn't called if a mod is expired, so there's the risk of `stacks` remaining at 4 (or whatever number) forever basically.

I think the easiest solution to resolve this is to add a status whenever compound bow is proc'd, then change this line to reset stacks to 0 if the duration = 0.  Then there would be one other duration check right before the stack increment check to reset stacks to 0 which should cover all cases?

@srliao maybe I'm overcomplicating this though and it should just work?

srliao: IMO this is fine since it doesn't matter if stacks remain at 4 as long as we don't add 4 stacks. However I meant to say that this check should be both here in the Amount call and also in the OnDamage call... since the OnDamage call needs to make sure it doesn't start off at 4 stacks when it should have reset. 

@jordanlovato Sorry about that. Misleading comment from me.

Tsym-or-Tysm (approved):

Tsym-or-Tysm:
Looks good to me - ready for merge from my perspective if @srliao has no further comments.

jordanlovato:
Oh lord, pruned the wrong branch, reopening..

jordanlovato:
Merged main to fix the imports.go conflict

Tsym-or-Tysm:
Merged - thanks!

PR #258, opened 2022-02-14
Add walk action (#239)

### LINKED ISSUE #239: Add Walk Cancels Consistently
See title.

Currently only Klee has a manual exception for this, but technically this can be done with every character.

Currently mostly only relevant for catalyst attackers... Not sure if there's any instance of this happening often for non-catalysts since they mostly use dashes/jumps.

Tsym-or-Tysm:
Resolves #239


srliao:
Going to hold this one until next is merged. Otherwise merge conflicts is going to be a pain

PR #292, opened 2022-02-20
yae miko change destruction of kitsune totem on Q to instant

HanSeoulOh:


Tsym-or-Tysm:
I haven't looked at the Yae implementation much so I'm going to punt this to srl since he's looked at the code more than me.

PR #300, opened 2022-02-22
catch for not enough energy for burst

HanSeoulOh:
it's not obvious to people what's going on when burst fails energy check. suggestion for how to fix

### DIFF
--- a/internal/tmpl/action/action.go
+++ b/internal/tmpl/action/action.go
@@ -166,7 +166,11 @@ func (a *Ctrl) execAction(n *core.ActionItem) (int, bool, error) {
 
 	//do one last ready check
 	if !c.ActionReady(n.Typ, n.Param) {
-		a.core.Log.NewEvent("queued action is not ready, should not happen; skipping frame", core.LogSimEvent, -1)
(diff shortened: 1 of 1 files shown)


internal/tmpl/action/action.go
```diff
@@ -166,7 +166,11 @@ func (a *Ctrl) execAction(n *core.ActionItem) (int, bool, error) {
 
 	//do one last ready check
 	if !c.ActionReady(n.Typ, n.Param) {
-		a.core.Log.NewEvent("queued action is not ready, should not happen; skipping frame", core.LogSimEvent, -1)
+		if n.Typ == core.ActionBurst && c.CurrentEnergy() < c.MaxEnergy() {
+			a.core.Log.NewEvent("insufficient energy for burst", core.LogActionEvent, c.CharIndex(), "have", c.CurrentEnergy())
```
srliao, line 170: if you want to add this log, this should be in the character template, not in actions.

HanSeoulOh: aight u figure it out then gg

Issue #295, opened 2022-02-21
Yae NAs Do Not Have Counters in Debug Log


Tsym-or-Tysm:
Currently they just all say "Normal", which is confusing.

PR #305, opened 2022-02-23
Bugfixes

shizukayuki:
- Fix oathsworneye scaling
- amos: use snapshot frame for travel time
- prototype/skyward: change post to pre
- thunderingfury: add active char check
- kazuha/abil.go: convert line endings to unix
- Add NA counter for shenhe/yaemiko (#295)
- Change Kazuha A4 to use expiry field (#297)
- Fix Sucrose C6 (#309)


srliao:
Hmm I wonder why issues aren't being linked.

should fix #295 #297 #309

PR #311, opened 2022-02-24
fix bennett self aura; fix minor typo

### LINKED ISSUE #259: Bennett Q self pyro aura duration is incorrect
From Charlie and vigne: https://discord.com/channels/845087716541595668/869210750596554772/942589094087786627
https://discord.com/channels/845087716541595668/869210750596554772/936250834340171817

The aura duration seems to currently be infinite or something?

Per Charlie, it seems that the pyro aura is linked to the buff duration? But the exact decay mechanism is unknown - maybe should look at Dim's repo for that.

srliao:
fix #259

### DIFF
--- a/internal/characters/bennett/bennett.go
+++ b/internal/characters/bennett/bennett.go
@@ -220,9 +220,10 @@ func (c *char) skillHoldLong() {
 
 }
 
+const burstStartFrame = 31
+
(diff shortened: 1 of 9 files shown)


Tsym-or-Tysm (approved):
I assume that you tested it and it works? If so then looks fine to me - changes seem fairly straightforward.

PR #313, opened 2022-02-24
Fix first kitsune tick multiplier

HanSeoulOh:


HanSeoulOh:
wen merge

srliao:
self review.. what could go wrong. oh well if it breaks tsym will have both of our heads (again)

PR #323, opened 2022-02-25
add ended status tracking for mods

srliao:
this should let you click on mods for charstat, predamagemod, enemy res and def mod and see the duration

only issue is that refresh only shows the duration of that refresh but if you want to see the entire time that the buff is active you gotta find the original added...

### DIFF
--- a/internal/tmpl/character/mods.go
+++ b/internal/tmpl/character/mods.go
@@ -1,0 +1,110 @@
+package character
+
+import "github.com/genshinsim/gcsim/pkg/core"
+
+func (c *Tmpl) AddPreDamageMod(mod core.PreDamageMod) {
+	ind := -1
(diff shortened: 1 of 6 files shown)


Tsym-or-Tysm:
Two problems with your changes:

1) You've introduced this null "Evt" value into logs - is that supposed to be used at all? If not then I'd prefer to not include it for clarity purposes
image
2) You've removed the message that shows up when a status/mod ends. In this config, Xiangling C1 is supposed to expire on frame 1064, but when I go there nothing shows up.

https://gcsim.app/viewer/share/IR0oj-nXW72bj1Y8D8vur

srliao:
@Tsym-or-Tysm hid the Evt field but the other one should be fixed in UI not back end. If you're ok with it otherwise should merge it and leave 2. as an issue

PR #325, opened 2022-02-25
fix initial mods added in wrong col

### LINKED ISSUE #293: Initialized Mods Are in Wrong Column in Debug
image

Issue occurs because some mods are initialized before each character gets their character index, so a lot of them group up in character index 0.

srliao:
fixes #293


Tsym-or-Tysm (approved):
Oh this fix works too and is much cleaner. Tested on a few configs and none seem to break - should be ready to merge.

PR #327, opened 2022-02-25
custom global delay, frames framework, kazuha frame update

### LINKED ISSUE #267: Add Global Delays for Each Action Type, and Add Default Setting for "Human Frames"
Related to the frames/particles issue, I did some frame counts, and my sense is that while we could fine tune a few of these frame counts, the overall problem is not really about particle timing or the ending animation cancels, but rather that people unconsciously do all these micro-waits in game, and then expect the sim to mostly just work out of the box, and bounce off somewhat hard from the sim when it doesn't.

I think that aside from the per-each action delay stuff, we should probably also add in a setting for global delays across each type of action, that kind of simulate something closer to human play, and then make that our "human default" settings because there's no way we can do an individualized setting for each character. So something like the below:
```
defaults swap_delay=8 burst_delay=20 skill_delay=10 charge_delay=8;

rosaria skill[delay=20];
```
Would set swap delay to 8, and would add the number of frames to all usages for each action in the sim. So for example each skill usage would have 10 frames of ending animation by default, and then Rosaria E would have another 20 frames (total of 30 frames of delay).

We can then just create some reasonable global delays based on a few videos and add them as a global option like so:
```
defaults frame_counts="human";
```
which could be treated as equivalent to the above `defaults` statement or something. That will probably catch like 90% of all use cases, and then the remaining people who need really fine grained control (mostly Hu Tao I'm guessing since her frame timing is really tight) can add in manual waits.

Related to #171 and #209.

srliao:
fixes #267 
@Tsym-or-Tysm this one probably need extensive testing

I tested with the following config and it works fine but I'm not overly confident on the logic especially on `/pkg/simulation/run.go` L114

```
options debug=true iteration=1 duration=90 workers=24 mode=apl attack_delay=100 skill_delay=100;

bennett char lvl=90/90 cons=5 talent=9,9,9; 
bennett add weapon="aquilafavonia" refine=1 lvl=90/90;

##Default Enemy
target lvl=100 pyro=0.1 dendro=0.1 hydro=0.1 electro=0.1 geo=0.1 anemo=0.1 physical=.1 cryo=.1;


active bennett;

bennett attack;
```

### DIFF
--- a/cmd/gcsim/.gitignore
+++ b/cmd/gcsim/.gitignore
@@ -3,3 +3,4 @@
 *.gz
 gcsim
 gcsim.exe
+config.txt
--- a/internal/characters/kazuha/abil.go
+++ b/internal/characters/kazuha/abil.go
@@ -6,6 +6,14 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core"
 )
 
+var hitmarks = [][]int{
+	{12},         //n1
+	{11},         //n2
+	{16, 25},     //n3
+	{15},         //n4
+	{15, 23, 31}, //n5
+}
+
 func (c *char) Attack(p map[string]int) (int, int) {
 
 	f, a := c.ActionFrames(core.ActionAttack, p)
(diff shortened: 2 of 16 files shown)


Tsym-or-Tysm:
I will have to review in more depth, but since you've already done this I think you should use this PR to also include a `frame_defaults=human` option that uses a reasonable setting for everything.

Based on the frame counts that I've done here in the "Rotation Frame Counts" sheet, it looks to me like setting a 5 frame delay on everything is the best way to go, maybe with a 8 frame delay on swaps or something. People can then override the settings as desired (i.e. if you set `frame_defaults=human attack_delay=0` then you'll get a 5 frame delay on everything except for attacks that are set at 0 frame delay).

https://docs.google.com/spreadsheets/d/1inm3-5TazgK9OKiEl79hp2T-f453DWTmkEDekLasJy8/edit#gid=1568164268

srliao:
added

srliao:
Adding the frames framework and kazuha frames update to this PR since the frames stuff depends on some of the changes made here

PR #335, opened 2022-02-25
add error if no target set

### LINKED ISSUE #334: sim doesn't warn user when there's no target
some newbie users stuck malding wondering why there's 0 damage. dev plz do something

srliao:
fixes #334


HanSeoulOh (approved):
based devs

PR #336, opened 2022-02-25
change yae totem status to oldest_totem_expiry

### LINKED ISSUE #332: Rename Yae oldestTotemTime Status?
The "oldestTotemTime" status seems to be poorly named. What it actually tracks is the frame that the oldest totem will expire on.

Maybe we should call it "oldestTotemExpiry"?

srliao:
fixes #332 

test config: https://gcsim.app/viewer/share/NzAud0AZxqV_otlG-JWEu

status duration is 900, set to attack only if expiry < 500, attack starts at frame 401 as expected

### DIFF
--- a/internal/characters/yaemiko/yaemiko.go
+++ b/internal/characters/yaemiko/yaemiko.go
@@ -22,7 +22,7 @@ type char struct {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


HanSeoulOh:
reluctant merge

PR #343, opened 2022-02-26
Add some statuses.

imring:
- Add `ayakaci` status for Ayaka's cryo infusion.
Example:
```
ayaka dash +if=.status.ayakaci == 0;
ayaka attack:2, charge;
```
- Add `ganyuburst` status for Ganyu's burst.
Example:
```
sr_seq: sucrose skill;
ay_seq: ayaka dash, burst;

ganyu burst;
chain sr_seq, ay_seq +if=.status.ganyuburst > 0;
```


srliao:
@imring I think Ayaka status should just be changed into a general infusion config instead. So should spin that one out into an issue on it's own

srliao:
@imring for now can you refuse the ayaka status? i feel like we'll forget if we leave it in there

imring:
> @imring for now can you refuse the ayaka status? i feel like we'll forget if we leave it in there

done

Issue #120, opened 2021-12-31
jean infusion field is currently bugged


srliao:
infusion/absorb check not coded in

srliao:
fixed in #426

PR #354, opened 2022-02-27
target refactor

### LINKED ISSUE #129: refactor target interface
@Tsym-or-Tysm as discussed, need to strip out the following methods:

```
	//basic info
	Type() TargettableType //type of target
	MaxHP() float64
	HP() float64

	//tags
	SetTag(key string, val int)
	GetTag(key string) int
	RemoveTag(key string)

	//target mods
	AddDefMod(key string, val float64, dur int)
	AddResMod(key string, val ResistMod)
	RemoveResMod(key string)
	RemoveDefMod(key string)
	HasDefMod(key string) bool
	HasResMod(key string) bool

```

so we're left with just
```

	Index() int            //should correspond to index
	SetIndex(ind int)      //update the current index

	//collision detection
	Shape() Shape

	//attacks
	Attack(*AttackEvent) (float64, bool)

	//reaction/aura stuff
	Tick()
	AuraContains(...EleType) bool
	AuraType() EleType

	//getting rid of
	Kill()
```

and instead any method that needs to know what type it is can either use a type switch (which seems to be just as fast as calling Type()) and cast it to the appropriate type

the stuff stripped out can be enemy specific

srliao:
This PR is to track progress on refactoring the target interface required for a couple reasons:

- To fix #129 
- To allow better system for #120 

The current Target interface is clunky and require implementing way too many methods that are not relevant to every target type. In addition, `Reactable` is only meant for enemies and cannot handle self-infused elements triggering reactions properly. `Reactable` needs to work with both player and enemies.


srliao:
dropping this refactor

PR #357, opened 2022-02-28
Fix barbara, kokomi skill, yaemiko a4, ttds, 4pc shim

### LINKED ISSUE #353: Thrilling Tales does not go on cooldown if it is passed via "active <catalyst user>" and the first action in the sim is a swap
https://gcsim.app/viewer/share/perm_k9mS0zCAEsa8xE-Ta6DRx -- in here, `active sucrose` is used but Fischl is the first unit to perform an action, so the sim swaps to her and she gets Thrilling Tales. A few seconds later, Sucrose is finally on the field but Thrilling Tales goes to Beidou when she leaves as it never went on cooldown.

https://gcsim.app/viewer/share/perm_ADNuSB2OTvWSaDolXS0MV -- same config with `active sucrose`, but the first action has sucrose on the field. Thrilling Tales functions as expected.

### LINKED ISSUE #359: barb has no element set

### LINKED ISSUE #360: yae a2 does not calculate using correct EM
image

shizukayuki:
- Add cooldown to 4pc shimenawa
- Fix kokomi skill adding burst bonus if `core.F = 0`
- Fix #353
- Fix #359
- Fix barbara skill consuming energy and proc'ing 4tom
- Fix #360

### DIFF
--- a/internal/artifacts/reminiscence/reminiscence.go
+++ b/internal/artifacts/reminiscence/reminiscence.go
@@ -17,11 +17,11 @@ func New(c core.Character, s *core.Core, count int, params map[string]int) {
 		m := make([]float64, core.EndStatType)
 		m[core.ATKP] = 0.18
 		c.AddMod(core.CharStatMod{
-			Key: "shim-2pc",
+			Key:    "shim-2pc",
+			Expiry: -1,
 			Amount: func() ([]float64, bool) {
 				return m, true
 			},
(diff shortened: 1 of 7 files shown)


HanSeoulOh (approved):
Looks good to me @srliao you may want to take a look at Diona and Kokomi changes. Will merge tonight if no comment.

PR #383, opened 2022-03-06
Implement Tartaglia C4

### LINKED ISSUE #306: Childe C4 is not implemented
I have the constellation so feel free to ask questions/have me test

It works like this: every 4 seconds that a target is marked by Riptide, a Riptide Slash or Riptide Flash will occur on the target. A Slash will occur if Childe is on the field and in melee stance. A Flash will occur if Childe is not on the field, or on the field in ranged stance.

Slash normally has a 1.5s cooldown and Flash normally has a 0.7s cooldown. But these cooldowns on C4 Slashes and Flashes are separate from the cooldown on "regular" Slashes and Flashes. 

C4 Slashes and Flashes _do_ share aura application ICDs and particle generation ICDs with regular Slashes and Flashes.

Also C4 Slashes and Flashes do not proc instantly. 4 seconds must pass before they activate.

shizukayuki:
- Resolves #306


srliao:
Hmmm to be perfectly honest I'm not sure how this works so I have no idea what I'm looking at... might take me a while to get through this.

Issue #373, opened 2022-03-04
animation delay happens even when the action isn't performed


kurt22i:
![image](https://user-images.githubusercontent.com/94938654/156825316-6e434711-d0d7-46f0-8cf2-bee9bb41c74f.png)
https://gcsim.app/viewer/share/Fu8vLy8n8j9r0eZuvoPVc

PR #386, opened 2022-03-08
fix delay bug #373

srliao:
```eula char lvl=80/90 cons=0 talent=6,6,6;
eula add weapon="dullblade" refine=1 lvl=1/20;
eula add set="paleflame" count=4;

options mode=sl skill_delay=100;

target lvl=100 resist=.1;
active eula;

eula skill;
eula attack;
eula skill;
```
image

```
eula char lvl=80/90 cons=0 talent=6,6,6;
eula add weapon="dullblade" refine=1 lvl=1/20;
eula add set="paleflame" count=4;

options mode=apl skill_delay=100;

target lvl=100 resist=.1;
active eula;

eula skill;
eula attack;
```

image

### DIFF
--- a/pkg/simulation/run.go
+++ b/pkg/simulation/run.go
@@ -106,33 +106,29 @@ func (s *Simulation) AdvanceFrame() error {
 		s.queue = append(s.queue, next...)
 	}
 
-	if len(s.queue) > 0 {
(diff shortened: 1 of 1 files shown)


srliao:
With this change this check here really isn't needed anymore:

image

srliao:
fixed #373

PR #396, opened 2022-03-18
Fix mods not being tracked in log

shizukayuki:
the viewer will have to be fixed before this PR can be merged.

the mod handling changes were made to fix the following bug, and to make the mods code consistent.
image


srliao:
image
Looks like it works fine without changes to the viewer?

srliao:
played around with this some. looks like it's working fine to me. will merge

PR #397, opened 2022-03-18
2.6 artifacts implementation

HanSeoulOh:
these were unveiled in 2.6 livestream

### DIFF
--- a/internal/artifacts/echoes/echoes.go
+++ b/internal/artifacts/echoes/echoes.go
@@ -32,11 +32,13 @@ func New(c core.Character, s *core.Core, count int, params map[string]int) {
 			if s.ActiveChar != c.CharIndex() {
 				return false
 			}
-			if args[1].(*core.AttackEvent).Info.AttackTag != core.AttackTagNormal {
+			atk := args[1].(*core.AttackEvent)
+			if atk.Info.AttackTag != core.AttackTagNormal {
 				return false
 			}
-			atk := args[1].(*core.AttackEvent)
+
 			if s.Rand.Float64() < prob {
+				//TODO: need to check if this actually snapshots here
 				snap := c.Snapshot(&atk.Info)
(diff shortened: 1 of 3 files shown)


internal/artifacts/echoes/echoes.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if count >= 4 {
+		s.Events.Subscribe(core.OnAttackWillLand, func(args ...interface{}) bool {
+			if s.ActiveChar != c.CharIndex() {
+				return false
+			}
+			if args[1].(*core.AttackEvent).Info.AttackTag != core.AttackTagNormal {
+				return false
+			}
+			atk := args[1].(*core.AttackEvent)
```
srliao, line 38: this line should prob be moved above line 35. no point casting/doing type check twice

internal/artifacts/echoes/echoes.go
```diff
@@ -0,0 +1,57 @@
...
+			if s.Rand.Float64() < prob {
+				snap := c.Snapshot(&atk.Info)
```
srliao, line 40: i THINK it might be better to use the stats from the attack rather than creating another snapshot. we'll have to see when it goes live but i feel it's more likely to use existing stats than snapshot again.

internal/artifacts/echoes/echoes.go
```diff
@@ -0,0 +1,57 @@
...
+			Key: "echoes-2pc",
+			Amount: func() ([]float64, bool) {
+				return m, true
+			},
+			Expiry: -1,
+		})
+	}
+
+	if count >= 4 {
+		s.Events.Subscribe(core.OnAttackWillLand, func(args ...interface{}) bool {
```
srliao, line 31: description reads to me more like it's onDamage than onAttackWillLand

we'll see when it goes live I guess

srliao:
i see that the  other set is still missing on hurt events. i'll add those later maybe next week

Issue #412, opened 2022-03-25
Should add a way to track how many normal attack active character doing in apl mode


t3201v:
+if syntax maybe?

PR #417, opened 2022-03-29
Add counter

t3201v:
pr for #412

### DIFF
--- a/internal/tmpl/character/helper.go
+++ b/internal/tmpl/character/helper.go
@@ -1,7 +1,5 @@
 package character
 
-import "github.com/genshinsim/gcsim/pkg/core"
-
 //advance normal index, return the current one
 func (c *Tmpl) AdvanceNormalIndex() {
 	c.NormalCounter++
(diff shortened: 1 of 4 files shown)


srliao:
looks good. although i wonder if `.counter` is the best name for it. also i THINK this actually returns the next hit? not sure if that would result in undesired behaviour since say if you're in a 5 hit counter and you just did hit 5 (4 actually), i think this will return 0? is that the desired behaviour?

internal/tmpl/character/helper.go
```diff
@@ -1,9 +1,20 @@
...
 func (c *Tmpl) AdvanceNormalIndex() {
 	c.NormalCounter++
 	if c.NormalCounter == c.NormalHitNum {
 		c.NormalCounter = 0
 	}
 }
+
+// c.NormalCounter returns the next hit, and always resets
+// when it reaches the max hit of the normal attack combo
+func (c *Tmpl) NextNormalCounter() int {
```
srliao, line 15: sorry i'm a bit confused on this one. Isn't c.NormalCounter already the next normal index?

Issue #209, opened 2022-01-30
Comprehensive Frame Update


Tsym-or-Tysm:
## Introduction

Purpose of this ticket is to keep track the recount of all characters in accordance with gcsim's frame framework, which differs slightly from existing count in that we require recovery frame and cancellable frame for all possible combination of actions.

We're looking for help from everyone with both recording and frame counting. If you are able to contribute please leave a comment below. Looking for help both in terms of just providing raw footage as well as performing the count. It'd help us greatly if you can provide help with either or (or both should you desire).

## Recording requirement

- Record at consistent 60fps 720p minimum
- For each character, record all possible combinations of various ability, including but not limited to
  - string of normal attacks i.e. N1 -> N5 -> N1 (usually done by button mashing)
  - cancelling of each normal attack with charge attack i.e. N1 -> CA, N2 -> CA, etc...
  - cancelling of skill with every other possible action i.e. NA, CA, skill, burst, dash, jump swap
  - cancelling of burst with every other possible action i.e. NA, CA, skill, burst, dash, jump swap
  - cancelling of CA with every other possible action i.e. NA, CA, skill, burst, dash, jump swap
  - any other combinations for character with special abilities i.e. skill hold, etc...
- We'll need 3 trials of every combination to help average out any recording error
- All footage should be done vs empty air (not hitting trees) as sim is using no hitlag frames
- Footage should start clearly from standing/rest position and end at standing/rest position to avoid cutting off any part of the animation
- Footage should not be obstructed in anyways by trees, rocks, enemies, etc... 

## Frame count requirements

```
NX/CA hitmark / NX/CA hitmark when infused:
start frame: start of the animation of NX
end frame: hitmark registered on target = 2 frames after the frame where the yellow mark first appears 

E hitmark / Q hitmark / D hitmark (only if special dash):
start frame: first frame the buttons bottom right grey out
end frame: frame the element gets applied (Pyro/Hydro/Electro/Dendro?/Cryo) / the first frame the Swirl (Anemo) or Crystallize (Geo) text appears

NX -> NY:
start frame: start of the animation of NX
end frame: start of the animation of NY

NX -> CA:
start frame: start of the animation of NX
end frame: first frame the buttons bottom right grey out

CA/E/Q -> E/Q / D -> E/Q (only if special dash):
start frame: first frame the buttons bottom right grey out
end frame: first frame the buttons bottom right grey out for the next action

[...]

PR #425, opened 2022-04-03
ayaka frames update

kurt22i:
- Implements recounted ayaka frames
- Frames taken from [this spreadsheet](https://docs.google.com/spreadsheets/d/1WPxQ7vf0_dp5g96aqGF0_JZSqjtYHOJC0b89zRS_WrA/edit?usp=sharing), and were counted by me from @k0l11's [video](https://www.youtube.com/watch?v=ulTSQRrR4n8)
- Could use a few more tests to make sure it's implemented properly, but has worked fine in the tests I've run so far
- Tested here: https://gcsim.app/viewer/share/perm_6D0Pnrw-HG_l-I8_Vq88H

### DIFF
--- a/internal/characters/ayaka/abil.go
+++ b/internal/characters/ayaka/abil.go
@@ -147,7 +147,7 @@ func (c *char) Skill(p map[string]int) (int, int) {
 		},
 	})
 
-	c.Core.Combat.QueueAttack(ai, core.NewDefCircHit(4, false, core.TargettableEnemy), 0, 38)
+	c.Core.Combat.QueueAttack(ai, core.NewDefCircHit(4, false, core.TargettableEnemy), 0, 33)
 
 	c.SetCD(core.ActionSkill, 600)
 	return f, a
--- a/internal/characters/ayaka/abil.go
+++ b/internal/characters/ayaka/abil.go
@@ -6,6 +6,8 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core"
 )
 
+var hitmarks = [][]int{{8}, {10}, {16}, {8, 15, 22}, {27}}
+
 func (c *char) Attack(p map[string]int) (int, int) {
 
 	f, a := c.ActionFrames(core.ActionAttack, p)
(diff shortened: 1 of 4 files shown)


k0l11:
I'm sure there isn't too much variance due to the macros, but the sheet doesn't have the numbers for all 3 trials on some cancels. Please finish counting all the shown cancels.
Also I'm not sure how you arrived at the numbers for N1 -> CA, I got this (this is just an example, I'm checking all of the frame counts rn):
1170-1147 = 23
1491-1468 = 23
1790-1766 = 24 
For easier verification I think it would also be useful to put these formulas into the sheet instead of just the raw frame number (that's why I added frame numbers to the video).
I think I'll recount these frames myself to cross-check.

kurt22i:
> 

Given how consistent the macros are, I honestly don't think it's necessary to count all 3 trials. It's good to have the data for future reference, but with macros, 2 trials is plenty sufficient for common combos and 1 for the others. Note that without macros I'm not sure if even 3 trials is enough.
As for the formulas, I will definitely add them in my future counts, it's something I should've been doing and should make it a lot easier to see how I counted. I have been using <start frame - endframe>, without a +1, since it takes 1 frame to queue the action in the sim anyway.

k0l11:
I'm of the opinion that if the cancel isn't redundant (can't be "canceled at any point"), then all 3 trials should be counted. 
I've also taken a look at @Kisa75 's counts and they seem to be accurate at first glance. There are some significant differences between the frames of their sheet and yours. Since both of you just put in the raw frame number it's hard to check the counts against the video but I believe Kisa's are more accurate.

kurt22i:
I have pushed an update to fix the erroneous N1->CA and Burst hitmark frames. With this, all frames should be correct.

As great as it would be to have counts for all 3 trials, I simply don't think it's a good use of time to triple-count macros. I would rather spend the time counting frames for other characters than increasing accuracy for one char from 99% to 100%. Of course, these counts would be good to have eventually and after all chars have baseline frames it wouldn't be a bad idea to go back and count the rest. Also all of this is just my opinion, if you want to count all 3 don't let me discourage you

k0l11:
I'm not the one calling the shots, but I think counting all 3 trials once in order to have it done "forever" is better than having "someone come back to it eventually" to count the other 2 trials. The accuracy argument seems fair, but then I have to question why the issue even requires 3 trials per cancel then. @srliao thoughts?

srliao:
> I'm not the one calling the shots, but I think counting all 3 trials once in order to have it done "forever" is better than having "someone come back to it eventually" to count the other 2 trials. The accuracy argument seems fair, but then I have to question why the issue #209 even requires 3 trials per cancel then. @srliao thoughts?

@kurt22i I actually agree with koli. Realistically we want even more than just 3 trials but 3 was a bare minimum we settled one. Even though macro seems fairly consistent you still want data backing that up. In addition, since we're already undertaking this much work, might as well do it correctly rather than having to come back later, which to be realistic is not going to happen.

In this particular case didn't Kisa also count all 3? Why not just use his third count?

kurt22i:
@k0l11 @srliao ok, fair enough. For Ayaka, I'll simply merge in Kisa's counts for trial 3 and some of trial 2 (actually only hadn't done this bc i couldnt find his counts, but i found them now). For future characters, I'll be sure to use 3 trials as well.. and actually, given how counting is generally faster than recording, I think my concerns about it slowing things down might have been unfounded anyway.

PR #453, opened 2022-04-10
Move Yoimiya's skill buff effect

jordanlovato:
* Previously Yoi's snapshot routine was apply the effect of her skill multiplier. This was causing additional talent multiplier in the calculation of her damage when multiple snapshots are made of the same attack.

### DIFF
--- a/internal/characters/yoimiya/abil.go
+++ b/internal/characters/yoimiya/abil.go
@@ -31,14 +31,14 @@ func (c *char) Attack(p map[string]int) (int, int) {
 		if c.Core.Status.Duration("yoimiyaskill") > 0 {
 			// ds.ICDTag = core.ICDTagNone
 			//multiplier
-			ai.Element = core.Pyro
+
 			ai.Mult = skill[c.TalentLvlSkill()] * mult[c.TalentLvlAttack()]
 			c.Core.Log.NewEvent("skill mult applied", core.LogCharacterEvent, c.Index, "prev", mult[c.TalentLvlAttack()], "next", ai.Mult, "char", c.Index)
 		} else {
 			ai.Mult = mult[c.TalentLvlAttack()]
 		}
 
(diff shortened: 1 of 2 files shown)


internal/characters/yoimiya/abil.go
```diff
@@ -27,15 +27,26 @@ func (c *char) Attack(p map[string]int) (int, int) {
 	}
 
 	for i, mult := range attack[c.NormalCounter] {
-		ai.Mult = mult[c.TalentLvlAttack()]
-		totalMV += mult[c.TalentLvlAttack()]
+		//infusion to normal attack only
+		if c.Core.Status.Duration("yoimiyaskill") > 0 {
```
srliao, line 31: L31 to L36 should stay in the snapshot function for readability. There's no real advantage/reason to split up the code here between Snapshot and Attack (unlike some other characters like Raiden which has a lot more complexity)

jordanlovato: @srliao Sounds good, will revert. I'm going to add a condition to the snapshot to prevent a bug where multiple calls to snapshot for a single attack was increasing the talent multiplier of the attack unexpectedly: https://discord.com/channels/845087716541595668/884485336657444894/962749496180244570

internal/characters/yoimiya/yoimiya.go
```diff
@@ -11,8 +11,9 @@ func init() {
 
 type char struct {
 	*character.Tmpl
-	a2stack  int
-	lastPart int
+	a2stack                     int
+	lastPart                    int
+	attacksWithSkillMultApplied map[*core.AttackInfo]bool
```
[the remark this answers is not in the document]
jordanlovato: Oh! I did not realize that we don't support multiple calls to `Snapshot`, that makes things much simpler, I will remove the logic

PR #454, opened 2022-04-10
Tweak Echoes ICD

jordanlovato:
* Fix icd on echoes set (previously could proc multiple times within 0.2s if you were lucky enough.

### DIFF
--- a/internal/artifacts/echoes/echoes.go
+++ b/internal/artifacts/echoes/echoes.go
@@ -14,7 +14,8 @@ func init() {
 
 func New(c core.Character, s *core.Core, count int, params map[string]int) {
 	prob := 0.36
-	probicd := 0
+	icd := 0
+
 	if count >= 2 {
 		m := make([]float64, core.EndStatType)
 		m[core.ATKP] = 0.18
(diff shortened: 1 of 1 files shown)


jordanlovato:
I believe this is a working test: https://gcsim.app/viewer/share/37wTeuUdu03fk0r7dbakx

internal/artifacts/echoes/echoes.go
```diff
@@ -27,35 +35,62 @@ func New(c core.Character, s *core.Core, count int, params map[string]int) {
...
-				prob = 0.36
-			} else {
-				if s.F > probicd {
-					prob += 0.2
-					probicd = s.F + 0.2*60
-				}
-				if prob > 1 {
-					prob = 1
-				}
 			}
```
srliao, line 67: this is missing a `return false`

internal/artifacts/echoes/echoes.go
```diff
@@ -27,35 +35,62 @@ func New(c core.Character, s *core.Core, count int, params map[string]int) {
...
 			if atk.Info.AttackTag != core.AttackTagNormal {
 				return false
 			}
 
-			if s.Rand.Float64() < prob {
-				//TODO: need to check if this actually snapshots here
-				// snap := c.Snapshot(&atk.Info)
-				dmgAdded := (atk.Snapshot.BaseAtk*(1+atk.Snapshot.Stats[core.ATKP]) + atk.Snapshot.Stats[core.ATK]) * 0.7
+			// if buff is already active then buff attack
+			if s.F < procExpireF {
```
srliao, line 60: there should be a `return false` inside this if statement at the end

PR #456, opened 2022-04-11
haran fix

t3201v:
Stacks, hook and refinement don't work properly.
Idk if recasting skill will change the stack snapshotted.

### DIFF
--- a/internal/weapons/sword/haran/haran.go
+++ b/internal/weapons/sword/haran/haran.go
@@ -25,7 +25,6 @@ func weapon(char core.Character, c *core.Core, r int, param map[string]int) stri
 	wavespikeICD := 0
 	wavespikeStacks := 0
 	maxWavespikeStacks := 2
-	stackSnapshot := 0
 
 	char.AddMod(core.CharStatMod{
 		Key: "haran-ele-bonus",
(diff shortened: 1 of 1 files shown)


HanSeoulOh:
can you explain what you mean by stack snapshot? How does it work? I don't think anyone on the team has Haran. Otherwise things look pretty good to me

t3201v:
> can you explain what you mean by stack snapshot? How does it work? I don't think anyone on the team has Haran. Otherwise things look pretty good to me

The add pre dmg function always refers to wavespikeStacks again every time it's called. Hence it will use 0 bc the stack has been reset to 0.

internal/weapons/sword/haran/haran.go
```diff
@@ -34,33 +35,36 @@ func weapon(char core.Character, c *core.Core, r int, param map[string]int) stri
...
+		stackSnapshot = wavespikeStacks
 		char.AddPreDamageMod(core.PreDamageMod{
 			Key:    "ripping-upheaval",
 			Expiry: c.F + 60*8,
 			Amount: func(atk *core.AttackEvent, t core.Target) ([]float64, bool) {
 				if atk.Info.AttackTag != core.AttackTagNormal {
 					return nil, false
 				}
-				val[core.DmgP] = (0.2 * float64(r) * 0.04) * float64(wavespikeStacks)
+				val[core.DmgP] = (0.15 + float64(r)*0.05) * float64(stackSnapshot)
```
srliao, line 67: this line should be after L59 (actually there's no reason for `stackSnapshot` variable). no point redoing the calculation every single time the functiong ets called

srliao (approved):

PR #457, opened 2022-04-11
zhongli stele tick fix

HanSeoulOh:
zhongli steles should only make an extra tick for each extra non-stele construct.

### DIFF
--- a/internal/characters/zhongli/stele.go
+++ b/internal/characters/zhongli/stele.go
@@ -78,7 +78,7 @@ func (c *char) resonance(src, max int) func() {
 		ae := c.steleSnapshot
 
 		//check how many times to hit
-		count := c.Core.Constructs.Count()
+		count := c.Core.Constructs.Count() - c.Core.Constructs.CountByType(core.GeoConstructZhongliSkill) + 1
 		if count > max {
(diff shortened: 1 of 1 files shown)


internal/characters/zhongli/stele.go
```diff
@@ -78,7 +78,7 @@ func (c *char) resonance(src, max int) func() {
 		ae := c.steleSnapshot
 
 		//check how many times to hit
-		count := c.Core.Constructs.Count()
+		count := c.Core.Constructs.Count() - c.Core.Constructs.CountByType(core.GeoConstructZhongliSkill) + 1
```
srliao, line 81: in theory `c.Core.Constructs.CountByType(core.GeoConstructZhongliSkill)` will never be <0 but just in case i would prefer something like

```go
count := c.Core.Constructs.CountByType(core.GeoConstructZhongliSkill)
if count  > 1 {
   count  = 1
}
count += c.Core.Constructs.Count()
```

HanSeoulOh: ok done btw, you can make the suggestions directly in github in the future so someone can just accept your suggestions

PR #462, opened 2022-04-12
Begin Substat Optimizer Cleanup

jordanlovato:
* Start splitting code for the optimizer routine into files (logger.go, regex.go, stats.go)
* Organize sections of config init code, start moving little pieces into smaller function units


jordanlovato:
Opening this draft PR to make it easier to feedback into the cleanup process, so that we can make sure that the organization makes sense for everyone

internal/substatoptimizer/regex.go
```diff
@@ -0,0 +1,67 @@
...
+	re.Options = regexp.MustCompile(`([a-z_]+)=([0-9.]+)`)
+
+	return &re
+}
+
+func (re *OptimRegex) scrubSimCfg(cfg string, sugarLog *zap.SugaredLogger) string {
+	if len(re.Mainstats.FindAllString(cfg, -1)) != len(re.GetCharNames.FindAllString(cfg, -1)) {
+		sugarLog.Error("Error: Could not identify valid main artifact stat rows for all characters based on flower HP values.")
+		sugarLog.Error("5* flowers must have 4780 HP, and 4* flowers must have 3571 HP.")
+		os.Exit(1)
```
srliao, line 37: I have to say.. the original usage of os.Exist(1) is not very Go like lol. Should really be returning an error in the function and handling that somehow (including exiting)

srliao:
So I see here: https://github.com/jordanlovato/gcsim/blob/ec39d14241e4808dbd2cbe4b881cfb91a610151e/internal/substatoptimizer/substatoptimizer.go#L75 you are trying to handle the error.

The better way to handle it is something like this:

```go
func RunSubstatOptim(simopt simulator.Options, verbose bool, additionalOptions string) error {
	
        //do stuff here

	cfg, err := simulator.ReadConfig(simopt.ConfigPath)
	if err != nil {
		return err
	}

	srcCleaned, err := re.scrubSimCfg(cfg)

        if err != nil {
                  if err == ErrNotBigDealInsertDescriptiveName {
                            //do stuff
                  } else {
                             return err
                   }
        }
```

Then somewhere you would have to declare a package level variable like so

```go
var ErrNotBigDealInsertDescriptiveName  = errors.New("some description goes here")
```

You'll see this kind of design pattern for example in handling sql error (http://go-database-sql.org/errors.html) because sometimes you do expect errors i.e. expecting zero row result

Not a big deal. I can always fix it later but thought I'd point it out.

jordanlovato:
@srliao noted TY! I'll fix this now. 

Note that this is exactly the feedback I'm looking for. I have a sense of how to break down large routines into readable code, but I'm not as knowledgeable on Golang design patterns. If you see other funky stuff that could be resolved in a simpler way, lmk!

jordanlovato:
Pushing all of my testing here: https://docs.google.com/spreadsheets/d/1WmyzuHnICg2qxyX8grfsCZEx3ec56Yvc7MYpyEbQ4S0/edit?usp=sharing

internal/substatoptimizer/logger.go
```diff
@@ -0,0 +1,25 @@
+package substatoptimizer
```
srliao, line 1: Given that most of the code have moved to `pkg/optimization` i think it's fine/better to move this package over as well. Or rather, if anything I think function *might* make more sense in the executable code since really this is just setting up and calling the optimization code

pkg/optimization/regex.go
```diff
@@ -0,0 +1,68 @@
...
+
+type OptimRegex struct {
+	Mainstats      *regexp.Regexp
+	GetCharNames   *regexp.Regexp
+	Substats       *regexp.Regexp
+	Options        *regexp.Regexp
+	InsertLocation *regexp.Regexp
+}
+
+var InvalidStats = errors.New("Error: Could not identify valid main artifact stat rows for all characters based on flower HP values.\n5* flowers must have 4780 HP, and 4* flowers must have 3571 HP.")
```
srliao, line 26: error messages should really just be a short sentence, all lower case (no new line). for example i would have just wrote for this one `var ErrInvalidStats = errors.New("unidentifiable main stats")`

also i believe the convention in go code is to name these variables starting with `Err`

pkg/optimization/regex.go
```diff
@@ -0,0 +1,68 @@
...
+
+func ReplaceSimOutputForChar(char, src, out string) string {
```
srliao, line 28: this looks like a helper function. does this function need to be exported?

pkg/optimization/regex.go
```diff
@@ -0,0 +1,68 @@
...
+		if REGEXP_LINE_MAINSTAT.MatchString(match) {
+			continue
+		}
+
+		clean = strings.Replace(clean, match, "", -1)
+		errorPrinted = true
+	}
+
+	if errorPrinted {
+		return clean, errors.New("Warning: Config found to have existing substat information. Ignoring...")
```
srliao, line 50: similar to what i mentioned above re the error message

pkg/optimization/regex.go
```diff
@@ -0,0 +1,68 @@
...
+	return clean, nil
+}
+
+func ParseOptimizerCfg(additionalOptions string, optionsMap map[string]float64) (map[string]float64, error) {
+	parsedOptions := REGEXP_LINE_OPTIONS.FindAllStringSubmatch(additionalOptions, -1)
+	for _, val := range parsedOptions {
+		if _, ok := optionsMap[val[1]]; ok {
+			optionsMap[val[1]], _ = strconv.ParseFloat(val[2], 64)
+		} else {
+			err := errors.New(fmt.Sprintf("Invalid substat optimization option found: %v: %v", val[1], val[2]))
```
srliao, line 62: same as above

also there's a `fmt.Errorf` if you need to use fmt to insert some values etc..

pkg/optimization/slice.go
```diff
@@ -0,0 +1,45 @@
...
+		idx:   make([]int, len(n)),
+	}
+	for i := range s.idx {
+		s.idx[i] = i
+	}
+	return s
+}
+
+// Gets the minimum of a slice of integers
+func MinInt(vars ...int) int {
```
srliao, line 36: not sure if this function needs to be exported

pkg/optimization/substat_details.go
```diff
@@ -0,0 +1,551 @@
+package optimization
```
srliao, line 1: i think this file might be more appropriately named `details.go`. in most cases you only see underscore for test files such as `logger_test.go` etc...

although i have seen some packages that used underscore to help split up files (and I have contemplated doing the same as well) where the package was too large. but in this case i think it's not necessary

pkg/optimization/substats.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"github.com/genshinsim/gcsim/internal/simulator"
+	"github.com/genshinsim/gcsim/pkg/core"
+	"go.uber.org/zap"
+)
+
+type SubstatOptimizer struct {
+	logger     *zap.SugaredLogger
+	optionsMap map[string]float64
+	Details    *SubstatOptimizerDetails
```
srliao, line 13: i don't think it's necessary to have a pointer reference here as the methods are already pointer methods

pkg/optimization/substats.go
```diff
@@ -0,0 +1,92 @@
...
+	fourStarFound := o.Details.setStatLimits()
+	if fourStarFound {
+		o.logger.Warn("Warning: 4* artifact set detected. Optimizer currently assumes that ER substats take 5* values, and all other substats take 4* values.")
+	}
+
+	fixedSubstatCount := o.optionsMap["fixed_substats_count"]
+	o.Details.setInitialSubstats(fixedSubstatCount)
+	o.logger.Info("Starting ER Optimization...")
+
+	o.Details.setCharProfilesCopy(o.Details.charProfilesERBaseline)
```
srliao, line 46: this feels kinda weird to me as `setCharProfileCopy` would already have access to `o.Details` so why does it need to be passed again?

pkg/optimization/substat_details.go
```diff
@@ -0,0 +1,551 @@
...
+import (
+	"fmt"
+	"github.com/genshinsim/gcsim/internal/simulator"
+	"github.com/genshinsim/gcsim/pkg/core"
+	"math"
+	"sort"
+	"strings"
+)
+
+type SubstatOptimizerDetails struct {
```
srliao, line 12: design wise i think this structure feels a bit icky. at a cursory glance i feel like the structure should be more like soemthing like the following

```
type Optimizer struct {
   Input SomeStruct
   WIP SomeStruct
}
```

where `Input` is your starting point and contains all the relevant information like the config, initial profile etc... and `WIP` (could be a slice here) is keeping track of each iteration where you try different stats to get to the optimized state

this current design/data structure feels a bit like a big dumb of different variables not organized in any fashion

pkg/optimization/substat_details.go
```diff
@@ -0,0 +1,551 @@
...
+}
+
+// Calculate per-character per-substat "gradients" at initial state using finite differences
+// In practical evaluations, adding small numbers of substats (<10) can be VERY noisy
+// Therefore, "gradient" evaluations are done in groups of 10 substats
+// Allocation strategy is to just max substats according to highest gradient to lowest
+// TODO: Probably want to refactor to potentially run gradient step at least twice:
+// once initially then another at 10 assigned liquid substats
+// Fine grained evaluations are too expensive time wise, but can perhaps add in an option for people who want to sit around for a while
+func (stats *SubstatOptimizerDetails) optimizeNonERSubstats() []string {
```
srliao, line 38: if i understand correctly this is return the debug msgs. is there a reason why we're returning this instead of just printing it out?

pkg/optimization/substat_details.go
```diff
@@ -0,0 +1,551 @@
...
+
+		if erStack == 0 {
+			initialMean = result.DPS.Mean
+			initialSD = result.DPS.SD
+		}
+
+		condition := result.DPS.Mean/initialMean-1 < -tolMean || result.DPS.SD/initialSD-1 > tolSD
+		// For Raiden, we can't use DPS directly as a measure since she scales off of her own ER
+		// Instead we ONLY use the SD tolerance as big jumps indicate the rotation is becoming more unstable
+		if char.Base.Key == core.Raiden {
```
srliao, line 352: i think this code here is more readable if rewritten as:
```go
if char.Base.Key == core.Raiden && result.DPS.SD/initialSD-1 > tolSD {
   break
}
```

as opposed to assigning to a variable condition, then checking if condition is true

pkg/optimization/substat_details.go
srliao: In addition, for both this function as well as optimizeER function. i feel like it'd be more logical if these functions returns optimal result given starting point (+ whatever other params). so a bit more "pure" functional if you will (not sure what the right wording is here)

srliao:
closing as this have been ported in https://github.com/srliao/gcsim/pull/17

PR #465, opened 2022-04-13
Change Albedo and Ningguang Blossoms/Gems to use skill snapshot

shizukayuki:
- Fix kazuha swirling his absorb dot

### DIFF
--- a/internal/tmpl/character/tmpl.go
+++ b/internal/tmpl/character/tmpl.go
@@ -168,22 +168,17 @@ func (c *Tmpl) AddPreDamageMod(mod core.PreDamageMod) {
 }
 
 func (c *Tmpl) DeletePreDamageMod(key string) {
-	ind := -1
-	for i, v := range c.PreDamageMods {
+	n := 0
+	for _, v := range c.PreDamageMods {
 		if v.Key == key {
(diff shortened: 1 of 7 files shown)


internal/tmpl/character/tmpl.go
```diff
@@ -167,6 +167,25 @@ func (c *Tmpl) AddPreDamageMod(mod core.PreDamageMod) {
...
+		}
+	}
+	// return if mod does not exists
+	if ind == -1 {
+		return
+	}
+
+	// remove mod
+	c.PreDamageMods[ind].Event.SetEnded(c.Core.F)
+	c.PreDamageMods[ind] = c.PreDamageMods[len(c.PreDamageMods)-1]
```
srliao, line 184: this would swap the last mod and the current mod right? wouldn't it be better to keep the ordering here?

something like this: https://github.com/genshinsim/gcsim/blob/460f90d8e46903b950587f0859b42910040b4ab0/internal/tmpl/construct/construct.go#L206

shizukayuki: yes though i don't think order matter for mods.. i guess it does for logs.

PR #475, opened 2022-04-16
Added skill_issue mode for bennett burst

IvanHdzF:
LIST OF CHANGES

1. recoded every single ability in sim to match in game hitbox
2. recoded parser to allow for setting target position
3. recoded parser to allow for changing default attack center point
4. recodedparser to allow for moving targets around
5. recoded queue and abil execute to handle that logic
6. made sure every abil in sim is respecting default attack center point

Just kidding, skill issue=0 is the same as current burst, =1 makes it hit 1 enemy only, =2 skips everything, trial and tested
https://gcsim.app/viewer/share/y-peJYe64DWofCPwJ06Bb 
https://gcsim.app/viewer/share/-RazN_MpUpdausi8WX5mP
https://gcsim.app/viewer/share/aRg45Ac51XqcVqNutXFwk

### DIFF
--- a/internal/characters/bennett/bennett.go
+++ b/internal/characters/bennett/bennett.go
@@ -249,7 +249,15 @@ func (c *char) Burst(p map[string]int) (int, int) {
 		Mult:       burst[c.TalentLvlBurst()],
 	}
 	//TODO: review bennett AOE size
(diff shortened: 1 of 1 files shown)


srliao:
closing in favour of proper targetting implementation.

PR #476, opened 2022-04-16
hu tao frames update

kurt22i:
- Updates Hu Tao frames
- Sheet: https://docs.google.com/spreadsheets/d/1azK_EOOxyNUPlowg6qY9VeINTNwS2cPQoPcnd9UqFUo/edit?usp=sharing
- Video: https://youtu.be/NNjnp2l84ng
- Tested here: https://gcsim.app/viewer/share/perm_0pgVGR3Pz-Qa1gHsMzaQI

### DIFF
--- a/internal/characters/ayato/frames.go
+++ b/internal/characters/ayato/frames.go
@@ -91,16 +91,7 @@ func (c *char) ActionInterruptableDelay(next core.ActionType, p map[string]int) 
 		c.Core.Status.Duration("soukaikanka") > 0 {
 		f := 23 - 5
 		f = int(float64(f) / (1 + c.Stat(core.AtkSpd)))
-
-		//this is necessary to prevent double animation delay when his E expires
-		if f > c.Core.Status.Duration("soukaikanka") {
-			c.skillExpired = true
-		}
 		return f
-	} else if c.skillExpired {
-		//if his E has just expired, do not do the animation delay again
-		c.skillExpired = false
-		return 0
 	}
 	//otherise use default implementation
 	return c.Tmpl.ActionInterruptableDelay(next, p)
--- a/internal/characters/ayato/frames.go
+++ b/internal/characters/ayato/frames.go
@@ -91,7 +91,16 @@ func (c *char) ActionInterruptableDelay(next core.ActionType, p map[string]int) 
 		c.Core.Status.Duration("soukaikanka") > 0 {
 		f := 23 - 5
 		f = int(float64(f) / (1 + c.Stat(core.AtkSpd)))
+
+		//this is necessary to prevent double animation delay when his E expires
+		if f > c.Core.Status.Duration("soukaikanka") {
+			c.skillExpired = true
+		}
 		return f
+	} else if c.skillExpired {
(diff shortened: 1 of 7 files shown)


internal/characters/ayato/frames.go
```diff
@@ -91,7 +91,16 @@ func (c *char) ActionInterruptableDelay(next core.ActionType, p map[string]int)
 		c.Core.Status.Duration("soukaikanka") > 0 {
 		f := 23 - 5
 		f = int(float64(f) / (1 + c.Stat(core.AtkSpd)))
+
```
srliao, line 94: Can you split out any non Hutao frame related changes into a separate PR?

internal/characters/hutao/abil.go
```diff
@@ -42,8 +42,8 @@ func (c *char) ChargeAttack(p map[string]int) (int, int) {
 		//trying to change animations mid-attack, but not sure how to fully test that
 		//[4:41 PM] jstern25| ₼WHO_SUPREMACY: this mostly checks out
 		//her e can't expire during q as well
-		if f > c.Core.Status.Duration("paramita") {
-			c.Core.Status.AddStatus("paramita", f)
+		if 6 > c.Core.Status.Duration("paramita") {
```
srliao, line 45: This shouldn't be a hard coded number here. I'm not sure what 6 is for.

internal/characters/hutao/frames.go
```diff
@@ -2,36 +2,117 @@ package hutao
...
-
-framecounts for 7 attempts of N2CJ (no hitlag):
-83, 85, 88, 89, 77, 82, 84
-
-first 3 not from the uploaded recording (as a n1cd player i cud barely pull it off :monkaS: )
-YouTube
-**/
-
-//var normalFrames = []int{13, 16, 25, 36, 44, 39}               // from kqm lib
-var normalFrames = []int{10, 13, 22, 33, 41, 36} // from kqm lib, -3 for hit lag
```
srliao, line 17: I think you should keep frames in here actually instead of moving it into `abil.go`. I know other files have done otherwise but this is a more sensible grouping.

internal/characters/hutao/frames.go
```diff
@@ -2,36 +2,117 @@ package hutao
...
+		case core.ActionSwap:
+			f = 42 - 2
+		}
+
+		//this is necessary to prevent double animation delay when paramita expires
+		if f > c.Core.Status.Duration("paramita") {
+			c.paramitaExpired = true
+		}
+		return f
+	} else if c.paramitaExpired {
```
srliao, line 111: Not sure what is going on here to be honest. Can you explain this change to me?

srliao (changes requested):

internal/characters/hutao/frames.go
kurt22i: This fixes the bug we found with Ayato, where when a status that changes animation delay runs out, two delays are played

the other option is to try to change the core code that handles this, but im not sure if its possible without a refactor (havent checked tho bc idk where it is lol)

srliao (approved):

Issue #474, opened 2022-04-16
substat optimizer gives 12 er subs to characters with er sands


kurt22i:
[input.txt](https://github.com/genshinsim/gcsim/files/8499060/config.txt)
[output.txt](https://github.com/genshinsim/gcsim/files/8499061/output2.txt)
Kokomi was given max ER subs even though this is not allowed.

PR #479, opened 2022-04-16
fix substat optimizer

kurt22i:
Fixes https://github.com/genshinsim/gcsim/issues/474


jordanlovato:
@kurt22i @srliao yup, looks good, this should use the appropriately calculated substat caps per stat type per character (and factoring mainstats into the calculation)

jordanlovato:
@kurt22i Do you happen to have tests/configurations for this PR (https://github.com/genshinsim/gcsim/pull/479) that either you or I could rerun under the build proposed in https://github.com/genshinsim/gcsim/pull/462

It would be nice to

1. Double check that the patch was properly merged
2. Check that we're seeing consistent-ish behavior between the prod build and this PR
Lmk! If not that's okay!

PR #480, opened 2022-04-17
add pp slide

kurt22i:
tested here: https://gcsim.app/viewer/share/perm_2jOHsz3OFCXPAqbWYngr9

### DIFF
--- a/internal/characters/hutao/abil.go
+++ b/internal/characters/hutao/abil.go
@@ -48,7 +48,7 @@ func (c *char) ChargeAttack(p map[string]int) (int, int) {
 		}
 
 		c.ppSlide = true
-		c.AddTask(func() { c.ppSlide = false }, 44)
+		c.AddTask(func() { c.ppSlide = false }, "ppslide", 44)
 		c.applyBB()
 		//charge land 182, tick 432, charge 632, tick 675
 		//charge land 250, tick 501, charge 712, tick 748
--- a/internal/characters/hutao/abil.go
+++ b/internal/characters/hutao/abil.go
@@ -47,6 +47,8 @@ func (c *char) ChargeAttack(p map[string]int) (int, int) {
 			// c.S.Status["paramita"] = c.Core.F + f //extend this to barely cover the burst
 		}
 
+		c.ppSlide = true
+		c.AddTask(func() { c.ppSlide = false }, 44)
 		c.applyBB()
 		//charge land 182, tick 432, charge 632, tick 675
 		//charge land 250, tick 501, charge 712, tick 748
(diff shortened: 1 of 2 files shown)


srliao:
I know I was the one that mentioned some sort of temporary variable but looking at it I'm not so sure I like this solution?

I think I would prefer, rather than adding to paramita status here: https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/abil.go#L39

Something like the following:

```go
func (c *char) ChargeAttack(p map[string]int) (int, int) {

	f, a := c.ActionFrames(core.ActionCharge, p)

        slide := 100 //whatever the slide duration is

	if c.Core.Status.Duration("paramita") > 0 {
		if f > c.Core.Status.Duration("paramita") {
                        c.Core.Status.AddStatus("pp-slide", slide)
		}
		c.applyBB()
	}

        // rest of charge attack
```

Then in the Snapshot function: https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/hutao.go#L126

Here should add a check for if `pp-slide` is active and ability used is charged or burst to also apply the buff

Ideally `slide` should get the number directly from `ActionInterruptableDelay(next core.ActionType, p map[string]int)`. Trouble this is this function as is only current take the `next` action as a param. Really this function should be refactor to take previous, next so that it can be called anytime in general where you want to know the cancelled frames

In general I'm not really a fan of using AddTask all over. Also this is cheaper anyways than creating a function, appending it to an array, and have the function called back later.

kurt22i:
> Then in the Snapshot function:
> 
> https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/hutao.go#L126
> 
> Here should add a check for if `pp-slide` is active and ability used is charged or burst to also apply the buff

Won't this not do anything since Hu Tao's burst doesn't snapshot?

srliao:
> > Then in the Snapshot function:
> > https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/hutao.go#L126
> > 
> > Here should add a check for if `pp-slide` is active and ability used is charged or burst to also apply the buff
> 
> ~Won't this not do anything since Hu Tao's burst doesn't snapshot?~ Actually maybe it does because it seems it technically snapshots on the same frame it does dmg?
> 
> In that case, what should I be putting in the snapshot function?

What do you mean by this? Everything has to call the `Snapshot` function in order to grab the stats? It's just a matter of extending the custom `Snapshot` function by long enough to cover the burst

Note that I have doubts about the timing on this anyways: https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/abil.go#L235

kurt22i:
> > > Then in the Snapshot function:
> > > https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/hutao.go#L126
> > > 
> > > Here should add a check for if `pp-slide` is active and ability used is charged or burst to also apply the buff
> > 
> > 
> > ~Won't this not do anything since Hu Tao's burst doesn't snapshot?~ Actually maybe it does because it seems it technically snapshots on the same frame it does dmg?
> > In that case, what should I be putting in the snapshot function?
> 
> What do you mean by this? Everything has to call the `Snapshot` function in order to grab the stats? It's just a matter of extending the custom `Snapshot` function by long enough to cover the burst
> 
> Note that I have doubts about the timing on this anyways:
> 
> https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/internal/characters/hutao/abil.go#L235

image
Is this what you mean? I don't think it is, I feel like I'm missing something here

srliao:
still not sure about this one. not implemented in rewrite for now. we'll see if it causes any issues

PR #481, opened 2022-04-17
Adjust swap action to match swap_delay and fix skill hold cancels

shizukayuki:
- Check for skill hold on cancel delays
- Fix swap event not getting delayed by swap_delay
- Fix xiangling C1 check


shizukayuki:
closing in favor of #490

PR #483, opened 2022-04-17
adjust substat optimizer to reliably give more cr to fav weapon users

k0l11:
In some configs the substat optimizer fails to give Favonius weapon users enough CR. This fix increases the gradient for CR so that the optimizer picks it instead of ATK% (+1000 seemed like a good number because that makes CR usually end up being chosen after CD (gradient is usually > 1000) but before ATK% (gradient is usually < 1000).
The main reason for this fix is that EM/EM/EM Fav Kazuha sometimes gets max'd out ATK% subs instead of crit rate subs leading to less dps and a higher standard deviation.

### DIFF
--- a/internal/substatoptimizer/substatoptimizer.go
+++ b/internal/substatoptimizer/substatoptimizer.go
@@ -391,9 +391,9 @@ func RunSubstatOptim(simopt simulator.Options, verbose bool, additionalOptions s
 
 			// fixes cases in which fav holders don't get enough crit rate to reliably proc fav (an important example would be fav kazuha)
 			// might give them "too much" cr (= max out liquid cr subs) but that's probably not a big deal
-			if charWithFavonius[idxChar] && idxSubstat == 1 {
+			if charWithFavonius[idxChar] && substat == core.CR {
 				substatGradients[idxSubstat] += 1000
-			}	
+			}
 
 			charProfilesCopy[idxChar].Stats[substat] -= 10 * substatValues[substat] * charSubstatRarityMod[idxChar]
 		}
--- a/internal/substatoptimizer/substatoptimizer.go
+++ b/internal/substatoptimizer/substatoptimizer.go
@@ -389,6 +389,12 @@ func RunSubstatOptim(simopt simulator.Options, verbose bool, additionalOptions s
 
 			substatGradients[idxSubstat] = substatEvalResult.DPS.Mean - initialMean
 
+			// fixes cases in which fav holders don't get enough crit rate to reliably proc fav (an important example would be fav kazuha)
+			// might give them "too much" cr (= max out liquid cr subs) but that's probably not a big deal
+			if charWithFavonius[idxChar] && idxSubstat == 1 {
+				substatGradients[idxSubstat] += 1000
+			}	
+
 			charProfilesCopy[idxChar].Stats[substat] -= 10 * substatValues[substat] * charSubstatRarityMod[idxChar]
 		}
 


internal/substatoptimizer/substatoptimizer.go
```diff
@@ -389,6 +389,12 @@ func RunSubstatOptim(simopt simulator.Options, verbose bool, additionalOptions s
 
 			substatGradients[idxSubstat] = substatEvalResult.DPS.Mean - initialMean
 
+			// fixes cases in which fav holders don't get enough crit rate to reliably proc fav (an important example would be fav kazuha)
+			// might give them "too much" cr (= max out liquid cr subs) but that's probably not a big deal
+			if charWithFavonius[idxChar] && idxSubstat == 1 {
```
jordanlovato, line 394: @k0l11 I think we want `if charWithFavonius[idxChar] && substat == core.CR {` to future proof us a bit better

internal/substatoptimizer/substatoptimizer.go
```diff
@@ -389,6 +389,12 @@ func RunSubstatOptim(simopt simulator.Options, verbose bool, additionalOptions s
...
+				substatGradients[idxSubstat] += 1000
```
jordanlovato, line 395: @k0l11 I think I understand the approach, I believe this would simply make me the optimizer significantly more likely to allocate CR in any given optimization pass and I don't doubt 1000 is a good number in testing. I think it's a reasonable approach.

Regardless, have you considered/tested instead adjusting the CR baseline here? https://github.com/genshinsim/gcsim/blob/16a53d7a21b0685cb9b6c67f6530791ae2c187d3/internal/substatoptimizer/substatoptimizer.go#L248

I wonder if it would be more consistent to increase the baseline CR provided to fav users, and to leave the CR gradient alone. It seems difficult to asses the appropriate number to supply to the gradient to correctly value CR in a fav context, but I'm also haven't dug into the problem too much nor am I very practiced with gradient descents generally. Let me know your thoughts!

jordanlovato: Note, if we do adjust the baseline stats, we also ought to adjust these two lines:

https://github.com/genshinsim/gcsim/blob/16a53d7a21b0685cb9b6c67f6530791ae2c187d3/internal/substatoptimizer/substatoptimizer.go#L365

https://github.com/genshinsim/gcsim/blob/16a53d7a21b0685cb9b6c67f6530791ae2c187d3/internal/substatoptimizer/substatoptimizer.go#L531

jordanlovato (changes requested):
@k0l11 1 adjustment + notes on approach

Let me know your thoughts

internal/substatoptimizer/substatoptimizer.go
k0l11: The problem with this approach is that the CR gradient is usually so low that it gets picked last. Increasing the CR baseline will just make it give more CR at baseline but it will not let it go above that baseline because of how the gradients work out.  

For Kazuha something like this can happen without any changes to optimizer:
- optimizer has a special case for favonius that always adds 4 liquid cr subs at baseline before optimizing crit rolls
- optimizer determines 2 liquid er rolls are the best for kazuha in that sim
- optimizer sees em gives the biggest dmg increase -> give 4 em rolls, since that's the cap for an em/em/em build in kqmc
- optimizer sees atk% gives the next biggest dmg increase -> give 10 atk% rolls
- optimizer has no more rolls left to give to cr because 2 er + 4 em + 10 atk% + 4 cr = 20 liquid subs (cap of kqmc) 

In some cases atk% is missing from the list of gradients so it can give cr subs to Kazuha, but that's not consistent.

internal/substatoptimizer/substatoptimizer.go
k0l11: Sounds good, I've adjusted that part.

internal/substatoptimizer/substatoptimizer.go
jordanlovato: @k0l11 Yeah I thought about it more yesterday and even increasing the CR baseline would run into the same complication with characters that want an EM hat versus a CR hat. Bummer, but makes sense to me that biasing the gradient towards CR would solve better for the kazuha case. 

One last consideration, perhaps this behavior should be limited to optimizer runs on characters that we whitelist? Kazu, venti, sucrose, Jean (maybe every anemo lmao?). I'm ambivalent on this thought, leaning towards a no because I think we always want to bias CR for fav-equipped characters, I don't think that changes based on the most relevant substats for the character. It's another question of by how much we bias, but starting with +1000 avg dmg delta seems like a reasonable enough starting point to me.

Let me know your thoughts, approving this

jordanlovato (approved):
@srliao Approving the CR bias, may need to adjust the 1000 bias to some smaller/larger number in the future. For now, let's run with this.

internal/substatoptimizer/substatoptimizer.go
k0l11: I agree with giving a CR bias to characters with Fav weapons and if issues come up later we can always reconsider the bias.

jordanlovato:
@k0l11 Do you happen to have tests/configurations for this PR (#483) that either you or I could rerun under the build proposed in https://github.com/genshinsim/gcsim/pull/462

It would be nice to
1. Double check that the patch was properly merged
2. We're seeing consistent-ish behavior between the prod build and this PR

Lmk! If not that's okay!

k0l11:
@jordanlovato 
```
# Options
options debug=true iteration=1000 duration=126 workers=60 mode=sl frame_defaults=human swap_delay=12;

# Character setup
# Ayaka
ayaka char lvl=90/90 cons=0 talent=9,9,9; 
ayaka add weapon="amenomakageuchi" refine=5 lvl=90/90;
ayaka add set="blizzardstrayer" count=5; 
ayaka add stats hp=4780 atk=311 atk%=0.466 cryo%=0.466 cd=0.622 ; #main

# Barbara
barbara char lvl=90/90 cons=6 talent=9,9,9; 
barbara add weapon="thrillingtalesofdragonslayers" refine=5 lvl=90/90;
barbara add set="maidenbeloved" count=5;
barbara add stats hp=4780 atk=311 hp%=0.466 hp%=0.466 heal=0.359 ; #main
 
# Ganyu
ganyu char lvl=90/90 cons=0 talent=9,9,9; 
ganyu add weapon="prototypecrescent" refine=5 lvl=90/90;
ganyu add set="blizzardstrayer" count=5;
ganyu add stats hp=4780 atk=311 atk%=0.466 cryo%=0.466 cd=0.622 ; #main

# Kazuha
kazuha char lvl=90/90 cons=0 talent=9,9,9; 
kazuha add weapon="favoniussword" refine=3 lvl=90/90;
kazuha add set="viridescentvenerer" count=5;
kazuha add stats hp=4780 atk=311 em=561 ; #main

# Execution
# Target
target lvl=100 resist=.1;

# Energy 
energy every interval=480,720 amount=1;

# Rotation: Ganyu Q E > Kazuha tEP Q > Barbara (E / N2) > Ayaka D N1 E Q > Kazuha tEP > Ganyu E 2CA > Ayaka D N1 E N2C
active ganyu;

ganyu burst[radius=0], skill;
kazuha skill, high_plunge, burst;
barbara skill;
ayaka dash, attack, skill, burst;
kazuha skill, high_plunge;
ganyu skill, aim[weakspot=1], aim;
ayaka dash, attack, skill, attack:2,charge;

ganyu burst[radius=0], skill;
kazuha skill, high_plunge, burst;
barbara attack:2;
ayaka dash, attack, skill, burst;
kazuha skill, high_plunge;
ganyu skill, aim[weakspot=1], aim;
ayaka dash, attack, skill, attack:2,charge;

restart; 
```
If Kazuha's crit rate is too low here you can easily see it in the standard deviation and the dps will be lower than when he gets enough crit rate to reliably proc Fav.

PR #488, opened 2022-04-18
childe frames update

kurt22i:
- Updates Childe frames
- Sheet: https://docs.google.com/spreadsheets/d/1iBcJqVG3uWhwyHwCbYQby0mdaLZJ_ZhJCo0I4djnsU8/edit?usp=sharing
- Videos: https://youtu.be/FSE7NhQSbHE and https://youtu.be/rpVDDq-F5ZM

This update is not done yet, I need to make sure the code style is ok before I continue

Specifically, line 141 and below is what needs to be reviewed

### DIFF
--- a/internal/characters/tartaglia/frames.go
+++ b/internal/characters/tartaglia/frames.go
@@ -6,62 +6,172 @@ func (c *char) ActionFrames(a core.ActionType, p map[string]int) (int, int) {
 	switch a {
 	case core.ActionAttack:
 		f := 0
+		a := 0
 		if c.Core.Status.Duration("tartagliamelee") > 0 {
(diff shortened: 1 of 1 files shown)


kurt22i:
closing this since the planned new frames system will make this much easier.

soloxcx:
one thing to add -- childe's `skill`, if used while moving, is 1 frame long. however using 1 frame will cause issues with his skill stealing the vaporize from his burst in the common sequence `burst, skill`. so I recommending making his moving skill frame long enough so this doesn't happen, perhaps 5 frames. also there may be situations where the user wants to use a nonmoving skill, so perhaps a param for choosing moving vs nonmoving is ideal.

Issue #486, opened 2022-04-18
animation delay plays twice when buffs expire


kurt22i:
![image](https://user-images.githubusercontent.com/94938654/163743969-8aa4ed19-69be-49c4-8d09-bb3730f1c4ab.png)

https://gcsim.app/viewer/share/perm_fj5xZjz-8DIyVgRn1jEY0

kurt22i:
![image](https://user-images.githubusercontent.com/94938654/163744187-f1116acd-d4fa-481e-bd43-fe40864e04df.png)
I'm pretty sure the issue has to do with this portion of code not removing delay when it should be

kurt22i:
> ![image](https://user-images.githubusercontent.com/94938654/163744187-f1116acd-d4fa-481e-bd43-fe40864e04df.png) I'm pretty sure the issue has to do with this portion of code not removing delay when it should be

Specifically, the issue I think is with https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/pkg/simulation/run.go#L131

This if statement removes prevents delay from being added when the delay has already been done. However, the way it does this is by calling the delay function every frame - so when the delay function suddenly returns something different (say, paramita expired, and tao's frames changed), this statement no longer works correctly.

This can be fixed two ways: 1. "snapshot" the delay or 2. Prevent any and all delay from being added more than 1 frame after the last action, since this should never happen anyway (to my knowledge)

srliao:
> 

`AnimationCancelDelay` function does not get called every frame though.

You can see here: https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/pkg/simulation/run.go#L80-L84

If delay is valid, it sets s.skip and on the next frame if s.skip > 0 it will skip the rest of the block of code

Or did you mean something else?

kurt22i:
> > 
> 
> `AnimationCancelDelay` function does not get called every frame though.
> 
> You can see here:
> 
> https://github.com/genshinsim/gcsim/blob/2fa6d62f942589b81606e5107be8cae5943df947/pkg/simulation/run.go#L80-L84
> 
> If delay is valid, it sets s.skip and on the next frame if s.skip > 0 it will skip the rest of the block of code
> 
> Or did you mean something else?

You're right, it does not get called every frame like I'd thought.

It is still the problem statement though. I looked through debug mode (in gcsim and vscode), and what happens is this:
[...]

Issue #485, opened 2022-04-17
Sim animation/frame system thoughts and possible rework


srliao:
These are just my thoughts on the issue in rough. To be refined...

## Issue

When gcsim was originally conceived/design, I had assumed that we would only run simulation at perfect gameplay as it presents a stable baseline for comparison purposes. For games such as WoW or FFXIV, such assumption I think were reasonable because both combat system are (to an extent) gated by GCD and it is therefore reasonable that majority of the players can press buttons fast enough to be using abilities with no delay between GCD.

However, as it turns out, this assumption is not turning out well for Genshin for a number of reasons:

1. Genshin is closer to an action combat game where mechanical abilities matters. Skilled players can execute the same combo much much faster than any average player
2. Due to the reaction system, playing "frame perfect" can result in significantly different reaction chains compared to someone who simply do not possess the mechanical skill to dos

As a result, there is a greater need for a system that allows for a range of animation cancelling (from "frame perfect" to perhaps no cancelling altogether) by the user.

However, due to my original assumptions, the system was designed such as every ability is simply represented by a number (or two now with the `AnimationCancelDelay` that was later added). This leads us to the hack solution that is:

- `skill[delay=x]`
- `option swap_delay` etc..

The implementation of these relies heavily upon adding delay AFTER the execution of a previous action and is also clunky at best. In addition, it leads to the ugly mess that is this:

![image](https://user-images.githubusercontent.com/906239/163734539-3de536d7-e6b4-4359-8910-afe6b098151f.png)

## Thoughts and Requirements

Ideally, animation should be treated as a first-class citizen so to speak where the system naturally understand that everything potentially has an animation and each animation can be cancelled at certain points (including as soon as it starts)

The following are some of the areas I would personally like to see addressed:

- Animation should be a first class citizen i.e. at any point in time the simulator should know what animation it is currently in (maybe even standing around)
- Action lists and action execution code should have a concept of "when" an action is used. This is in contrast from the current assumption where actions are always executed frame perfect. Basically we want something that matches closer to the game i.e. player can press buttons whenever (whether or not they are registered is a different issue). That being said, there needs to be some sane default/presets as well as backward compatibility in syntax
- There should be better code organization as well as separation of concern. It might make more sense for each action to be self containing (i.e. contains all the information about the animation as well as what the ability does). The current organization in `frames.go` is difficult to read/check and leaves much to be desired
- The simulation run code needs to be able to handle 0 frame actions and not waste a frame doing so
[...]

PR #490, opened 2022-04-18
Refactor actions/animation framework

srliao:
This refactor is a draft to address #485 and #486


internal/player/exec.go
```diff
@@ -15,29 +15,113 @@ var ErrActionNotReady = errors.New("action is not ready yet; cannot be executed"
...
+	p.State.OnStateEnd = nil
+	return nil
+}
+
+func (p *Player) walkFrame(next core.ActionType) int {
+	return 1
+}
+
+func (p *Player) jump(c core.Character, param map[string]int) error {
+	//TODO: how many frames should this consume?
```
kurt22i, line 117: it varies by character, and also by whether or not they're jump cancelling or just jumping

srliao: that's fine. for now it needs to replicate current behaviour. don't want to add new behaviour for now

PR #513, opened 2022-05-05
Add configurable random delay between actions

nlsun:
To better simulate human error.

### DIFF
--- a/pkg/core/core.go
+++ b/pkg/core/core.go
@@ -39,6 +39,39 @@ type Delays struct {
 	Dash   int
 	Jump   int
 	Swap   int
+
+	// Minimum random delay before each skill
+	SkillJitterMin int
+	// Maximum random delay before each skill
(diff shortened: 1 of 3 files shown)


nlsun:
@srliao I hear you may be interested in this or know someone who is

k0l11:
Thanks for the PR. A lot of things have changed since this PR has been submitted, so I'd like to give you a quick explanation of why I'm closing the PR: 
We have a different plan for this now. First, there will be [changes made to gscl](https://github.com/genshinsim/gcsim/issues/838) (the domain specific language that the gcsim configs are written in) to support types which kind of mirror golang's type system. After these changes are merged, we want to give the user the ability to set a function that will be executed at the beginning of every action which can use the previous and next action to do conditionals. In this function you can then `wait` with a certain amount of frames (which can either be fixed or random). With this feature, you can then model jitter in the way that's being proposed here.

PR #540, opened 2022-05-11
added ability for xiao to use E midair.

Charlie-Zheng:
Technically you can plunge after an air E, but that's too complicated to implement, so I'm not implementing it.


internal/characters/xiao/frames.go
```diff
@@ -130,6 +130,7 @@ func BurstJumpFrames(next core.ActionType) int {
 	case core.ActionHighPlunge:
 		return 6 - 5
 	case core.ActionLowPlunge:
+	case core.ActionSkill: //Xiao E can be used in the air
```
srliao, line 133: is there meant to be a `fallthrough` here?

Issue #123, opened 2021-12-31
ec aoe ticks + bounces


srliao:
need to implement ec bounce ticks + aoe syncing

HanSeoulOh:
i am interested in this *cracks knuckles* how do I get started

Issue #546, opened 2022-05-13
Instructors does not activate on freeze


Charlie-Zheng:
https://gcsim.app/viewer/share/qdJL7RLJ3UeG1clQRIBbo

In this sim, at frame 1100, Eula procs freeze with 4pc Instructors equipped, but does not activate the effect.

![image](https://user-images.githubusercontent.com/16330291/168188684-78592300-ab06-468b-b981-4fee3327621c.png)
![image](https://user-images.githubusercontent.com/16330291/168188692-1b254dde-dddb-485d-a35f-ef62266cbe15.png)

RarePossum:
https://gcsim.app/viewer/share/K8_X7NEDTwn_VBukT_rLk Adding another sim to demonstate

PR #547, opened 2022-05-13
Fixes Instructors not proccing on freeze (Issue #546)

Charlie-Zheng:
Moved the onFreeze event emit to occur before the tryFreeze function returns.


Charlie-Zheng:
fixes #123

PR #554, opened 2022-05-14
Fix characters not dying

kurt22i:
Fixes the bug where characters could continue performing actions even without any HP.
Tested here: https://gcsim.app/viewer/share/N7yi5q_q9J3eFnHOpTII0


srliao:
fixed in rewrite

PR #566, opened 2022-05-22
Guisim

IvanHdzF:
Initial GUI implementation gor gcsim
Supports weapon comparisons with different main stats

TODO: Add artifact set comparisons


srliao:
can you rename the folder to something lower case? lol

IvanHdzF:
should be fine now I think :)

Issue #302, opened 2022-02-22
when using apl mode with flag +limit, the more iteration, the less yae dps


moriyasuwakoda:
`yaemiko skill:3 +limit=1;`

## with +limit flag
iteration: 1, dps: 17071
iteration 10, dps: 3393
iteration: 100, dps: 359

## without +limit flag
iteration: 1, dps: 19328
iteration: 10, dps: 19114
iteration: 100: dps: 18936

---

```
####----GENERATED CHARACTER BLOCK DO NOT EDIT----####
yaemiko char lvl=90/90 cons=0 talent=6,8,6;
yaemiko add weapon="kagurasverity" refine=1 lvl=90/90;
yaemiko add stats def=625.0 hp=1.709e+4 atk=1645 er=1.000 em=345.0 cr=0.7360 cd=2.078 electro%=0.6160;

####----END GENERATED CHARACTER BLOCK DO NOT EDIT----####

options duration=14 iteration=1 mode=apl;
target lvl=90 pyro=0.1 dendro=0.1 hydro=0.1 electro=0.1 geo=0.1 anemo=0.1 physical=.1 cryo=.1;

active yaemiko;

yaemiko skill:3 +limit=1;
```

HanSeoulOh:
Thanks for finding this, looks like limit could be being set globally across instances instead of per simulation instance. @srliao

Tsym-or-Tysm:
Related to #233 - not sure if it's the same bug or not?

srliao:
i'll have to take a look but i believe it might be the same issue. i think has to do with the profile cloning code not properly cloning limit (too many pointers everywhere)

thanks for submitting this

srliao:
so i can't seem to reproduce this bug. if I run this config

```
options debug=true iteration=10000 duration=10 workers=60 mode=apl;

xingqiu char lvl=90/90 cons=6 talent=9,9,9; 
xingqiu add weapon="dullblade" refine=1 lvl=90/90;

##Default Enemy
target lvl=100 resist=0.1;

##Actions List
active xingqiu;

xingqiu attack +limit=2;
```

then if limit is working correctly xingqiu should only auto twice. this is the output


```
------------------------------------------
xingqiu contributed the following dps:
        Normal 0 (49.44% of total, 1.00 average damage procs): avg 20.01 [min: 19.52 | max: 29.28]
[...]

PR #570, opened 2022-05-26
fix config not cloned

srliao:
Configs were not being cloned in wasm. This results in +limit=1 not working properly


srliao:
fixes #302

PR #596, opened 2022-06-12
Fix substat optimizer not using fixed_substats_count in some areas

skippi:
Two areas where this is the case:

1. Substat deduction from using main stats of the same type. Optimizer shouldn't always deduct 2.
2. Substat readdition when formatting substat allocation. Optimizer shouldn't always add 2.


srliao:
@skippi can we close this in favour of the implementation in `next`?

skippi:
I'm in favor. Can port these changes into the upcoming rewrite.

PR #625, opened 2022-06-30
Kuki feature update

IvanHdzF:
C4 now generates particles on each proc with a %45 chance of doing so.

Additionally the ICD tag changed to none, since it isn't shared with the skill

Disclaimer: This PR is NOT meant to be a bugfix, instead, it's a feature addition, so please treat is as such


IvanHdzF:
https://gcsim.app/viewer/share/perm_y39jVvbC2kY1LbZCXw1ZM
This is a test link

PR #629, opened 2022-07-10
Includes pre_recovery field in AddEnergy log entry

isundaylee:
Not sure if there's a particular reason why `pre_recovery` field is not included in the `energy` event emitted from `AddEnergy`, but if there isn't, here's a patch to add it. This makes it more consistent to the `energy` events emitted elsewhere.

Tested locally by inspecting the output data.

First time contributing here, so let me know if I missed anything. Thanks!


isundaylee:
Gonna clean up the rebase shortly

PR #630, opened 2022-07-10
Fixes build.sh for Mac

isundaylee:
Mac's version of `date` only has `-u`, not `--utc`. I tested on a Debian 10 machine and verified that `-u` and `--utc` give the same output on Linux.


isundaylee:
No longer relevant for the `next` branch.

PR #632, opened 2022-07-11
Support outputting debug log for min/max-DPS runs

isundaylee:
If the new CLI option `--debugMinMax` is given, will additionally output min/max-DPS runs' debug logs in `debug_min_dps_run` and `debug_max_dps_run` JSON fields in the output.

Tested with doing a sim and tallying up the damage from the 3 debug runs:

```
$ tools/gcsim_tools.py show-damages /tmp/gcsim.out.gz.gz | pc -1 | get_sum
698147

$ tools/gcsim_tools.py show-damages /tmp/gcsim.out.gz.gz --debug-log-flavor debug_min_dps_run | pc -1 | get_sum
423780

$ tools/gcsim_tools.py show-damages /tmp/gcsim.out.gz.gz --debug-log-flavor debug_max_dps_run | pc -1 | get_sum
809334
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,6 @@ internal/characters/replace.txt
 ./cmd/gcsim/tests
 ./cmd/gcsim/*.gz
 ./cmd/gcsim/*.json
+cmd/gcsim/gcsim
 !release-notes.txt
 \.vscode/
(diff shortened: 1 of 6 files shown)


srliao:
now we just need ui support for this. thanks for the pr =)

PR #640, opened 2022-07-17
Heizou frames update

IvanHdzF:
Heizou frame update, Srl, please check the part 2 commit, I think the return at the end on skill function must be "f,a" instead "return skillChargeStart + dur + f, a" for the cancels to work?

### DIFF
--- a/internal/characters/heizou/burst.go
+++ b/internal/characters/heizou/burst.go
@@ -25,24 +25,24 @@ func (c *char) Burst(p map[string]int) (int, int) {
 	//TODO: does heizou burst snapshot?
 	snap := c.Snapshot(&ai)
 
-	c.AddTask(func() {
-		for i, t := range c.Core.Targets {
-			// skip non-enemy targets
-			if t.Type() != core.TargettableEnemy {
-				continue
-			}
-			if c.Base.Cons >= 4 {
-				c.c4(i)
-			}
-			if i > 4 {
-				break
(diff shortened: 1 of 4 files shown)


IvanHdzF:
Uh oh, there seems to be a problem with how stacks are added when holding

IvanHdzF:
stack bug fixed

IvanHdzF:
Working as intended
https://gcsim.app/viewer/share/perm_nYrESkZlFIPphhrS3hnJ8

IvanHdzF:
Since this PR has now went beyond the original scope i will summarize all the changes done to heizou here:
1.- Complete frame update of heizou abilities with cancels included, data provided by @k0l11 
2.-Windmuster Iris now snapshot, as per evidence provided https://discord.com/channels/763583452762734592/996610362482901032/997031486136000562
3.- Aura check callback has been changed to a function to avoid doing the check AFTER the aura element consumption has occured and hence not doing an iris instance of damage when it should have
4.- Hold skill didn't properly add stacks before at the correct intervals, numbers tweaked to properly fix this

https://gcsim.app/viewer/share/perm_o2o1Bp6YJgmJu0bgEYLMQ
Working link

IvanHdzF:
Made the update needed
https://gcsim.app/viewer/share/perm_TCliXT_ZWUSYXlfQJwWyJ
Working link

PR #645, opened 2022-07-27
fix transient blossom icd reset timing

MrPillow1:
correction to the transient blossom icd reset timing, where the timing is now after the construct has been created, as opposed what used to be when the skill was used. source: https://discord.com/channels/763583452762734592/763610791839924224/1001895536552595536

this issue used to give rise to bugs like what happens in here (https://gcsim.app/viewer/share/heOJY9AJNuV7PON6x3psV) at the 25.92s mark.


srliao:
@MrPillow1 Shizuka added fix to next on https://github.com/genshinsim/gcsim/pull/652 so will be closing this

PR #666, opened 2022-08-08
Klee frame update

skippi:
EDIT: Tagging #209.


internal/characters/klee/skill.go
```diff
@@ -69,19 +76,18 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
...
 	for i := 0; i < minehits; i++ {
-		c.Core.QueueAttack(ai, combat.NewCircleHit(c.Core.Combat.Player(), 1, false, combat.TargettableEnemy), 0, skillStart+160, c.c2)
+		c.Core.QueueAttack(ai, combat.NewCircleHit(c.Core.Combat.Player(), 1, false, combat.TargettableEnemy), 0, mineHitmark, c.c2)
 	}
 
-	c.c1(skillStart + 30)
+	c.c1(bounceHitmarks[0])
 
-	c.SetCD(action.ActionSkill, 1200)
+	c.SetCDWithDelay(action.ActionSkill, 1200, 33)
```
skippi, line 85: This is bugged when used with `.ready.klee.skill`. Not sure what to do here to fix it, other than to propose `c.SetCD(action.ActionSkill, 1233)`.

internal/characters/klee/skill.go
```diff
@@ -9,10 +9,18 @@ import (
...
+var bounceHitmarks = []int{71, 111, 140}
+const mineHitmark = 240
 
 func init() {
-	skillFrames = frames.InitAbilSlice(67)
+	skillFrames = frames.InitAbilSlice(75)
+	skillFrames[action.ActionAttack] = 66
+	skillFrames[action.ActionCharge] = 69
+	skillFrames[action.ActionSkill] = 68
+	skillFrames[action.ActionBurst] = 34 // TODO: program early cancel for skill trigger
```
skippi, line 20: I haven't added the E -> Q cancel that triggers a skill use without triggering the CD (useful for CW). Do we know if this is still relevant? I heard this was patched recently.

k0l11: It is still relevant, the only thing that got fixed is gaining multiple stacks by pressing E in quick succession.

soloxcx: that part wasn't patched, you can cancel a phantom E and get a stack of cw/kagura.

the bug that was patched was double tapping E and getting a 2nd stack with 1 E usage

PR #693, opened 2022-08-10
Yanfei frames update

k0l11:
video: https://youtu.be/O43Ws1bgyXY
sheet: https://docs.google.com/spreadsheets/d/1NslqbafUoW4rhl1D2oVeSsn0SXfvAwnxAlIlwvKY_5Q/edit?usp=sharing

also:
- adjust q burst buff name
- fix yanfei a4 hitlag in multitarget

### DIFF
--- a/internal/characters/yanfei/asc.go
+++ b/internal/characters/yanfei/asc.go
@@ -39,6 +39,9 @@ func (c *char) a4() {
 		if atk.Info.AttackTag != combat.AttackTagExtra && !crit {
 			return false
 		}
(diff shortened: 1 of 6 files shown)


internal/characters/yanfei/asc.go
```diff
@@ -39,6 +39,9 @@ func (c *char) a4() {
 		if atk.Info.AttackTag != combat.AttackTagExtra && !crit {
 			return false
 		}
+		// make it so a4 only applies hitlag once per A4 proc and not everytime an enemy gets hit
```
srliao, line 42: just leaving a note here that i have no idea if this is what happens or not

PR #722, opened 2022-08-11
fix ReceiveParticle on Jean C2 / Venti C4

k0l11:
The ReceiveParticle method can't be overriden unless it is moved under `internal/template/character`, so the custom ReceiveParticle methods in Jean's/Venti's code were never called. The solution is to listen to the OnParticleReceived event instead.
also:
- moves all Jean/Venti constellations to cons.go


shizukayuki (changes requested):
Can you move them to `cons.go`?

k0l11:
moved all of their cons to cons.go

shizukayuki (approved):

PR #738, opened 2022-08-13
partially fix kazuha c2 and c6

### LINKED ISSUE #605: kazuha c6 should be added via predamagemod
title

currently:

image

k0l11:
This PR is for improving the implementation of Kazuha's C2 and C6. Even though there are still some uncertainties, I believe this PR will make things more accurate when compared to before. However, this change spams the status log in debug with kazuha-c2 getting refreshed every 0.5s.

Kazuha C6 dmg up part didn't check for normal/charged/plunge and used a custom snapshot function, when it should be dynamic (fixes #605).

for sure:
- C2 ticks every 0.5s to apply the C2 buff to active char and Kazuha for 1s

unsure:
- Q field starts and ends when? assumption: starts on Initial Hit hitlag end and lasts until the last tick (needed for C2)
- C6 starts when? assumption: on E hitmark / Q Initial Hit hitlag end

TODO:
- [ ] update docs after merge

### DIFF
--- a/internal/characters/kazuha/cons.go
+++ b/internal/characters/kazuha/cons.go
@@ -66,10 +66,6 @@ func (c *char) c6() {
 	c.AddAttackMod(character.AttackMod{
 		Base: modifier.NewBaseWithHitlag("kazuha-c6-dmgup", 60*5), // 5s
 		Amount: func(atk *combat.AttackEvent, _ combat.Target) ([]float64, bool) {
-			// skip if not active
-			if c.Core.Player.Active() != c.Index {
-				return nil, false
-			}
 			// skip if not normal/charged/plunge
 			if atk.Info.AttackTag != combat.AttackTagNormal &&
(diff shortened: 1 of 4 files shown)


internal/characters/kazuha/cons.go
```diff
@@ -3,14 +3,56 @@ package kazuha
...
+// The Autumn Whirlwind field created by Kazuha Slash has the following effects:
+// - Increases Kaedehara Kazuha's own Elemental Mastery by 200 for its duration.
+// - Increases the Elemental Mastery of characters within the field by 200.
+// The Elemental Mastery-increasing effects of this Constellation do not stack.
+func (c *char) c2() {
+	// don't tick if Q isn't up anymore
+	if c.Core.Status.Duration(burstStatus) == 0 {
+		return
+	}
+
```
srliao, line 21: should add a src check here just in case someone manages to reach 50% cdr

srliao (approved):

PR #745, opened 2022-08-15
Add guoba self infuse

skippi:
Anemo absorptions should use Guoba as an aura-check target.
https://cdn.discordapp.com/attachments/453700913115168768/1008584949672190024/output_comp.mp4


srliao (approved):

srliao:
A bit hacky but this is more of a problem of the target interface (lack of proper gadget support) than anything else. Should be ok for now. Will need to redo guoba once we get proper gadget support in

also doesn't fix the fact that you can technically have multiple guoba at once in game (right now it only allows for one)

PR #746, opened 2022-08-15
Remove extra optimizer output

skippi:
Some of the debug messages were leftover from testing.

### DIFF
--- a/pkg/optimization/details.go
+++ b/pkg/optimization/details.go
@@ -281,7 +281,7 @@ func (stats *SubstatOptimizerDetails) optimizeERSubstats(tolMean float64, tolSD 
 	)
 
 	for idxChar, char := range stats.charProfilesERBaseline {
-		charDebug = stats.findOptimalERforChar(idxChar, char, tolMean, tolSD)
+		stats.findOptimalERforChar(idxChar, char, tolMean, tolSD)
 		opDebug = append(opDebug, charDebug...)
 	}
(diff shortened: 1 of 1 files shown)


pkg/optimization/details.go
```diff
@@ -370,7 +359,7 @@ func (stats *SubstatOptimizerDetails) findOptimalERforChar(idxChar int, char pro
 		}
 	}
 
-	return debug
+	return []string{}
```
srliao, line 362: Shouldn't we just get rid of the return altogether?

pkg/optimization/details.go
```diff
@@ -288,7 +281,7 @@ func (stats *SubstatOptimizerDetails) optimizeERSubstats(tolMean float64, tolSD
 	)
 
 	for idxChar, char := range stats.charProfilesERBaseline {
-		charDebug = stats.findOptimalERforChar(idxChar, char, tolMean, tolSD)
+		stats.findOptimalERforChar(idxChar, char, tolMean, tolSD)
 		opDebug = append(opDebug, charDebug...)
```
srliao, line 285: nit picking here but the `charDebug` variable should just be deleted altogether since it doesn't seem to be used

skippi: Not a nitpick, we should remove these whenever it's apparent.

srliao (approved):

PR #751, opened 2022-08-15
Add bennett buff range check

skippi:
Players should not get the bennett buff if they are out of range.


internal/characters/bennett/burst.go
```diff
@@ -45,18 +45,24 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 		Durability: 50,
 		Mult:       burst[c.TalentLvlBurst()],
 	}
-	//TODO: review bennett AOE size
-	c.Core.QueueAttack(ai, combat.NewCircleHit(c.Core.Combat.Player(), 5, false, combat.TargettableEnemy), 37, 37)
+	radius := 6.0
```
srliao, line 48: I think this should be changed to a `const` instead

srliao (approved):

PR #752, opened 2022-08-15
Change default/primary target to use closest target

skippi:
This changes the system to be more reflective of how we perceive the in-game targeting.

### DIFF
--- a/pkg/core/combat/combat.go
+++ b/pkg/core/combat/combat.go
@@ -36,6 +36,7 @@ type Opt struct {
 	DamageMode    bool
 	DefHalt       bool
 	EnableHitlag  bool
+	closestTarget int
 }
 
 func New(opt Opt) *Handler {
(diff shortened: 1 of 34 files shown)


pkg/core/combat/combat.go
```diff
@@ -60,20 +62,32 @@ func (h *Handler) Target(i int) Target {
 	return h.targets[i]
 }
 
+func (h *Handler) cachePrimaryTarget() {
```
skippi, line 65: Although I've added it here, I don't think caching will necessary give a performance increase.

srliao: I haven't looked at the PR in detail yet but if you are caching then will also need to call cache function when you set player pos

srliao:
Moving to https://github.com/genshinsim/gcsim/discussions/761 for discussion for now.

PR #758, opened 2022-08-16
add cons condition

imring:
example: https://next.gcsim.app/v3/viewer/share/f87ed6c3-9752-46ad-b6e1-046a8b955c16
```
let c1 = .cons.kokomi >= 1;

fn kokoatk(c, max) {
    kokomi attack:2;
    if c1 { kokomi attack; }

    if c != max {
        kokomi dash;
        kokoatk(c+1, max);
    }
}
```

| field1     | field2                | field3 | field4 | description
| ---        | ---                   | ---    | ---    | ---
| `cons`   | `any char name`       | none   | none   | get char cons

### DIFF
--- a/pkg/conditional/conditional.go
+++ b/pkg/conditional/conditional.go
@@ -39,6 +39,8 @@ func Eval(c *core.Core, fields []string) int64 {
 		return evalWeapon(c, fields)
 	case ".keys":
 		return evalKeys(c, fields)
+	case ".cons":
+		return evalConstellation(c, fields)
 	default:
 		//check if it's a char name; if so check char custom eval func
 		name := strings.TrimPrefix(fields[0], ".")
--- a/pkg/conditional/cons.go
+++ b/pkg/conditional/cons.go
@@ -1,0 +1,31 @@
+package conditional
+
+import (
+	"strings"
+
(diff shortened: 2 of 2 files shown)


srliao:
Hmm just thinking about this some, I think it might make more sense to put these things (along with other things I screwed up such as cooldown, energy, etc...) into the character files so that the condition says `.kokomi.cons` or `.kokomi.energy` instead which makes more sense logically.

Could do this by adding a Condition function to the template character at `internal\template\character\charater.go`

```go
func (c *Character) Condition(k string) int64 {
	switch k {
	case "con":
		return int64(c.Base.Cons)
	default:
		return 0
	}
}
```

and then just make sure for any characters providing an override that by default it calls the template function. For example, `albedo.go` would have the following:

```go
func (c *char) Condition(k string) int64 {
	switch k {
	case "skill":
		fallthrough
	case "elevator":
		if c.skillActive {
			return 1
		}
		return 0
	case "c2stacks":
		return int64(c.c2stacks)
	default:
		return c.Character.Condition(k) 
	}
}
```

srliao:
merging this for now. I think what I'll do is later on add an option to also use `.<char>.whatever` so both works

PR #760, opened 2022-08-16
fix ganyu c4 spelling mistake and adjust constellation names

k0l11:
spelling mistake made ganyu c4 not work


srliao (changes requested):
Minor thing but keys should be a const not a field in a struct

k0l11:
fixed

srliao (approved):

PR #776, opened 2022-08-17
fix fischl oz conditional breaking & oz spawn from q

### LINKED ISSUE #608: fischl oz spawn after `f` causing apl to break
fischl oz currently spawns after `f`, causing conditionals to fail

needs to change the spawn oz function to "spawn" oz right away but do not start the pew pew timer for a user given delay

k0l11:
fixes #608 

- replaces the fischloz status with a bool
- changes the semantics of .fischl.oz: instead of specifiying the amount of frames it's still active for, now it instead just specifies whether oz is "active" or not (similar to .albedo.elevator) 
- fixes Oz spawn on Q: Oz should spawn at the end of animation, but he always spawned at the earliest cancel before.

core:
- fixes 0 delay tasks not being executed right away in normal task queue

live:
https://gcsim.app/v3/viewer/share/a6808c9b-140f-44d1-821f-d7ca3f10898a
after fix:
https://gcsim.app/v3/viewer/share/d9f39e3b-8d0d-4206-82db-29d82b29e8ba

config:
```
options swap_delay=12 debug=true iteration=100 duration=100 workers=30;

fischl char lvl=90/90 cons=0 talent=9,9,9;
fischl add weapon="seasonedhuntersbow" refine=5 lvl=90/90;

# ----
target lvl=100 resist=0.1;

# ----
active fischl;
while 1 {
	fischl skill;
	if .fischl.oz == 0 {
		print("shouldn't happen");
	}
}
```

TODO:
- [ ] update docs after merge

### DIFF
--- a/internal/characters/fischl/skill.go
+++ b/internal/characters/fischl/skill.go
@@ -41,6 +41,18 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 	//hitmark is 5 frames after oz spawns
 	c.Core.QueueAttack(ai, combat.NewCircleHit(c.Core.Combat.Player(), 1, false, combat.TargettableEnemy), skillOzSpawn, skillOzSpawn+5)
 
+	c.Core.Tasks.Add(func() {
+		// activate status at CanQueueAfter because adding the status later than that breaks things
+		// if you do E -> D then everything's fine, but any other cancel breaks the conditional
+		// example:
+		// "fischl skill" followed by a ".status.fischloz == 0" check
+		dur := 600 + skillOzSpawn - skillFrames[action.ActionDash]
+		if c.Base.Cons == 6 {
+			dur += 120
+		}
+		c.Core.Status.Add("fischloz", dur)
+	}, skillFrames[action.ActionDash])
+
 	//set on field oz to be this one
 	c.Core.Tasks.Add(func() {
 		c.queueOz("Skill")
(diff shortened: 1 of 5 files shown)


internal/characters/fischl/skill.go
```diff
@@ -41,6 +41,18 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
...
+	c.Core.Tasks.Add(func() {
+		// activate status at CanQueueAfter because adding the status later than that breaks things
+		// if you do E -> D then everything's fine, but any other cancel breaks the conditional
+		// example:
+		// "fischl skill" followed by a ".status.fischloz == 0" check
+		dur := 600 + skillOzSpawn - skillFrames[action.ActionDash]
+		if c.Base.Cons == 6 {
+			dur += 120
+		}
+		c.Core.Status.Add("fischloz", dur)
```
shizukayuki, line 53: Can set it to `skillOzSpawn+dur` instead of using a task.

shizukayuki (changes requested):
Since you removed the status from `queueOz` you need to add it back here too.
https://github.com/genshinsim/gcsim/blob/97f667514fbc1b5d46562fce5da8727f8fb685f1/internal/characters/fischl/burst.go#L26-L28

k0l11:
I've removed the oz status for a bool that gets used for her custom oz conditional

shizukayuki (approved):

srliao (approved):

PR #811, opened 2022-08-19
refactor conditionals

srliao:
As previously discussed in https://github.com/genshinsim/gcsim/pull/758, refactoring conditionals to start with character name instead for all relevant conditions

Also force sim to error out when invalid conditions are present


srliao:
closing in favour of #814

PR #820, opened 2022-08-20
add kill_target

imring:
example: https://gcsim.app/v3/viewer/share/824dc52a-3531-4fe7-b5cf-7a074b918a5e
- available only in damage mode;
- if you are using the function for dynamic duration, set more hp for the target;


srliao:
https://gcsim.app/v3/viewer/share/b655eb3f-0df3-4a86-9e3f-ca953c10590c confirming that dps works as expected

srliao (approved):

PR #824, opened 2022-08-21
implement kaeya c2, rewrite kaeya q ticks

### LINKED ISSUE #766: kaeya c2 is not implemented
Every time Glacial Waltz defeats an opponent during its duration, its duration is increased by 2.5s, up to a maximum of 15s.
- The description is incorrect. The duration is extended upon any enemy defeat, not just enemies defeated by Glacial Waltz.

k0l11:
fixes #766 

tested here (turn on status and character logs): https://gcsim.app/v3/viewer/share/dd07faf3-ea73-4c56-b13b-fb0539917776

- implements kaeya c2
- adds a status for kaeya's burst ("kaeya-q")
- rewrite kaeya q tick logic to depend on burst status duration instead of the burst duration being hardcoded 8s
- adds a few comments for Kaeya's constellations

TODO:
- [ ] update docs after merge

### DIFF
--- a/internal/characters/kaeya/burst.go
+++ b/internal/characters/kaeya/burst.go
@@ -55,7 +55,6 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 	c.burstTickSrc = c.Core.F
 	for i := 0; i < count; i++ {
 		// each icicle will start at i * offset (i.e. 0, 40, 80 OR 0, 30, 60, 90)
-		// assume each icicle will last for 8 seconds
 		// assume damage dealt every 120 (since only hitting at the front)
 		// on icicle collision, it'll trigger an aoe dmg with radius 2
 		// in effect, every target gets hit every time icicles rotate around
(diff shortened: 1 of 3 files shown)


internal/characters/kaeya/burst.go
```diff
@@ -33,34 +38,39 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
...
+	// Glacial Waltz will generate 1 additional icicle, and will regenerate 15 Energy when cast.
 	if c.Base.Cons == 6 {
 		count++
 	}
 	offset := 120 / count
 
+	c.burstTickSrc = c.Core.F
 	for i := 0; i < count; i++ {
 		// each icicle will start at i * offset (i.e. 0, 40, 80 OR 0, 30, 60, 90)
 		// assume each icicle will last for 8 seconds
```
skippi, line 58:
```suggestion
```
We should remove this since it's outdated.

internal/characters/kaeya/burst.go
```diff
@@ -69,3 +79,23 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
...
+		if c.Core.Status.Duration(burstKey) == 0 {
+			return
+		}
+		// check if it's still the same burst
+		if c.burstTickSrc != src {
+			c.Core.Log.NewEvent("kaeya burst tick ignored, src diff", glog.LogCharacterEvent, c.Index).
+				Write("src", src).
+				Write("new src", c.burstTickSrc)
+			return
+		}
```
skippi, line 95: I think this log won't output unless we do the src check before status check.

skippi (changes requested):

skippi (approved):

PR #825, opened 2022-08-21
bugfixes - hutao c6, kuki e & c6, noelle, heal logs, superconduct

k0l11:
Hu Tao:
- hutao c6 should check every 2s regardless of hurt and c6 icd but should only proc when at < 2 hp for this check only
- removed unnecessary c6Check call on hutao e (hutao e hp drain calls Drain(), which emits OnCharacterHurt to trigger c6Check anyways)

Kuki:
- adjusted kuki e to use Drain() in order to be able to proc HPDown related effects (like her C6)
- fix incorrect kuki e cd (was 90f = 1.5s instead 15s, my fault)

Noelle:
- fix noelle e cast not having a chance to heal
- fix noelle e hitmark being after E -> E/Q cancel
- fix noelle q cast hit 1 and 2 not having a chance to heal
- fix noelle c4 not proccing if e is recast while shield is still up 

core:
- heal logs used c.Base.HP instead of c.HPCurrent, so the logs were showing wrong previous / current hp values
- fixes superconduct not working because of missing Ele

hitlag related:
- hutao c6 icd should not be affected by hitlag
- kuki c6 icd should not be affected by hitlag
- noelle a1 icd should not be affected by hitlag

### DIFF
--- a/internal/characters/kuki/skill.go
+++ b/internal/characters/kuki/skill.go
@@ -24,15 +24,16 @@ func init() {
 func (c *char) Skill(p map[string]int) action.ActionInfo {
 	// only drain HP when above 20% HP
 	if c.HPCurrent/c.MaxHP() > 0.2 {
+		hpdrain := 0.3 * c.HPCurrent
+		// The HP consumption from using this skill can only bring her to 20% HP.
+		if 0.7*c.HPCurrent/c.MaxHP() <= 0.2 {
+			hpdrain = (c.HPCurrent/c.MaxHP() - 0.2) * c.MaxHP()
+		}
 		c.Core.Player.Drain(player.DrainInfo{
 			ActorIndex: c.Index,
 			Abil:       "Sanctifying Ring",
-			Amount:     .30 * c.HPCurrent,
+			Amount:     hpdrain,
 		})
-		// The HP consumption from using this skill can only bring her to 20% HP.
-		if c.HPCurrent/c.MaxHP() < 0.2 {
-			c.HPCurrent = 0.2 * c.MaxHP()
-		}
 	}
 
 	ai := combat.AttackInfo{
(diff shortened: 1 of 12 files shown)


internal/characters/kuki/skill.go
```diff
@@ -22,11 +22,18 @@ func init() {
...
-	//remove some hp
-	if 0.7*(c.HPCurrent/c.MaxHP()) > 0.2 {
-		c.HPCurrent = 0.7 * c.HPCurrent
-	} else if (c.HPCurrent / c.MaxHP()) > 0.2 { //check if below 20%
-		c.HPCurrent = 0.2 * c.MaxHP()
+	// only drain HP when above 20% HP
+	if c.HPCurrent/c.MaxHP() > 0.2 {
+		hpdrain := 0.3 * c.HPCurrent
+		// The HP consumption from using this skill can only bring her to 20% HP.
+		if 0.7*c.HPCurrent/c.MaxHP() <= 0.2 {
```
srliao, line 29: I think this is more readable...

```go
if (c.HPCurrent - hpdrain)/c.MaxHP() <= 0.2 {
   hpdrain = hpCurrent - 0.2*c.MaxHP()
}
```

can make the `0.2` a constant or something `const hpDrainThreshold = 0.2`

internal/characters/noelle/attack.go
```diff
@@ -49,35 +47,9 @@ func (c *char) Attack(p map[string]int) action.ActionInfo {
...
-					Src:     heal,
-					Bonus:   a.AttackEvent.Snapshot.Stats[attributes.Heal],
-				})
-				done = true
-			}
-		}
-
-	}
+	// TODO: don't forget this when implementing her CA
+	c.healDone = false
```
srliao, line 51: shouldn't this be a local variable instead of struct field? just in case there's some weird timing going on re. charge attacks? not too sure re the animation frames

srliao (changes requested):

srliao (approved):

PR #856, opened 2022-08-24
new 3.0 sets

imring:


pkg/shortcut/artifacts.go
```diff
@@ -20,6 +20,7 @@ var SetNameToKey = map[string]keys.Set{
 	"cw":                     keys.CrimsonWitchOfFlames,
 	"crimson":                keys.CrimsonWitchOfFlames,
 	"witch":                  keys.CrimsonWitchOfFlames,
+	"deepwoodmemories":       keys.DeepwoodMemories,
```
k0l11, line 23: pls add "deepwood" as alias

pkg/shortcut/artifacts.go
```diff
@@ -33,6 +34,7 @@ var SetNameToKey = map[string]keys.Set{
 	"gladiatorsfinale":       keys.GladiatorsFinale,
 	"glad":                   keys.GladiatorsFinale,
 	"gladiators":             keys.GladiatorsFinale,
+	"gildeddreams":           keys.GildedDreams,
```
k0l11, line 37: pls add "gilded" and "gd" as alias

pkg/shortcut/artifacts.go
```diff
@@ -35,6 +36,8 @@ var SetNameToKey = map[string]keys.Set{
 	"glad":                   keys.GladiatorsFinale,
 	"gladiators":             keys.GladiatorsFinale,
 	"gildeddreams":           keys.GildedDreams,
+	"glided":                 keys.GildedDreams,
```
k0l11, line 39: should be "gilded"

internal/artifacts/gildeddreams/gildeddreams.go
```diff
@@ -0,0 +1,103 @@
...
+		add := func(args ...interface{}) bool {
+			atk := args[1].(*combat.AttackEvent)
+			if c.Player.Active() != char.Index {
+				return false
+			}
+			if atk.Info.ActorIndex != char.Index {
+				return false
+			}
+
+			s.emCount = 0
```
srliao, line 60: this section here should just be cached outside of the function. no real reason to recompute every time the reaction gets triggered

srliao: keep it as part of the fields for `s` though i think

srliao (approved):

PR #863, opened 2022-08-26
add hunters path weapon

skippi:


internal/weapons/bow/hunterspath/hunterspath.go
```diff
@@ -0,0 +1,90 @@
...
+	// every 12s.
+	w := &Weapon{}
+	r := p.Refine
+
+	dmgBonus := 0.09 + 0.03*float64(r)
+	val := make([]float64, attributes.EndStatType)
+	for i := attributes.PyroP; i <= attributes.DendroP; i++ {
+		val[i] = dmgBonus
+	}
+	val[attributes.PhyP] = 0
```
srliao, line 47: My fault on this one but the better fix would probably be to move Dendro index up to before Phys here: https://github.com/genshinsim/gcsim/blob/daa7e3c700413706a365a8f4e8da48d084da17ce/pkg/core/attributes/stats.go#L30

In theory shouldn't break anything since nothing should be hard coded? Maybe except UI? 🤔

srliao (approved):

PR #869, opened 2022-08-26
Initial Commit

Charlie-Zheng:
Not done yet, have some issues with decay still

### DIFF
--- a/pkg/enemy/damage.go
+++ b/pkg/enemy/damage.go
@@ -88,7 +88,7 @@ func (t *Enemy) calc(atk *combat.AttackEvent, evt glog.Event) (float64, bool) {
 	emBonus := (2.78 * em) / (1400 + em)
 	var reactBonus float64
 	//check melt/vape
-	if atk.Info.Amped {
+	if atk.Info.AmpType == combat.Melt || atk.Info.AmpType == combat.Vaporize {
 		reactBonus = t.Core.Player.ByIndex(atk.Info.ActorIndex).ReactBonus(atk.Info)
 		// t.Core.Log.Debugw("debug", "frame", t.Core.F, core.LogPreDamageMod, "char", t.Index, "char_react", char.CharIndex(), "reactbonus", char.ReactBonus(atk.Info), "damage_pre", damage)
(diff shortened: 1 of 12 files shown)


pkg/enemy/damage.go
```diff
@@ -88,7 +88,7 @@ func (t *Enemy) calc(atk *combat.AttackEvent, evt glog.Event) (float64, bool) {
 	emBonus := (2.78 * em) / (1400 + em)
 	var reactBonus float64
 	//check melt/vape
-	if atk.Info.Amped {
+	if atk.Info.AmpType == combat.Melt || atk.Info.AmpType == combat.Vaporize {
```
srliao, line 91: i don't think the amped flag should be used here. we can use another flag if needed to track

pkg/reactable/catalyze.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+)
+
+func (r *Reactable) tryAggravate(a *combat.AttackEvent) {
+	if a.Info.Durability < ZeroDur {
+		return
+	}
+
+	switch a.Info.Element {
```
srliao, line 14: this check isn't necessary since you checked in the switch statement already.

the other ones have it because they can be triggered from multiple source elements

pkg/reactable/catalyze.go
```diff
@@ -0,0 +1,109 @@
...
+	a.Info.Amped = true
+	a.Info.AmpType = combat.Aggravate
+}
+
+func (r *Reactable) trySpread(a *combat.AttackEvent) {
+	if a.Info.Durability < ZeroDur {
+		return
+	}
+
+	switch a.Info.Element {
```
srliao, line 40: same comment as above

pkg/reactable/catalyze.go
Charlie-Zheng: The other reactions check to ensure that the trigger is correct. I think this should still do that for consistency

PR #870, opened 2022-08-27
Catalyze Reactions

Charlie-Zheng:


pkg/enemy/damage.go
```diff
@@ -88,7 +88,7 @@ func (t *Enemy) calc(atk *combat.AttackEvent, evt glog.Event) (float64, bool) {
 	emBonus := (2.78 * em) / (1400 + em)
 	var reactBonus float64
 	//check melt/vape
-	if atk.Info.Amped {
+	if atk.Info.AmpType == combat.Melt || atk.Info.AmpType == combat.Vaporize {
```
srliao, line 91: Should add another flag here. Don't use amp flag

pkg/reactable/catalyze.go
```diff
@@ -0,0 +1,109 @@
...
+	a.Info.Amped = true
+	a.Info.AmpType = combat.Aggravate
+}
+
+func (r *Reactable) trySpread(a *combat.AttackEvent) {
+	if a.Info.Durability < ZeroDur {
+		return
+	}
+
+	switch a.Info.Element {
```
srliao, line 40: The other ones are not done to figure out the trigger is correct, but to figure out which element is incoming to apply the correct calcs. The default case is technically unreachable.

So in this case the switch on `a.Info.Element` is redundant and should be removed. However, you still need to check for `r.Durability[attributes.Quicken]`

pkg/enemy/damage.go
Charlie-Zheng: What should I called the flag? Catalyzed?

Charlie-Zheng:
I've updated the code to reflect the requested changes

PR #872, opened 2022-08-28
add collei implementation

skippi:


internal/characters/collei/collei.go
```diff
@@ -3,35 +3,52 @@ package collei
 import (
 	tmpl "github.com/genshinsim/gcsim/internal/template/character"
 	"github.com/genshinsim/gcsim/pkg/core"
+	"github.com/genshinsim/gcsim/pkg/core/event"
 	"github.com/genshinsim/gcsim/pkg/core/keys"
 	"github.com/genshinsim/gcsim/pkg/core/player/character"
 	"github.com/genshinsim/gcsim/pkg/core/player/character/profile"
 )
 
+var dendroEvents = []event.Event{event.OnOverload} // TODO: put all dendro events here
```
skippi, line 12: Charlie's PR is in review. We'll need to update this once merged.

skippi:
Also missing ICDs for skill, burst, and A1. Need a verification on what those are before merging.

srliao:
> Also missing ICDs for skill, burst, and A1. Need a verification on what those are before merging.

Skill as follows:
- attack tag: `ElementalArt`
- attack icd tag: `ElementalArt`
- attack icd group: `ColleiBoomerangForward` and `ColleiBoomerangBack`

ICD group (both forward and back) reset timer is 5s, sequence for both damage and durability is `[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`

Collei A1 it has default ICD group but it's own attack icd tag (not ElementalArt), call it `ColleiA1` or something 

Burst as follows:
- attack tag: `ElementalBurst`
- attack icd tag: `ElementalBurst`
- attack icd group: `ColleiElementalBurst`

ICD group reset timer is 3, durability sequence is `[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` and damage sequence is `[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]`

This applies to all burst ticks

Collei C6 is has no ICD (icd tag = none)

internal/characters/collei/collei.go
```diff
@@ -3,35 +3,52 @@ package collei
...
 	c.Character = tmpl.NewWithWrapper(s, w)
 
-	c.EnergyMax = 80
+	c.EnergyMax = 60
 	c.NormalHitNum = normalHitNum
-	c.BurstCon = 3
-	c.SkillCon = 5
+	c.BurstCon = 5
+	c.SkillCon = 3
+	c.burstExtendCount = 0
```
k0l11, line 34: in Go, you technically don't have to initialize these 4 ints and bools, they're 0 and false by default

internal/characters/collei/asc.go
```diff
@@ -0,0 +1,79 @@
...
+		}, "collei-a1")
+	}
+}
+
+func (c *char) a4() {
+	for _, event := range dendroEvents {
+		c.Core.Events.Subscribe(event, func(args ...interface{}) bool {
+			atk := args[1].(*combat.AttackEvent)
+			char := c.Core.Player.ByIndex(atk.Info.ActorIndex)
+			if !char.StatusIsActive(a4Key) {
```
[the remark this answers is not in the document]
skippi: The `a4Key` status only applies during burst, so I don't think the check is necessary.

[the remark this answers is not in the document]
k0l11: I was thinking about the case where your active char gets the a4 status on the very last tick for 1s and then triggers a reaction after burst status expires but before their a4 status expires, but considering how ExtendStatus is implemented I guess this check is unnecessary.

internal/characters/collei/skill.go
```diff
@@ -0,0 +1,82 @@
...
+		StrikeType: combat.StrikeTypeDefault,
+		Element:    attributes.Dendro,
+		Durability: 25,
+		Mult:       skill[c.TalentLvlSkill()],
+	}
+	for _, hitmark := range skillHitmarks {
+		c.Core.QueueAttack(
+			ai,
+			combat.NewCircleHit(c.Core.Combat.Player(), 2, false, combat.TargettableEnemy),
+			skillHitmarks[0],
```
k0l11, line 47: this should snap on skillRelease

internal/characters/collei/asc.go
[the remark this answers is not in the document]
skippi: Actually you are right, there will be extra log messages for the case you listed. I'll upload a fix.

k0l11:
For her E ICD, I would suggest using ICDTagNone instead of ICDTagSkill even though it's not accurate to how they coded it. It's an easy fix for her Sac Bow interaction (2 boomerangs can coexist and each hit should be able to apply dendro/proc a reaction)

internal/characters/collei/cons.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+		})
+	}
+}
+
+func (c *char) c6() {
+	ai := combat.AttackInfo{
+		ActorIndex: c.Index,
+		Abil:       "Forest of Falling Arrows (C6)",
+		AttackTag:  combat.AttackTagElementalBurst,
```
k0l11, line 67: Her C6 has a custom attacktag. I'm not sure what we usually use in this case, but maybe just using AttackTagNone would be fine?

scripts/keymap/character_data.json
```diff
@@ -94,6 +94,11 @@
   "element": "cryo",
   "weapon_class": "bow"
  },
+ "dori": {
```
k0l11, line 97: this doesn't look right

scripts/keymap/character_data.json
```diff
@@ -264,11 +269,21 @@
...
+  "rarity": 5,
+  "element": "",
+  "weapon_class": "sword"
+ },
  "thoma": {
   "rarity": 4,
   "element": "pyro",
   "weapon_class": "polearm"
  },
+ "tighnari": {
```
k0l11, line 282: this doesn't look right

app/package.json
```diff
@@ -1,63 +0,0 @@
-{
```
k0l11, line 1: I've got no clue about frontend stuff, but is this supposed to get deleted?

internal/characters/collei/skill.go
```diff
@@ -0,0 +1,94 @@
...
+	skillFrames[action.ActionDash] = 54
+	skillFrames[action.ActionJump] = 53
+	skillFrames[action.ActionSwap] = 66
+}
+
+func (c *char) Skill(p map[string]int) action.ActionInfo {
+	// The game has ICD as AttackTagElementalArt, ICDTagElementalArt,
+	// ICDGroupColleiBoomerangForward, and ICDGroupColleiBoomerangBack. However,
+	// we believe this is unnecessary, so just use ICDTagNone.
+	ai := combat.AttackInfo{
```
k0l11, line 35: every E hit has 
```
		CanBeDefenseHalted: true,
		IsDeployable:       true,
```

internal/characters/collei/skill.go
```diff
@@ -80,7 +80,9 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 			duration += 180
 		}
 		c.AddStatus(sproutKey, duration, true)
-		c.QueueCharTask(func() { c.a1Ticks(src) }, sproutHitmark)
+		c.QueueCharTask(func() {
+			c.a1Ticks(src, nil)
```
k0l11, line 84: create the snapshot here and pass it to a1Ticks. make a1Ticks take combat.Snapshot directly instead of a pointer and remove the ``if snap == nil```. see kaeya q for example

internal/characters/collei/skill.go
```diff
@@ -0,0 +1,100 @@
...
+	// The game has ICD as AttackTagElementalArt, ICDTagElementalArt,
+	// ICDGroupColleiBoomerangForward, and ICDGroupColleiBoomerangBack. However,
+	// we believe this is unnecessary, so just use ICDTagNone.
+	ai := combat.AttackInfo{
+		ActorIndex:         c.Index,
+		Abil:               "Floral Brush",
+		AttackTag:          combat.AttackTagElementalArt,
+		ICDTag:             combat.ICDTagNone,
+		ICDGroup:           combat.ICDGroupDefault,
+		StrikeType:         combat.StrikeTypeDefault,
```
k0l11, line 41: should be combat.StrikeTypeSlash

k0l11 (approved):

PR #881, opened 2022-08-29
fix sucrose burst doing extra absorb damage

skippi:
Sucrose elemental ticks only apply after the first hit, making a total of 3 elemental hits rather than 4. Issue was made in discord: https://discord.com/channels/845087716541595668/869210750596554772/1013250508636573746

https://gcsim.app/v3/viewer/share/f9f1e2dd-e7fb-4ab6-a777-6946aa143862

### DIFF
--- a/internal/characters/sucrose/burst.go
+++ b/internal/characters/sucrose/burst.go
@@ -71,8 +71,9 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 	for i := 137; i <= duration+5; i += 113 {
 		c.Core.QueueAttackWithSnap(ai, snap, combat.NewCircleHit(c.Core.Combat.PrimaryTarget(), 5, false, combat.TargettableEnemy), i, cb)
 
+		firstHit := i == 137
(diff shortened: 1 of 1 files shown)


internal/characters/sucrose/burst.go
```diff
@@ -71,8 +71,9 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 	for i := 137; i <= duration+5; i += 113 {
 		c.Core.QueueAttackWithSnap(ai, snap, combat.NewCircleHit(c.Core.Combat.PrimaryTarget(), 5, false, combat.TargettableEnemy), i, cb)
 
+		firstHit := i == 137
```
srliao, line 74: Isn't it better here to just not add the task at all if not first hit?

srliao (approved):

k0l11:
ingame you can get the first tick to do the absorb dmg, so it's working as intended rn.
https://youtu.be/e3mp4Qm82dY

PR #888, opened 2022-08-31
fix itto c6

### LINKED ISSUE #868: Itto C6 is bugged
via @MrPillow1 from discord: https://discord.com/channels/845087716541595668/884485336657444894/1010011264120012951
https://gcsim.app/v3/viewer/share/7046a8de-9e19-4854-a540-384cb1ddb2c9
Itto is doing CAF (1 Stack) -> CAF (1 Stack) here, when it should actually be CA1 (1 Stack) -> CAF (1 Stack) assuming the CAF doesn't proc C6. Will work on this after #729 is merged.

k0l11:
fixes #868 
tested here: https://gcsim.app/v3/viewer/share/6b93aeac-2b70-42ab-aec7-5189ce22bf08

### DIFF
--- a/internal/characters/itto/charge.go
+++ b/internal/characters/itto/charge.go
@@ -74,12 +74,14 @@ func (s SlashType) String() string {
 	return slashName[s]
 }
 
-func (s SlashType) Next(stacks int) SlashType {
+func (s SlashType) Next(stacks int, hasC6 bool) SlashType {
 	switch s {
 	// idle -> charge (based on stacks)
 	case InvalidSlash:
-		if stacks == 1 {
+		if stacks == 1 && !hasC6 {
 			return FinalSlash
+		} else if stacks == 1 && hasC6 {
+			return LeftSlash
 		} else if stacks > 1 {
 			return LeftSlash
 		}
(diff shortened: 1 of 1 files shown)


internal/characters/itto/charge.go
```diff
@@ -173,7 +175,8 @@ func (c *char) ChargeAttack(p map[string]int) action.ActionInfo {
 
 	prevSlash := c.slashState
 	stacks := c.Tags[strStackKey]
-	c.slashState = prevSlash.Next(stacks)
+	hasC6 := c.Base.Cons >= 6 && c.Core.Rand.Float64() < 0.5
+	c.slashState = prevSlash.Next(stacks, hasC6)
```
skippi, line 179: I think `hasC6` as a name is misleading, as this variable is only true half the time. Do you think something like `didC6Proc` would be more readable?

skippi (changes requested):

internal/characters/itto/charge.go
k0l11: How about `c6Proc`?

skippi: I'm okay with that.

skippi (approved):

PR #893, opened 2022-09-02
fix yaemiko charge ICD and durability

skippi:
Technically does not match dm, but this is essentially ICDTagNone in game.

### DIFF
--- a/internal/characters/yaemiko/charge.go
+++ b/internal/characters/yaemiko/charge.go
@@ -20,15 +20,17 @@ func init() {
 }
 
 func (c *char) ChargeAttack(p map[string]int) action.ActionInfo {
+	// Supposed to be ICDTagYaeCharged and ICDGroupYaeCharged. However, it's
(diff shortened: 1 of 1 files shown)


internal/characters/yaemiko/charge.go
```diff
@@ -20,15 +20,17 @@ func init() {
...
 
 func (c *char) ChargeAttack(p map[string]int) action.ActionInfo {
+	// Supposed to be ICDTagYaeCharged and ICDGroupYaeCharged. However, it's
+	// essentially no ICD because it takes ~36f for the charge to hit again.
 	ai := combat.AttackInfo{
 		ActorIndex: c.Index,
 		Abil:       "Charge Attack",
 		AttackTag:  combat.AttackTagExtra,
-		ICDTag:     combat.ICDTagElementalArt,
+		ICDTag:     combat.ICDTagNone,
```
k0l11, line 29: you can use ICDTagExtraAttack and then it would match ingame

k0l11 (approved):

PR #906, opened 2022-09-05
replace cam cancel burst frames

k0l11:
based on https://docs.google.com/spreadsheets/d/1zCwdd6_KYFqMD4OQ_llGLdDshoZTu_1pmAMysxGDQvs/edit?usp=sharing
video: https://youtu.be/KlgZlck5kQQ

TODO:
- [ ] update docs at some point after merge (need to add the video and sheet to the page of every affected char)

### DIFF
--- a/internal/characters/eula/burst.go
+++ b/internal/characters/eula/burst.go
@@ -15,11 +15,12 @@ const burstHitmark = 100
 const lightfallHitmark = 35
 
 func init() {
-	burstFrames = frames.InitAbilSlice(122)
-	burstFrames[action.ActionDash] = 121
-	burstFrames[action.ActionJump] = 121
-	burstFrames[action.ActionSwap] = 121
-	burstFrames[action.ActionWalk] = 117
+	burstFrames = frames.InitAbilSlice(123) // Q -> E
+	burstFrames[action.ActionAttack] = 120  // Q -> N1
+	burstFrames[action.ActionDash] = 122    // Q -> D
+	burstFrames[action.ActionJump] = 121    // Q -> J
+	burstFrames[action.ActionWalk] = 117    // Q -> Walk
+	burstFrames[action.ActionSwap] = 120    // Q -> Swap
 }
 
 const (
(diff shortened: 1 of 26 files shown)


internal/characters/eula/burst.go
```diff
@@ -15,11 +15,12 @@ const burstHitmark = 100
 const lightfallHitmark = 35
 
 func init() {
-	burstFrames = frames.InitAbilSlice(122)
-	burstFrames[action.ActionDash] = 121
-	burstFrames[action.ActionJump] = 121
-	burstFrames[action.ActionSwap] = 121
-	burstFrames[action.ActionWalk] = 117
+	burstFrames = frames.InitAbilSlice(123) // Q -> E
```
skippi, line 18: The tap and hold versions seem to have a difference in frame data (122 vs 123). Is this something that's okay to ignore?

internal/characters/tartaglia/burst.go
```diff
@@ -19,19 +19,18 @@ const (
...
-	burstMeleeFrames = frames.InitAbilSlice(102)
-	burstMeleeFrames[action.ActionAttack] = 101
-	burstMeleeFrames[action.ActionSkill] = 102
-	burstMeleeFrames[action.ActionDash] = 102
-	burstMeleeFrames[action.ActionJump] = 102
-	burstMeleeFrames[action.ActionSwap] = 100
+	burstMeleeFrames = frames.InitAbilSlice(103) // Q -> D
+	burstMeleeFrames[action.ActionAttack] = 102  // Q -> N1
+	burstMeleeFrames[action.ActionSkill] = 102   // Q -> E
+	burstMeleeFrames[action.ActionDash] = 102    // Q -> D
```
skippi, line 25: I believe this should be the jump action.
```suggestion
	burstMeleeFrames[action.ActionJump] = 102    // Q -> J
```

skippi (changes requested):

internal/characters/eula/burst.go
k0l11: I think that we are ok with that, it's essentially the same cancel but recorded for completion.

skippi (approved):

PR #907, opened 2022-09-05
fix itto c6 and replace old hitbox methods

k0l11:
previous C6 fix didn't recalculate the C6 proc when checking for CA1/CA2 -> CAF, which is wrong.
also replaces old hitbox methods with the newer calls.
tested here: https://gcsim.app/v3/viewer/share/7e139306-0bd8-4055-a461-e1d55f4e0047

### DIFF
--- a/internal/characters/itto/charge.go
+++ b/internal/characters/itto/charge.go
@@ -175,8 +175,12 @@ func (c *char) ChargeAttack(p map[string]int) action.ActionInfo {
 
 	prevSlash := c.slashState
 	stacks := c.Tags[strStackKey]
-	c6Proc := c.Base.Cons >= 6 && c.Core.Rand.Float64() < 0.5
-	c.slashState = prevSlash.Next(stacks, c6Proc)
+	if !c.c6CarryOver {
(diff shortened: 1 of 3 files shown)


internal/characters/itto/charge.go
```diff
@@ -229,7 +238,9 @@ func (c *char) ChargeAttack(p map[string]int) action.ActionInfo {
 		Frames: func(next action.Action) int {
 			f := chargeFrames[curSlash][next]
 			// handle CA1/CA2 -> CAF frames
-			if next == action.ActionCharge && curSlash.Next(c.Tags[strStackKey], c6Proc) == FinalSlash {
+			c.c6Proc = c.Base.Cons >= 6 && c.Core.Rand.Float64() < 0.5
```
shizukayuki, line 241: this should be outside, since the frames func is called more than once.

shizukayuki (changes requested):
i don't think the `c6CarryOver` field is needed. you can just set `c6Proc` at char init

shizukayuki (approved):

skippi (approved):

PR #924, opened 2022-09-11
add target collision checks

srliao:


pkg/gcs/sysfunc.go
```diff
@@ -256,11 +256,11 @@ func (e *Eval) setTargetPos(c *ast.CallExpr, env *Env) (Obj, error) {
 	}
 
 	//check if index is in range
-	if idx < 1 || idx >= e.Core.Combat.TargetsCount() {
-		return nil, fmt.Errorf("index for set_default_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.TargetsCount()-1, idx)
+	if idx < 1 || idx >= e.Core.Combat.EnemyCount() {
+		return nil, fmt.Errorf("index for set_default_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.EnemyCount()-1, idx)
```
skippi, line 260: I don't think this should have a `-1`.

pkg/gcs/sysfunc.go
```diff
@@ -290,11 +290,11 @@ func (e *Eval) killTarget(c *ast.CallExpr, env *Env) (Obj, error) {
 	}
 
 	//check if index is in range
-	if idx < 1 || idx >= e.Core.Combat.TargetsCount() {
-		return nil, fmt.Errorf("index for kill_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.TargetsCount()-1, idx)
+	if idx < 1 || idx >= e.Core.Combat.EnemyCount() {
+		return nil, fmt.Errorf("index for kill_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.EnemyCount()-1, idx)
```
skippi, line 294: Same as earlier response.

pkg/gcs/sysfunc.go
```diff
@@ -196,8 +196,8 @@ func (e *Eval) setDefaultTarget(c *ast.CallExpr, env *Env) (Obj, error) {
 	}
 
 	//check if index is in range
-	if idx < 1 || idx >= e.Core.Combat.TargetsCount() {
-		return nil, fmt.Errorf("index for set_default_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.TargetsCount()-1, idx)
+	if idx < 1 || idx > len(e.Core.Combat.Enemies()) {
+		return nil, fmt.Errorf("index for set_default_target is invalid, should be between %v and %v, got %v", 1, len(e.Core.Combat.Enemies()), idx)
```
skippi, line 200: Is this different from `EnemyCount()`?

internal/characters/xiangling/skill.go
```diff
@@ -38,34 +37,36 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
...
+	// 			10,
+	// 			c.c1,
+	// 			part,
+	// 		)
+
+	// 		c.Core.Tasks.Add(func() {
+	// 			c.guoba.Durability[attributes.Pyro] = 0
+	// 		}, infuseWindow+1) // +1 since infuse window is inclusive
+	// 	}, delay+i*100-10) // 10 frame window to swirl
+	// }
```
skippi, line 69: We should remove this since it's unused.

internal/characters/xiangling/guoba.go
```diff
@@ -6,25 +6,80 @@ import (
...
+			SetEnded(p.c.Core.F + infuseWindow + 1)
+		p.Durability[attributes.Pyro] = infuseDurability
+		p.Core.Tasks.Add(func() {
+			p.Durability[attributes.Pyro] = 0
+		}, infuseWindow+1) // +1 since infuse window is inclusive
+		//queue this in advance because that's how it is on live
+		p.breath()
+		// case 113, 213, 313, 413: //pew pew window
+		// log.Printf("pew at %v\n", p.timer)
+		// p.breath()
```
skippi, line 59: Similar case here with comments.

skippi (changes requested):

pkg/gcs/sysfunc.go
srliao: Code here is actually wrong, should be

```go
	if idx < 1 || idx > e.Core.Combat.EnemyCount() {
		return nil, fmt.Errorf("index for set_default_target is invalid, should be between %v and %v, got %v", 1, e.Core.Combat.EnemyCount(), idx)
	}

	e.Core.Combat.SetEnemyPos(idx-1, x, y)
```

Used to be target[0] was Player so SetTarget (enemy really) should be 1+; However, now that it's independent enemy array, first enemy is actually 0 now, but for syntax consistency we should still allow 1 to len(...)

pkg/gcs/sysfunc.go
srliao: Nope. Fixed

pkg/gcs/sysfunc.go
srliao: See comment before

skippi (approved):

PR #925, opened 2022-09-11
fix quicken doing extra attach/refill

skippi:
~~The attach/refill logic is already handled in parent calls to tryQuicken().~~

EDIT:
Need to check if quicken is already on target before doing a refill/attach.
EC test (https://dendro.gcsim.app/v3/viewer/share/891468b7-88a1-4a59-918d-004b74d7f874)
Quicken test (https://dendro.gcsim.app/v3/viewer/share/acc417f2-0f0c-4f1c-a6c0-8089c892fa6b)

### DIFF
--- a/pkg/reactable/catalyze.go
+++ b/pkg/reactable/catalyze.go
@@ -15,10 +15,10 @@ func (r *Reactable) tryAggravate(a *combat.AttackEvent) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


srliao:
i'm fairly certain #935 fixes this but will probably need to write some test cases to confirm

PR #926, opened 2022-09-11
show duration stats in detail page

skippi:
Preview: https://gcsim.app/v3/viewer/share/4b307aea-7226-458c-9d0d-19c538c78622

### DIFF
--- a/pkg/result/results.go
+++ b/pkg/result/results.go
@@ -102,6 +102,7 @@ func CollectResult(data []simulation.Result, mode bool, chars []string, detailed
 
 	// var dd float64
 
+	result.Duration.Min = math.MaxFloat64
 	// Loop through each iteration to build overall statistics
 	for iteration, v := range data {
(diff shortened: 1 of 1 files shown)


skippi:
We can also update the web viewer if we want to keep the work in one PR.

pkg/result/results.go
```diff
@@ -513,6 +515,7 @@ func (r *Summary) PrettyPrint() string {
 	}
 
 	sb.WriteString("------------------------------------------\n")
+	sb.WriteString(fmt.Sprintf("Average duration of %.2f seconds (min: %.2f max: %.2f std: %.2f)\n", r.Duration.Mean, r.Duration.Min, r.Duration.Max, r.Duration.SD))
```
skippi, line 518: Should we only print this in HP/TTK mode?

srliao: I guess it should be constant with SD of 0 in non HP/TTK mode? It's fien for now imo...

srliao (approved):

PR #931, opened 2022-09-12
fix the sim duration

imring:
sets the sim duration to 1 second, if the duration is less than 1 second
before: https://gcsim.app/v3/viewer/share/b401ea03-6014-4b5a-b9e4-8ed6e9202735
after: https://gcsim.app/v3/viewer/share/4ff97203-15ad-44d6-8b6c-400b93ed8fd7

*need more tests*

### DIFF
--- a/pkg/simulation/run.go
+++ b/pkg/simulation/run.go
@@ -66,9 +66,14 @@ func (s *Simulation) Run() (Result, error) {
 
 	s.stats.Seed = s.C.Seed
 
+	duration := s.C.F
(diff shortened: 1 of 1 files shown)


srliao:
@imring can we close this as wontfix? I'm of the opinion that even if it's technically "cheating" a bit, the calculation is accurate

imring:
> @imring can we close this as wontfix? I'm of the opinion that even if it's technically "cheating" a bit, the calculation is accurate

yes, close it

PR #933, opened 2022-09-13
Updated and Fixed Enka Importto 3.0 1st half

ChinPTeo:
-Added Collei and Tighnari
-Fixed Ayaka and Mona
-Fixed level 1 rank 1 weapons bug


srliao (approved):
"looks" right :pepela:

Issue #390, opened 2022-03-11
implement "srp"


srliao:
Relevant discussions here: https://discord.com/channels/763583452762734592/851428030094114847/951672464969183272

Changes required:

- [ ] Sucrose burst absorption should do anemo damage first, then absorbed element
- [ ] Kazuha burst absorption should do absorbed element damage first, then anemo
- [ ] Aura attachment needs to happen AFTER damage calculation (resulting in not able to swirl absorbed element if target has no initial element)

The last point is actually consistent in general and not related to "srp"

Also... SRP not real xD only the last point matters and that appears to be a general quality of the reaction system and not specific to anemo/swirl/kazuha/sucrose etc...

PR #935, opened 2022-09-14
Refactor reactable to account for modifiers and add dendro reactions

srliao:
Refactoring reactable and implement dendro reactions based on notes here: https://github.com/genshinsim/gcsim/discussions/855


srliao:
This PR should also fix #390 once it's completed, as attach/refill code will be moved to the end of damage

PR #943, opened 2022-09-19
fix anemo absorption location and naming

k0l11:
This PR fixes cases in which the absorption location was not reset on every ability use (Kazuha E, Traveler (Anemo) E and Q) and adjusts the variable naming/logging to use "absorb" instead of "infuse" (+ a few formatting adjustments).

### DIFF
--- a/internal/characters/heizou/heizou.go
+++ b/internal/characters/heizou/heizou.go
@@ -5,7 +5,6 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core"
 	"github.com/genshinsim/gcsim/pkg/core/action"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
-	"github.com/genshinsim/gcsim/pkg/core/combat"
 	"github.com/genshinsim/gcsim/pkg/core/keys"
 	"github.com/genshinsim/gcsim/pkg/core/player/character"
 	"github.com/genshinsim/gcsim/pkg/core/player/character/profile"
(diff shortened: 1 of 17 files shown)


internal/characters/kazuha/burst.go
```diff
@@ -26,8 +26,9 @@ func init() {
 
 func (c *char) Burst(p map[string]int) action.ActionInfo {
 
-	c.qInfuse = attributes.NoElement
-	c.infuseCheckLocation = combat.NewCircleHit(c.Core.Combat.Player(), 1.5, false, combat.TargettableEnemy, combat.TargettablePlayer, combat.TargettableGadget)
+	c.qAbsorb = attributes.NoElement
+	c.absorbCheckLocation = combat.NewCircleHit(c.Core.Combat.Player(), 1.5, false, combat.TargettableEnemy, combat.TargettablePlayer, combat.TargettableGadget)
```
srliao, line 30: since both burst and skill uses this, wouldnt this cause problem if you do skill -> teleport -> burst?

internal/characters/heizou/heizou.go
```diff
@@ -33,8 +31,6 @@ func NewChar(s *core.Core, w *character.CharWrapper, _ profile.CharacterProfile)
 	c.SkillCon = 3
 	c.BurstCon = 5
 
-	c.infuseCheckLocation = combat.NewCircleHit(c.Core.Combat.Player(), 0.1, false, combat.TargettableEnemy)
```
srliao, line 36: no change to any of heizou's ability required?

k0l11: The only thing that can absorb is Heizou's Q, but that is implemented as an attack that's centered on primary target, so I don't think I have to change anything there?

skippi (approved):

PR #944, opened 2022-09-19
diluc e window should be hitlag extendable

k0l11:
via Charliex3000#9403 from discord: https://discord.com/channels/845087716541595668/869210750596554772/1021493420914638859

### DIFF
--- a/internal/characters/diluc/diluc.go
+++ b/internal/characters/diluc/diluc.go
@@ -17,13 +17,14 @@ func init() {
 
 type char struct {
 	*tmpl.Character
-	eCounter   int
-	eWindowKey string
-	a4buff     []float64
-	c2buff     []float64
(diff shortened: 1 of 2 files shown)


internal/characters/diluc/diluc.go
```diff
@@ -17,12 +17,12 @@ func init() {
...
 type char struct {
 	*tmpl.Character
-	eCounter int
-	eWindow  int
-	a4buff   []float64
-	c2buff   []float64
-	c2stack  int
-	c4buff   []float64
+	eCounter   int
+	eWindowKey string
```
srliao, line 21: this should be changed to a const instead

skippi (approved):

skippi (approved):

Issue #835, opened 2022-08-22
rewrite ShieldBonusMod


k0l11:
the current implementation of AddShieldBonusMod doesn't allow for hitlag extension which is needed for the following mods:
- [ ] 4 tom shield bonus mod (add missing hitlag extension)

k0l11:
closing in favor of #859

Issue #834, opened 2022-08-22
rewrite stampercentmod


k0l11:
the current implementation of AddStamPercentMod doesn't allow for hitlag extension which is needed for the following mods:
- wine and song stam percent mod
- bloodstained stam percent mod

PR #945, opened 2022-09-19
add logging of infusions as a status

k0l11:
This PR addresses the fact that weapon infusions aren't being logged anywhere. To fix this issue, I adjusted the AddWeaponInfuse method to add a hitlag extendable (only if dur != -1) status for every infusion. This eliminates the need to add an extra status when you need to keep track of the weapon infusion like in Aloy's case and makes it easier to tell in debug whether an infusion is up or not.

Some other char specific changes:
- adjusted aloy's e key names 
- removed logging for chongyun's infusion, because it can be seen as a status now
- Keqing can do 5N1C + N1 hitlagless with her electro infusion ingame, so I adjusted the infusion duration to barely cover that combo. She will only be able to do 5N1C with hitlag like she does ingame (this is because the status extension is fractional and hitlag rounds up for animation, so you lose a little bit of time when hitlagged in the sim).

TODO:
- [ ] update docs at some point after merge (every infusion doubles as a status now)

### DIFF
--- a/internal/characters/aloy/skill.go
+++ b/internal/characters/aloy/skill.go
@@ -23,7 +23,7 @@ func init() {
 }
 
 const (
-	rushingIceKey = "rushingice"
+	rushingIceKey = "aloy-rushing-ice"
 )
 
 // Skill - Handles main damage, bomblet, and coil effects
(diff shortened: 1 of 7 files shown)


srliao:
@k0l11 is there a reason why we are using AddStatus instead of just generating a new log event in WeaponInfusion?

k0l11:
I'm using status, because you're supposed to be able to see that the infusion gets hitlag extended (it should be part of the "mods affected" list).

srliao:
Ok so this really relates to #834 and #835 then? I.e. infusion needs to be hitlag affected?

k0l11:
Infusion is already getting hitlag affected, it's just that the infusion itself is not being logged anywhere. The easiest way to achieve this is to add a status.

srliao:
The infusion hitlag code is a bit hacky atm: 

https://github.com/genshinsim/gcsim/blob/17ecccadd21469e9a4057954be6ed8465a2a4cd1/pkg/core/player/player.go#L183-L192

Ideally I think infusion should just be refactored to using the `modifier` package (haven't though too much about the exact implementation details. 

Barring that though, I think rather than relying on status to log it, it should just use it's own dedicated logging. i.e. https://github.com/genshinsim/gcsim/blob/17ecccadd21469e9a4057954be6ed8465a2a4cd1/pkg/core/player/infusion/infusion.go#L9-L15

Is changed to something like 

```go
type WeaponInfusion struct {
	Key             string
	Ele             attributes.Element
	Tags            []combat.AttackTag
	Expiry          float64
	CanBeOverridden bool
        evt             events.Event
}
```

And update the evt expiry the same way all the other mods updates it (and additional event logs as needed)

k0l11:
closing this in favor of #958, because fixing infusions properly requires a big refactor

PR #950, opened 2022-09-23
implement burning

k0l11:
TODO in this PR:
- [x] fix burning dmg pyro app icd (should be global 2s icd shared between characters, but not between enemies)
- [x] fix burning damage sequence (there seems to be no limit to the amount of burning dmg that can be taken by a target in 2s)

TODO in later PR:
- [ ] write tests
- [ ] add OnBurning/Burning check to existing code where necessary (Dendro Resonance +30 EM, 4p CWoF, Collei A1+A4, Tighnari C4)

### DIFF
--- a/pkg/enemy/hitlag.go
+++ b/pkg/enemy/hitlag.go
@@ -66,6 +66,6 @@ func (e *Enemy) Tick() {
 	}
 	e.timePassed += left
 
-	e.Reactable.Tick()
 	queue.Run(&e.queue, e.timePassed)
+	e.Reactable.Tick()
 }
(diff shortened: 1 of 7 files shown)


pkg/enemy/hitlag.go
```diff
@@ -66,6 +66,6 @@ func (e *Enemy) Tick() {
 	}
 	e.timePassed += left
 
-	e.Reactable.Tick()
 	queue.Run(&e.queue, e.timePassed)
+	e.Reactable.Tick()
```
k0l11, line 70: this is to ensure that 8 ticks at 20 durability burning fuel are possible without messing with durability/decay rate

srliao:
Yoooooolo

PR #955, opened 2022-09-30
add dori implementation

skippi:


internal/characters/dori/burst.go
```diff
@@ -0,0 +1,101 @@
...
+
+			// dori self application
+			c.Core.Events.Emit(event.OnCharacterHurt, 0)
+			p, ok := c.Core.Combat.Player().(*avatar.Player)
+			if !ok {
+				panic("target 0 should be Player but is not!!")
+			}
+			idx := c.Core.Player.ActiveChar().Index
+			if c.Core.F > icdSrc[idx] + combat.ICDGroupResetTimer[combat.ICDGroupDoriBurst] {
+				p.ApplySelfInfusion(attributes.Electro, 25, 9.5*60) // TODO: find actual duration
```
k0l11, line 63: should theoretically be subject to tax since it applies electro via an attack and not via a modifier.

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/player"
+)
+
+var burstFrames []int
+
+const (
+	burstHitmark    = 28
+	burstHealPeriod = 12 * 60 / 6
```
k0l11, line 19: don't think this calc and variable is needed, because it's just every 120 frames

internal/characters/dori/cons.go
```diff
@@ -0,0 +1,112 @@
...
+		return false
+	}, "dori-c2")
+}
+
+// The character connected to the Jinni will obtain the following buffs based on their current HP and Energy:
+// ·When their HP is lower than 50%, they gain 50% Incoming Healing Bonus.
+// ·When their Energy is less than 50%, they gain 30% Energy Recharge.
+func (c *char) c4() {
+	// c4 should last for the duration of the burst
+	// lasts 12.5 second, ticks every 0.5s; adds mod to active char for 2s
```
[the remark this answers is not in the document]
skippi: Blame nayde.

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	c2Travel, ok := p["c2_travel"]
+	if !ok {
+		c2Travel = 10
+	}
+
+	for i := 0; i < 6; i++ {
+		c.Core.Tasks.Add(func() {
+			c.c2(c2Travel)
```
k0l11, line 80: missing a cons check

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,109 @@
...
+				Bonus:   snap.Stats[attributes.Heal],
+			})
+			// Energy regen to active char
+			active := c.Core.Player.ActiveChar()
+			active.AddEnergy("Alcazarzaray's Exactitude: Energy Regen", burstenergy[c.TalentLvlBurst()])
+		}, burstHealPeriod*i+11)
+	}
+	c.Core.Tasks.Add(func() {
+		if c.Base.Cons >= 4 {
+			c.c4()
```
k0l11, line 96: tbh, just change c4 to be part of the dmg ticks above, because it ticks at the exact same rate as the dmg ticks (removing the for loop and task queuing in c4 + adding calls to c4 in the task inside the burst dmg for loop in burst.go (cons check!) should do the trick)

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/player"
+)
+
+var burstFrames []int
+
+const (
+	burstHitmark    = 28
+	burstHealPeriod = 120
```
k0l11, line 19: if you're gonna have a variable for the heal tick rate, then add one for the dmg tick rate as well, because it feels weird to only have it for one of them. (or remove it and use 120 below)

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,109 @@
...
+			// dori self application
+			c.Core.Events.Emit(event.OnCharacterHurt, 0)
+			p, ok := c.Core.Combat.Player().(*avatar.Player)
+			if !ok {
+				panic("target 0 should be Player but is not!!")
+			}
+			idx := c.Core.Player.ActiveChar().Index
+			if c.Core.F > icdSrc[idx]+combat.ICDGroupResetTimer[combat.ICDGroupDoriBurst] {
+				dur := combat.Durability(25)
+				if p.AuraCount() > 0 {
```
k0l11, line 64: i don't understand why this check is needed, why not just supply 20 to ApplySelfInfusion?

internal/characters/dori/attack.go
```diff
@@ -0,0 +1,65 @@
...
+			StrikeType:         combat.StrikeTypeBlunt,
+			Element:            attributes.Physical,
+			Durability:         25,
+			Mult:               mult[c.TalentLvlAttack()],
+			HitlagFactor:       0.01,
+			HitlagHaltFrames:   attackHitlagHaltFrame[c.NormalCounter][i] * 60,
+			CanBeDefenseHalted: true,
+		}
+		// c6 key check
+		if c.StatusIsActive(c6key) {
```
k0l11, line 44: this part is unneeded now

internal/characters/dori/cons.go
```diff
@@ -0,0 +1,107 @@
...
+			return false
+		}
+		if c.StatusIsActive(c6icd) {
+			return false
+		}
+		if atk.Info.AttackTag != combat.AttackTagNormal {
+			return false
+		}
+
+		c.AddStatus(c6icd, 6, false) // 0.1s*60 icd
```
k0l11, line 94: should be subject to hitlag extension

internal/characters/dori/burst.go
skippi: Doesn't this matter since this is an attack and not a self infusion? Tax doesn't happen if there's an aura already on the unit.

skippi: Resolved: check is intentional since it is needed to apply 0.8x tax. Right now, `ApplySelfInfusion()` does not tax.

internal/characters/dori/attack.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/genshinsim/gcsim/internal/frames"
+	"github.com/genshinsim/gcsim/pkg/core/action"
+	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+)
+
+var (
+	attackFrames          [][]int
+	attackHitmarks        = [][]int{{27}, {19, 33}, {60}}
+	attackHitlagHaltFrame = [][]float64{{0.1}, {0, 0}, {0.08}}
```
k0l11, line 15: you need an extra array for CanBeDefenseHalted with [][]bool{{true}, {false, false}, {true}}, because rn you would have hitlag on N2-1 and N2-2. see sayu's code for example

internal/characters/dori/burst.go
```diff
@@ -0,0 +1,114 @@
...
+				Message: "Alcazarzaray's Exactitude: Healing",
+				Src:     bursthealpp[c.TalentLvlBurst()]*c.MaxHP() + bursthealflat[c.TalentLvlBurst()],
+				Bonus:   snap.Stats[attributes.Heal],
+			})
+			// Energy regen to active char
+			active := c.Core.Player.ActiveChar()
+			active.AddEnergy("Alcazarzaray's Exactitude: Energy Regen", burstenergy[c.TalentLvlBurst()])
+		}, 120*i+11)
+	}
+	c.Core.Tasks.Add(func() {
```
k0l11, line 99: remove this part entirely

internal/characters/dori/skill.go
```diff
@@ -0,0 +1,121 @@
...
+				c6key,
+				attributes.Electro,
+				180,
+				true,
+				combat.AttackTagNormal,
+				combat.AttackTagExtra,
+				combat.AttackTagPlunge,
+			)
+		}
+	}, skillFrames[action.ActionAttack]) // TODO:It activates on the attack cancel frames?
```
k0l11, line 77: it activates on cast, but adds the cast frames to the duration. (I see "Duration" (probably 3s) + 0.8s in dm)

internal/characters/dori/attack.go
```diff
@@ -0,0 +1,61 @@
...
+			Durability:         25,
+			Mult:               mult[c.TalentLvlAttack()],
+			HitlagFactor:       0.01,
+			HitlagHaltFrames:   attackHitlagHaltFrame[c.NormalCounter][i] * 60,
+			CanBeDefenseHalted: true,
+		}
+		c.QueueCharTask(func() {
+			c.Core.QueueAttack(
+				ai,
+				combat.NewCircleHit(c.Core.Combat.Player(), 1, false, combat.TargettableEnemy),
```
k0l11, line 46: her NA are radius 2

internal/characters/dori/cons.go
```diff
@@ -0,0 +1,103 @@
...
+			Abil:       "Special Franchise",
+			AttackTag:  combat.AttackTagNone,
+			ICDTag:     combat.ICDTagDoriC2,
+			ICDGroup:   combat.ICDGroupDefault,
+			StrikeType: combat.StrikeTypeDefault,
+			Element:    attributes.Electro,
+			Durability: 25,
+			Mult:       0.5,
+		}
+		c.Core.QueueAttack(ai, combat.NewDefSingleTarget(c.Core.Combat.DefaultTarget, combat.TargettableEnemy), 0, travel)
```
k0l11, line 37: should be sphere (circle in sim) with radius 1

k0l11 (approved):

internal/characters/dori/skill.go
```diff
@@ -0,0 +1,118 @@
...
+			ICDGroup:   combat.ICDGroupDefault,
+			StrikeType: combat.StrikeTypeDefault,
+			Element:    attributes.Electro,
+			Durability: 25,
+			Mult:       skillAfter[c.TalentLvlSkill()],
+		}
+		for i := 0; i < c.afterCount; i++ {
+			c.Core.QueueAttack(
+				ae,
+				combat.NewDefSingleTarget(c.Core.Combat.DefaultTarget, combat.TargettableEnemy),
```
k0l11, line 112: should have the same hitbox data as her skill

internal/characters/dori/cons.go
```diff
@@ -0,0 +1,107 @@
...
+			return false
+		}
+
+		c.AddStatus(c6icd, 6, true) // 0.1s*60 icd
+		// heal party members
+
+		c.Core.Player.Heal(player.HealInfo{
+			Caller:  c.Index,
+			Target:  -1,
+			Message: "Dori-c6-onNormal",
```
k0l11, line 100: rename to "dori-c6-onnormal" or something

PR #962, opened 2022-10-01
add wood weapons

skippi:
- [x] Need to decide on use case for this weapon. Should we spawn a leaf gadget to pick up? Or should buffs be set to a character via param? Etc.
- [x] Verify hitlag extension on buffs.


srliao:
@skippi how about a pickup_delay param, where if not set it does not pick up and if set to any positive number > 0 (or maybe >= 0), then it picks up after that many frames?

k0l11:
I'm ok with that param

internal/weapons/claymore/forestregalia/forestregalia.go
```diff
@@ -0,0 +1,93 @@
...
+type Weapon struct {
+	Index int
+}
+
+func (w *Weapon) SetIndex(idx int) { w.Index = idx }
+func (w *Weapon) Init() error      { return nil }
+
+const (
+	icdKey  = "forestregalia-icd"
+	buffKey = "forest-sanctuary"
```
k0l11, line 30: Pls rename the buffKey variables to the weapon names to keep in line with most of the other weapon buffs. Also, having the Forest Regalia and Sapwood Blade buff names overlap isn't good because that means these buffs overwrite each other instead of stacking if a character happens to pick up a leaf from both Forest Regalia and Sapwood Blade.

k0l11: Nvm, I just noticed that the description says these buffs can't stack...

k0l11 (approved):

k0l11:
resolves #1091

PR #963, opened 2022-10-01
add bloom five dendro core limit

skippi:
There should never be more than 5 dendro cores at any time.

### DIFF
--- a/pkg/reactable/bloom.go
+++ b/pkg/reactable/bloom.go
@@ -76,6 +76,21 @@ type DendroCore struct {
 
 func (r *Reactable) addBloomGadget(a *combat.AttackEvent) {
 	r.core.Tasks.Add(func() {
+		first := -1
+		bloomCount := 0
+		for i := 0; i < r.core.Combat.GadgetCount(); i++ {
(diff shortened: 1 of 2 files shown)


pkg/reactable/bloom.go
```diff
@@ -76,6 +76,21 @@ type DendroCore struct {
 
 func (r *Reactable) addBloomGadget(a *combat.AttackEvent) {
 	r.core.Tasks.Add(func() {
+		first := -1
+		bloomCount := 0
+		for i := 0; i < r.core.Combat.GadgetCount(); i++ {
```
imring, line 81: i don't think it's right, cuz GadgetCount has a condition if gadget is nil
if the array of gadgets is `[nil, gadget, gadget]`, GadgetCount returns 2, since the first element is nil and your loop will only cycle the first two elements (`[nil, gadget]`)

imring: either rewrite the condition:
```go
if r.core.Combat.GadgetCount() >= 5 {
  // remove the first element
}
```
or rewrite RemoveGadget (get rid of nil)

srliao: fixed in #970

Issue #611, opened 2022-06-23
fix AoE/hitbox data


k0l11:
The AoE/hitboxes of abilities in the sim are often arbitrary. This should be eventually changed to match the actual data for all characters.
- [ ] albedo
- [ ] aloy
- [ ] amber
- [ ] ayaka
- [ ] ayato
- [ ] barbara
- [ ] beidou
- [ ] bennett
- [ ] candace
- [ ] chongyun
- [ ] collei
- [ ] cyno
- [ ] diluc
- [ ] diona
- [ ] dori
- [ ] eula
- [x] fischl
- [ ] ganyu
- [ ] gorou
- [ ] heizou
- [ ] hutao
- [ ] itto
- [ ] jean
- [ ] kaeya
- [ ] kazuha
- [ ] keqing
- [ ] klee
- [ ] kokomi
- [ ] kuki
- [ ] layla
- [ ] lisa
- [ ] mona
- [ ] nahida
- [ ] nilou
- [ ] ningguang
- [ ] noelle
- [ ] qiqi
- [ ] raiden
- [ ] razor
- [ ] rosaria
- [ ] sara
- [ ] sayu
- [ ] shenhe
- [ ] sucrose
- [ ] tartaglia
- [ ] thoma
- [ ] tighnari
- [ ] traveleranemo
- [ ] travelerdendro
- [ ] travelerelectro
- [ ] travelergeo
- [ ] venti
- [ ] xiangling
- [ ] xiao
- [ ] xingqiu
- [ ] xinyan
- [ ] yaemiko
- [ ] yanfei
- [ ] yelan
- [ ] yoimiya
- [ ] yunjin
- [ ] zhongli

originally about Hu Tao CA AoE:
When uncancelled, you could maybe make it hit multiple enemies if you're lucky but when cancelled with Jump/Dash this choice becomes very questionable.

RarePossum:
Also applies to Kazuha's N1 at least, maybe other attacks.

JohnJillky:
Yeah a lot of attacks seem to be unrealistically categorized as AoE. E.g., I can have Eula autoing on 5 enemies and gcsim thinks she hits all 5. In practice, you might get more like 2-3, and often 1. I think the "AoE-ness" of attacks should be adjustable

PR #965, opened 2022-10-02
adjust yaemiko hitbox data

skippi:
Fixes yaemiko's part of #611. I'm hoping to get some discussion going for how to approach this, because I think these AoE fixes are necessary with the addition of bloom.

- [ ] Figure out how we want to handle box-type rotation (EG, yaemiko facing a target to change hitbox of her normals/charge).


skippi:
Closing since it will be covered in future PR.

PR #970, opened 2022-10-02
update bloom limit - max 5

srliao:
continuation of the PR here: https://github.com/genshinsim/gcsim/pull/963

changed it so that the limit is enforced by core and force gadgets to have types. should make it easier for future gadgets that also have limits

### DIFF
--- a/internal/tests/reactable/bloom_test.go
+++ b/internal/tests/reactable/bloom_test.go
@@ -1,9 +1,9 @@
 package reactable_test
 
 import (
+	"math"
 	"testing"
 
-	"github.com/genshinsim/gcsim/pkg/core"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 	"github.com/genshinsim/gcsim/pkg/core/combat"
 	"github.com/genshinsim/gcsim/pkg/reactable"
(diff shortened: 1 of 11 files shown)


internal/tests/reactable/bloom_test.go
```diff
@@ -126,5 +126,93 @@ func TestECBloom(t *testing.T) {
...
+		if v == nil || v.GadgetTyp() != combat.GadgetTypDendroCore {
+			continue
+		}
+		if v.Src() < f {
+			f = v.Src()
+			oldest = i
+		}
+	}
+	og := c.Combat.Gadget(oldest)
+	if og.Src() != 3+reactable.DendroCoreDelay {
```
skippi, line 214: Consider: checking for the oldest bloom by `key` instead of `src`. The key should be `2` in this test.

pkg/core/combat/gadget.go
```diff
@@ -1,33 +1,86 @@
...
+
+func (h *Handler) RemoveGadget(key int) bool {
+	for i, v := range h.gadgets {
+		if v != nil && v.Key() == key {
+			//set to nil for now; we should clean up every so often???
+			//TODO: how often do we clean out nil entries? if at all?
+			h.gadgets[i] = nil
+			return true
+		}
 	}
```
skippi, line 37:
```suggestion
	h.ReplaceGadget(key, nil)
```

pkg/gadget/gadget.go
```diff
@@ -11,32 +11,39 @@ import (
...
+	if g.OnKill != nil {
+		g.OnKill()
 	}
-	g.core.Combat.RemoveGadget(g.Index())
+	g.core.Combat.RemoveGadget(g.Key())
 }
 
 func (g *Gadget) Type() combat.TargettableType { return combat.TargettableGadget }
+func (g *Gadget) Src() int                     { return g.src }
+func (g *Gadget) GadgetTyp() combat.GadgetTyp  { return g.gadgetTyp }
```
skippi, line 46: Was `GadgetTyp` mispelled here intentionally?

pkg/gadget/gadget.go
```diff
@@ -51,8 +58,9 @@ func (g *Gadget) Tick() {
 	}
 	if g.Duration != -1 && g.Duration > 0 {
 		g.Duration--
-		if g.Duration == 0 {
-			g.Kill()
+		if g.Duration == 0 && g.OnExpiry != nil {
+			g.OnExpiry()
+			g.core.Combat.RemoveGadget(g.Key())
```
skippi, line 63: The gadget still needs to be removed even if `OnExpiry == nil`.

skippi (changes requested):

internal/tests/reactable/bloom_test.go
srliao: should be the same. src might be more reliable because each reaction is 2 frames apart. no guarantee that key will always be 2. if we do key should then store the key

skippi (approved):

PR #976, opened 2022-10-03
fix yelan skill not procing damage properly

skippi:
The damage function needs to be delayed by 1f due to how we handle tasks in dendro update.


srliao:
"fixed" the task queue instead

PR #990, opened 2022-10-04
fix dori messing up string print names in keys

skippi:


pkg/shortcut/characters.go
```diff
@@ -121,4 +120,5 @@ var CharNameToKey = map[string]keys.Char{
 	"heizou":            keys.Heizou,
 	"collei":            keys.Collei,
```
srliao, line 121: unrelated but collei appears to be after tighnari in the keys above

srliao (approved):

PR #996, opened 2022-10-06
change bloom dendro core delay to 20f

skippi:
Based off these videos, the dendro core is counted as a gadget starting at ~20f from the bloom reaction trigger. This is also the time when the 5 dendro core limit is checked. The delay on damage is likely caused by the 12f linger of the core after being ruptured, hyperbloomed, or burgeoned.

The spawn delay is probably not always 20f. My hypothesis is that it is variable depending on the spawn location of the gadget. However for the sim, 20f suffices.

https://discord.com/channels/845087716541595668/869210750596554772/1027611051124064356

### DIFF
--- a/pkg/reactable/bloom.go
+++ b/pkg/reactable/bloom.go
@@ -10,7 +10,7 @@ import (
 	"github.com/genshinsim/gcsim/pkg/gadget"
 )
 
-const DendroCoreDelay = 45
+const DendroCoreDelay = 20
 
 func (r *Reactable) tryBloom(a *combat.AttackEvent) {
 	//can be hydro bloom, dendro bloom, or quicken bloom


skippi:
Do we want to add any logging while we're here? Without knowledge of the underlying system, it's difficult to debug how the dendro cores spawn and how the limit is hit. I've suggested adding `LogGadgetEvent` here and will continue to do so.

srliao:
> Do we want to add any logging while we're here? Without knowledge of the underlying system, it's difficult to debug how the dendro cores spawn and how the limit is hit. I've suggested adding `LogGadgetEvent` here and will continue to do so.

Up to you. Either in this PR or another one let me know. Otherwise looks good to merge

srliao (approved):

srliao:
actually i'll just merge as is and we can do another PR for logging later. at least this can go live now

PR #1019, opened 2022-10-18
fischl radius update

soloxcx:
fischl a4 now has radius of 0.5m
fischl a4 now hits gadgets
fischl a4 now procs on hyperbloom trigger
fischl q now has radius of 0.5m
fischl c4 now has radius of 5m
fischl e summon now has radius of 2m by default, 3m if c2
fischl e summon now centers on primary target

### DIFF
--- a/internal/characters/fischl/skill.go
+++ b/internal/characters/fischl/skill.go
@@ -35,11 +35,13 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 		Mult:       birdSum[c.TalentLvlSkill()],
 	}
 
+	var radius float64 = 2
 	if c.Base.Cons >= 2 {
 		ai.Mult += 2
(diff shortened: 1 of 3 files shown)


internal/characters/fischl/skill.go
```diff
@@ -35,11 +35,13 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 		Mult:       birdSum[c.TalentLvlSkill()],
 	}
 
+	var radius float64 = 2
```
skippi, line 38: Consider:
```suggestion
	radius := 2.0
```

skippi (commented):
Left a note, but otherwise looks good to me. Any thoughts @k0l11 ?

k0l11 (approved):

PR #1020, opened 2022-10-18
refactor xiao a1 into modifier

### LINKED ISSUE #1000: xiao a1 refactored as a mod
image

right now the user must enable the `character` debug setting to see the effect of xiao's a1, because it is written directly into his snapshot AttackInfo [here](https://github.com/genshinsim/gcsim/blob/084b12073c0ea675b630177380e3363fb263f445/internal/characters/xiao/xiao.go#L84). it won't show in a list of damage mods which is perhaps confusing

skippi:
This fixes #1000.

### DIFF
--- a/internal/characters/xiao/asc.go
+++ b/internal/characters/xiao/asc.go
@@ -2,7 +2,6 @@ package xiao
 
 import (
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
-	"github.com/genshinsim/gcsim/pkg/core/combat"
 	"github.com/genshinsim/gcsim/pkg/core/player/character"
 	"github.com/genshinsim/gcsim/pkg/modifier"
 )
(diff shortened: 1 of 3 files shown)


internal/characters/xiao/burst.go
```diff
@@ -30,13 +30,14 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 	// TODO: No indication of that in library though
 	c.AddStatus(burstBuffKey, 900+burstStart, true)
```
skippi, line 31: I find it a bit strange that the burst buff starts as soon as he casts. So in this case, does the A1 buff start ticking on cast too?

k0l11: yes

internal/characters/xiao/asc.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/player/character"
+	"github.com/genshinsim/gcsim/pkg/modifier"
+)
+
+const a1Key = "xiao-a1"
+
+func (c *char) a1() {
+	m := make([]float64, attributes.EndStatType)
+	c.AddAttackMod(character.AttackMod{
```
k0l11, line 14: you aren't doing checks using atk and t, so no need to make this an AttackMod. should be a statmod instead

internal/characters/xiao/asc.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/player/character"
+	"github.com/genshinsim/gcsim/pkg/modifier"
+)
+
+const a1Key = "xiao-a1"
+
```
k0l11, line 11: missing talent description comment

k0l11 (approved):

PR #1026, opened 2022-10-19
change .mods and .status to return StatusDuration()

skippi:
Sometimes it's useful to know how long a status lasts before it expires. This also isn't a breaking change for most configs, so it should be safe enough to apply.
- [ ] update docs at some point


k0l11:
this is a breaking change for any config that checks for == 1

k0l11 (approved):

PR #1029, opened 2022-10-19
Added param to determine whether Sucrose Venti and Anemo MC self absorb

RarePossum:
Added `self_absorb` param to Venti, Sucrose and Anemo MC bursts. With value `0`, characters will not check themselves for self absorption, while other values will. By Default, Venti and Anemo MC are disabled while Sucrose is enabled

https://gcsim.app/v3/viewer/share/6ade3731-2446-4c6c-9835-900594fc921b


k0l11:
pls rebase your branch

PR #1031, opened 2022-10-19
fix dori c2 infinite stacking per burst cast

skippi:
When Dori casted her burst, it would accidently attach six OnHeal event handlers that would trigger her C2 OnHeal. We fix this by removing the event and just proc'ing her C2 six times.

This also confirms that the C2 still works even if Dori is max HP.


k0l11 (approved):

skippi:
Context for if C2 works at max HP:
https://imgur.com/ox13neD
https://cdn.discordapp.com/attachments/845087716541595671/1032332088487194654/Genshin_Impact_2022.10.19_-_19.37.52.02_Trim.mp4
https://discord.com/channels/845087716541595668/845087716541595671/1032331062354907217

PR #1040, opened 2022-10-20
attack handling - initial refactor

k0l11:
some relevant parts in the core code:
attack(): https://github.com/genshinsim/gcsim/blob/58a77b3930638aeb81adee69ee9487e9f4b02fc5/pkg/core/combat/attack.go#L12
target version of Attack() and ApplyAttack(): https://github.com/genshinsim/gcsim/blob/58a77b3930638aeb81adee69ee9487e9f4b02fc5/pkg/enemy/attack.go#L14
bloom gadget version of Attack() and ApplyAttack(): https://github.com/genshinsim/gcsim/blob/58a77b3930638aeb81adee69ee9487e9f4b02fc5/pkg/reactable/bloom.go#L121

part 1:
- adjusts all hitbox functions (NewDefSingleTarget, NewCircleHit, NewDefBoxHit) to no longer require the following parameters: 
  - self (was used for swirl aoe and bloom self damage)
    - swirl aoe: represented via NoSelfHarm field in AttackInfo struct that has to explicitly be set to true in this case
    - bloom self damage: represented via PlayerHarm field in AttackInfo. if this is set to true then the attack will only target the player
  - a list of TargettableTypes (combat.TargettablePlayer, combat.TargettableEnemy, combat.TargettableGadget)
    - combat.TargettablePlayer: PlayerHarm has to be set to true for an attack to be able to hit the Player. If that's the case, then currently no other target types will be hit by the attack (subject to change, might want another field in AttackInfo to specify this behaviour).
    - combat.TargettableEnemy and combat.TargettableGadget are considered by every attack that doesn't have PlayerHarm set to true
- adjusts all calls to the hitbox functions to reflect the changes:
  - refactors Ocean Hued Clam to no longer set a DamageSrc and self to true, because that isn't necessary
  - bloom, hyperbloom, burgeon and nilou bountiful core attacks now set PlayerHarm for the self damage
  - the swirl aoe attack now sets NoSelfHarm to make it not deal dmg to the DamageSrc

part 2:
- adjusts AttackEvent and AttackPattern
- refactors attack()
- sets basis for player damage, adjusts related events and subscriptions
- fixes AbsorbCheck not checking for element on gadgets

TODO:
- [X] make different attack() functions for combat.TargettableEnemy and combat.TargettableGadget respectively to address the event (OnDamage shouldn't be emitted on hits to gadgets) and logging (no more 0 dmg logs to gadgets in the debug log) problem
  - no logs, no emits, no callbacks for combat.TargettableGadget for now
    - might want to introduce an extra log category for this (probably not part of this PR)
    - basically only calls the gadget's implementations of Attack() and ApplyAttack()
- [X] AbsorbCheck needs to also check gadgets for auras (DMC Q can hold Cryo and Anemo characters should be able to pick that up for their absorptions)


pkg/reactable/bloom.go
```diff
@@ -133,12 +143,16 @@ func (s *DendroCore) Attack(atk *combat.AttackEvent, evt glog.Event) (float64, b
 		x, y := s.Gadget.Pos()
 		enemies := s.Core.Combat.EnemyByDistance(x, y, combat.InvalidTargetKey)
 		if len(enemies) > 0 {
-			s.Core.QueueAttack(ai, combat.NewCircleHit(s.Core.Combat.Enemy(enemies[0]), 1, false, combat.TargettableEnemy), -1, 5)
+			ap := combat.NewCircleHit(s.Core.Combat.Enemy(enemies[0]), 1)
+			s.Core.QueueAttack(ai, combat.NewCircleHit(s.Core.Combat.Enemy(enemies[0]), 1), -1, 5)
```
skippi, line 147: The `ap` variable is created here despite not being used until later. I assume it's supposed to be used in this line?

skippi (changes requested):

skippi (approved):

PR #1041, opened 2022-10-21
update db warning

soloxcx:
adjust db warning to tell users that it is not a dps leaderboard.

based on discussion here: https://discord.com/channels/845087716541595668/1006934337612173413/1032717676830801930.

not 100% sold on my wording, open to suggestions on the best way to phrase it


soloxcx:
I had initially removed the text color to make it white -- but it was actually my dark mode browser skin misleading me into thinking the original colors were different

srliao (approved):

PR #1044, opened 2022-10-22
UI refactor

srliao:
This PR is to separate out the UI into a separate node package, isolating it from the wasm logic. This way the UI can be used for both web, desktop, and remote server (so @unleashurgeek can run sims off his iPad on his couch while watching League games)


unleashurgeek:
closing in favor of https://github.com/genshinsim/gcsim/pull/1044

PR #1059, opened 2022-10-30
fix LogElementEvent not showing quicken bloom changes

skippi:
Since quicken bloom is programmed as a 0 delay task, we need to defer LogElementEvent to account for durability changes.


unleashurgeek (dismissed):

k0l11:
pls rebase this PR to resolve conflict

k0l11:
this fixes the quick bloom issue, but it breaks the element logs if two reactions happen on the same frame. see <https://discord.com/channels/845087716541595668/983391844631212112/1191526441276411984>. i will include this problem as part of #390

PR #1105, opened 2022-11-11
add custom monster targets

skippi:
- [ ] Decide which mobs should be included in gcsim.
- [ ] Finalize syntax.
- [ ] Design storage for monster data.


pkg/gcs/ast/parseTarget.go
```diff
@@ -46,6 +46,30 @@ func parseTarget(p *Parser) (parseFn, error) {
...
+				}
+				params, err := p.acceptOptionalTargetParams()
+				if err != nil {
+					return nil, err
+				}
+				err = enemy.ConfigureTarget(&r, item.Val, params)
+				if err != nil {
+					return nil, err
+				}
+				p.res.Settings.DamageMode = true
```
unleashurgeek, line 62: With this current parsing logic, I think target configuration will be order dependent? Not sure if this is something we want or how we'd want to account for this if not (some struct union logic I guess would do the trick?)

IE: the following will create a dummy with resistance `0.1` in all categories
```
target lvl=100 resist=0.2 type=dummy;
```

whereas the following will result in resistances set to `0.2`
```
target lvl=100 type=dummy resist=0.2;
```

pkg/enemy/attack.go
```diff
@@ -155,29 +165,65 @@ func (e *Enemy) attack(atk *combat.AttackEvent, evt glog.Event) (float64, bool)
...
-				100, //TODO: should be subject to global delay maybe??
+				100, // TODO: should be subject to global delay maybe??
 			)
 		}
 	}
 
+	if e.prof.ParticleDrops != nil {
+		e.tryHPDropParticle()
+	}
+
```
unleashurgeek, line 176: NIT: is there a way we can merge the `ParticleDropThreshold` logic with `tryHPDropParticle`? Looks like your current logic makes these two settings mutually exclusive but wondering if there is a better way which could simplify logic here

pkg/enemy/enemy.go
```diff
@@ -19,6 +19,12 @@ type EnemyProfile struct {
 	ParticleDropThreshold float64                        `json:"-"` // drop particle every x dmg dealt
 	ParticleDropCount     float64                        `json:"-"`
 	ParticleElement       attributes.Element             `json:"-"`
+	ParticleDrops         []HpDrop                       `json:"-"`
+	ResistFrozen          bool                           `json:"-"`
+	HpBase                float64                        `json:"-"`
+	HpGrowCurve           int                            `json:"-"`
+	Id                    int                            `json:"-"`
+	MonsterName           string                         `json:"-"`
```
unleashurgeek, line 27: Nice to have: a `modified` bool which will tell me if this `EnemyProfile` deviated from the default. Then in UI I can prefix that enemy as modified/custom to make it clear it is not default

pkg/gcs/ast/parseTarget.go
skippi: I am for this idea. Even though it's a breaking change, it's only "broken" for people who abused syntax.

pkg/enemy/attack.go
skippi: With the struct processing from above, I think fixing this is relatively easy. Working on it.

pkg/enemy/enemy.go
unleashurgeek: Yeah so my thought is either we expose the original `EnemyProfile` in the stats as well and show a diff, or have a `modified` bool and we just indicate that "something has deviated from the original profile" but leave it to the viewer to figure out what

k0l11:
continued in #1961

Issue #905, opened 2022-09-05
Can't reference target status with conditionals


skippi:
Useful for mona-like status where the target is the holder. We'll need to decide on a syntax.

PR #1106, opened 2022-11-11
add .target0 conditional namespace

skippi:
This includes the following conditionals:
- `.target0.mods.[name]`: status/res/def/mods check
- `.target0.element.[name]`: durability check

- [ ] Verify syntax design.
- [ ] Figure out how to handle deprecation better.


k0l11:
closes #905

Charlie-Zheng:
I'd like to mention that the set_target_pos syntax is 1 indexed. It would be confusing for conditionals to be 0 indexed and the set_target_pos to be 1 indexed.

PR #1107, opened 2022-11-11
Additional information for .json output

Kisa75:
Added additional information to .json output, for example relevant simulator settings or target data.

### DIFF
--- a/pkg/simulator/simulator.go
+++ b/pkg/simulator/simulator.go
@@ -151,14 +151,15 @@ func GenerateResult(cfg string, simcfg *ast.ActionList, opts Options) (result.Su
 		//    Minor: increase if new schema is backwards compatible with previous
 		//        Ex - added new data for new graph on UI. UI still functional if this data is missing
 		// Increasing the version will result in the UI flagging all old sims as outdated
-		SchemaVersion:    result.Version{Major: 4, Minor: 0}, // MAKE SURE UI VERSION IS IN SYNC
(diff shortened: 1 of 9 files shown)


pkg/simulator/simulator.go
```diff
@@ -151,14 +151,15 @@ func GenerateResult(cfg string, simcfg *ast.ActionList, opts Options) (result.Su
...
-		BuildDate:        opts.BuildDate,
-		MaxIterations:    simcfg.Settings.Iterations,
-		Config:           cfg,
-		SampleSeed:       strconv.FormatUint(uint64(CryptoRandSeed()), 10),
-		TargetDetails:    simcfg.Targets,
-		InitialCharacter: simcfg.InitialChar.String(),
+		SchemaVersion:     result.Version{Major: 4, Minor: 0}, // MAKE SURE UI VERSION IS IN SYNC
+		SimVersion:        opts.Version,
+		BuildDate:         opts.BuildDate,
+		SimulatorSettings: simcfg.Settings,
```
unleashurgeek, line 157: Add back max iterations. This either needs to be added back here, or you can go add expose `Iterations` through `SimulatorSettings`, but then you'll also need to update the UI to change where it is reading the max iterations from.
```suggestion
    MaxIterations:     simcfg.Settings.Iterations,
		SimulatorSettings: simcfg.Settings,
```

unleashurgeek (approved):

PR #1129, opened 2022-12-04
fix certain shield strength mods getting removed too early

k0l11:


internal/characters/zhongli/asc.go
```diff
@@ -5,6 +5,6 @@ func (c *char) a1() {
 		if c.Tags["shielded"] == 0 {
 			return 0, false
 		}
```
unleashurgeek, line 7: nit: don't need this condition since `c.Tags["a1"] = 0` when shield is not up

internal/characters/thoma/asc.go
```diff
@@ -12,7 +12,7 @@ func (c *char) a1() {
 		if !c.StatusIsActive("thoma-a1") {
 			return 0, false
 		}
-		return float64(c.a1Stack) * 0.05, true
```
unleashurgeek, line 15: similar to zhongli, shouldn't need the ifs and should just rely on `a1Stack = 0` at the right times. What concerns me is looking through the implementation we only reset `a1Stack` to 0 on character swap? So if I shield -> get max a1 stack -> stay on thoma and use his skill again it will start at max stacks. Not finding anything on KQM TC for how long these stacks are supposed to last

unleashurgeek (approved):

k0l11:
closed in favor of addressing this in a PR that resolves #835

Issue #913, opened 2022-09-05
Mona C1 requires ReactBonusMod refactor


k0l11:
Mona C1 adds a ReactBonusMod with a timer for Vaporize, Electro-Charged and Hydro Swirl DMG. This ReactBonusMod should snapshot for Vaporize on abilities like Xiangling Q, but the Electro-Charged and Hydro Swirl part should not snapshot. It is not possible to model this interaction at the moment.
See https://library.keqingmains.com/evidence/characters/hydro/mona#mona-c1-snapshot-for-vape.

k0l11:
should probably add the different react bonuses as stats to the stats array

PR #1135, opened 2022-12-08
fix reactBonus

imring:
for many reactions the reaction bonus is not displayed.
for example: https://gcsim.app/v3/viewer/share/8818ff23-055f-4669-9943-cf3a9c7afdfb (491 frame)
image
with fix:
image


k0l11:
Idk about merging this considering that ReactBonus mod is due for a rewrite (#913) and that this would log the ReactBonus at the time of the attack hitting, which would ignore its snapshotting properties for some reactions.

PR #1141, opened 2022-12-12
add faruzan implementation

skippi:
- [x] A4 might need to be reworked.


k0l11:
missing updates to web app related to faruzan

internal/characters/faruzan/asc.go
```diff
@@ -0,0 +1,53 @@
...
+func (c *char) a4() {
+	count := 0
+	c.Core.Events.Subscribe(event.OnEnemyHit, func(args ...interface{}) bool {
+		atk := args[1].(*combat.AttackEvent)
+		if atk.Info.Element != attributes.Anemo {
+			return false
+		}
+
+		switch atk.Info.AttackTag {
+		case combat.AttackTagNormal,
```
k0l11, line 24: missing combat.AttackTagElementalArtHold

internal/characters/faruzan/asc.go
```diff
@@ -0,0 +1,53 @@
...
+			combat.AttackTagPlunge,
+			combat.AttackTagElementalArt,
+			combat.AttackTagElementalBurst:
+		default:
+			return false
+		}
+
+		char := c.Core.Player.ByIndex(atk.Info.ActorIndex)
+		if char.StatusIsActive(burstBuffKey) && !char.StatusIsActive(a4ICDKey) {
+			char.AddStatus(a4Key, 6, false)
```
k0l11, line 35: both of these should only apply to Faruzan (i.e. A4 icd is shared between characters), so please rework A4 to reflect this. In that case, you can drop `count`.

internal/characters/faruzan/asc.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
+)
+
+const (
+	a4Key    = "faruzan-a4"
+	a4ICDKey = "faruzan-a4-icd"
+)
+
```
k0l11, line 14: nit: Missing a4 description? You added ability descriptions for all the other functions.

internal/characters/faruzan/burst.go
```diff
@@ -0,0 +1,176 @@
...
+				if !ok {
+					continue
+				}
+				applyBurstShred(trg)
+			}
+		}, 43+i)
+		count += 1
+	}
+
+	// In-game refreshes 0.1s. We give buff every 69f to reduce spam.
```
k0l11, line 91: Please rewrite this to only refresh every 4s (right before buff would expire without hitlag) to further reduce spam.

internal/characters/faruzan/burst.go
```diff
@@ -0,0 +1,176 @@
...
+		if c.burstSrc != currSrc {
+			return
+		}
+		if !combat.WillCollide(field, c.Core.Combat.Player(), 0) {
+			return
+		}
+		for _, char := range c.Core.Player.Chars() {
+			c.applyBurstBuff(char)
+		}
+	}, 43+int(duration/6.0)*6)
```
k0l11, line 118: You can simplify this to `43+duration`

internal/characters/faruzan/burst.go
```diff
@@ -0,0 +1,176 @@
...
+				return
+			}
+			for _, char := range c.Core.Player.Chars() {
+				c.applyBurstBuff(char)
+			}
+		}, 43+i)
+	}
+
+	// Last refresh to account for 0.1s tick period
+	c.Core.Tasks.Add(func() {
```
k0l11, line 108: Please consider reducing redundancy by defining a function variable up above like so:
`buffFunc := func() { ... }`

internal/characters/faruzan/burst.go
skippi: It was done this way since the refresh time is one frame off. For example, with the truncation math above, these would be the refresh points:

C2: 1105 -> 1104
C0: 765 -> 764

But I can see that this is slightly redundant since it's a one frame difference. I will be addressing this in the next set of commits.

internal/characters/xiangling/guoba.go
```diff
@@ -88,21 +89,28 @@ func (p *panda) HandleAttack(atk *combat.AttackEvent) float64 {
...
 		return 0, false
 	}
 
-	p.Core.Log.NewEvent("guoba hit by sucrose E", glog.LogCharacterEvent, p.c.Index)
+	// don't take damage, trigger swirl reaction only on sucrose E or faruzan E
+	switch p.Core.Player.Chars()[atk.Info.ActorIndex].Base.Key {
+	case keys.Sucrose:
+		p.Core.Log.NewEvent("guoba hit by sucrose E", glog.LogCharacterEvent, p.c.Index)
+	case keys.Faruzan:
+		if atk.Info.Abil != faruzan.SkillName {
```
k0l11, line 105: `SkillName` and `faruzan e` in the logs is ambigious considering that her pressurized collapse can be triggered via multiple ways. Please consider renaming this variable and the log below to indicate that it's the Pressurized Collapse.

internal/characters/faruzan/asc.go
```diff
@@ -0,0 +1,58 @@
...
+			combat.AttackTagElementalArt,
+			combat.AttackTagElementalArtHold,
+			combat.AttackTagElementalBurst:
+		default:
+			return false
+		}
+
+		active := c.Core.Player.ByIndex(atk.Info.ActorIndex)
+		if active.StatusIsActive(burstBuffKey) && !c.StatusIsActive(a4ICDKey) {
+			c.AddStatus(a4Key, 6, false)
```
k0l11, line 42: Do even need this a4Key? Nothing in the descsription indicates you need this. If you remove it you can then further reduce nesting by making the burst buff and icd check into early returns.

internal/characters/faruzan/cons.go
```diff
@@ -0,0 +1,74 @@
...
+				return nil, false
+			}
+			return m, true
+		},
+	})
+}
+
+const c6ICDKey = "faruzan-c6-icd"
+
+func (c *char) c6Collapse() {
```
k0l11, line 56: this shouldn't proc on 0 dmg, need to add a check like: 
```
dmg := args[2].(float64)
...
if dmg == 0 {
	return false
}
```

internal/characters/faruzan/burst.go
```diff
@@ -0,0 +1,168 @@
...
+				if !ok {
+					continue
+				}
+				applyBurstShred(trg)
+			}
+		}, 43+i)
+		count += 1
+	}
+
+	// In-game refreshes 0.1s. We give buff every 69f to reduce spam.
```
k0l11, line 91: comment is outdated now

k0l11: reminder to add web related stuff btw

k0l11 (approved):

PR #1148, opened 2022-12-17
gcsl rework - add functional programming support

imring:
original topic: https://trello.com/c/8umdzCNv/63-syntax-improvements

## Summary of Changes
- changed `varMap` in eval to map to `Obj` instead of `number` to allow for variable types (more than just numbers) <https://github.com/genshinsim/gcsim/pull/1148/commits/226d6d7eb2fedb04f1d08a67f064da10402328db>
- change function AST node to an `Expr` so that functions can be used in expressions such as `return somefn` <https://github.com/genshinsim/gcsim/pull/1148/commits/003404aa64ee1474e559f840117c997734aada66>
- split built-in functions to separate map <https://github.com/genshinsim/gcsim/pull/1148/commits/7960704464a57846ba6425b10a8e30b7d3ed705e>
- change parseCall to allow for `somefn()()` (i.e. where `somefn` is a function returning another function. This is enabled by the earlier change to functions to be an `Expr` instead of a `Stmt` (so now functions can be evaluated recursively via `evalExpr`) <https://github.com/genshinsim/gcsim/pull/1148/commits/b05e90a66c2b762ed6a8b8ef9e96d69be7d36da1>
- added boolean variable support (eval code change not used in syntax yet) <https://github.com/genshinsim/gcsim/pull/1148/commits/56dfdc3752c8a7021e48d8bb9a0ddf6e36703e10>
- added `onTick` and various math sysfuncs <https://github.com/genshinsim/gcsim/pull/1148/commits/13172611212bfb4e162be74eefc84dea265b790c>
- added simple map structure to AST to allow for character actions to be parsed as system functions calls instead of special ActionStmt nodes <https://github.com/genshinsim/gcsim/pull/1148/commits/6bc1ba3ae34f8ed01d8ed32e2d49cc3ee57deb83>
- add system function for action execution <https://github.com/genshinsim/gcsim/pull/1148/commits/e82534ba8a7a129dfc470b80e50acc75b87eef75>
- remove unused ActionStmt (replaced by system function) <https://github.com/genshinsim/gcsim/pull/1148/commits/cbb88c45722a75f990b3861819ae1fc0a2ac6ce1>
- refactor `simulator` package to remove direct reference to `gcs.Eval`. Instead, change `simulator` to require an `Evaluator` interface, which is implemented by `gcs.Eval` <https://github.com/genshinsim/gcsim/pull/1148/commits/3dff8d6ba26076dc3f35371950d595c79d8edb89>, <https://github.com/genshinsim/gcsim/pull/1148/commits/43aafc1d8e76b8b648ae8999afbf8b3525ad925f>
- fixed repl pkg error handling (broken in previous changes): <https://github.com/genshinsim/gcsim/pull/1148/commits/21b63fbad42dba18bbc93893105fd7f8164091c6>
- adjust `BlockStmt` to return last line evaluated if no explicit return (cleaner for testing): <https://github.com/genshinsim/gcsim/pull/1148/commits/b4efdb7acf4d77d44c2f196ee7eb2846814dff18>
- refactor various info holding struct for character, weapon, sets, enemy and ActionList into `core/info` pkg, with the intention of consolidating all these info holding structs <https://github.com/genshinsim/gcsim/pull/1148/commits/1254cdeb4c2d283783c318e6b54bd4f5fdeff191>, <https://github.com/genshinsim/gcsim/pull/1148/commits/48e23c8fb2cff04c073c0b7d810de4b7a353ea1e>, <https://github.com/genshinsim/gcsim/pull/1148/commits/212928515aa6b0284ab7d4233f50f783aa711821>
- change the `simulator` core loop into a stateFn (similar to srsim implementation) to facilitate easier jumping/skipping of states (i.e. skipping directly from queue to next frame if action is wait). Added effect of resolving #1301 <https://github.com/genshinsim/gcsim/pull/1148/commits/c3da54d5861994a9e7f1bcfc0afed9af4326214c>
- **breaking change**: removed old `roll` stats syntax that was experimental <https://github.com/genshinsim/gcsim/pull/1148/commits/89e712f8bf25a8304803d845c9f9b38be8c514b4>
- add alias `sleep` for `wait` to gcsl in preparation for eventually deprecating `wait` in favour of `sleep` (`sleep` is a more accurate word to describe what is actually happening) <https://github.com/genshinsim/gcsim/pull/1148/commits/50aef12c5f1e4ae669e9c5e98e41287b5f1e82e5>
- add `delay` system func, which adds a delay to the next action for a specified number of frames only when next action is ready to be used <https://github.com/genshinsim/gcsim/pull/1148/commits/593831bed8bc1f8cedff30a8f9d6cc78519ce939>
- added correct handling of `wait(0)` (previously not allowed and would cause crash) to correctly do nothing <https://github.com/genshinsim/gcsim/pull/1148/commits/eccd46265faafb995bbd36527cd3229ee75695ca>
- refactored how functions are handled in AST; added special node `ExprType` to AST specifically for the purpose of storing variable and function typing information. Typing function is where this is required because you can have a function that takes another function etc... which makes the typing definition recursive in nature. <https://github.com/genshinsim/gcsim/pull/1148/commits/79fab82373f16e0573b65d282081b689fda242d6>
- fixed function env in Eval to allow proper closure when returning functions (same commit as above) <https://github.com/genshinsim/gcsim/pull/1148/commits/79fab82373f16e0573b65d282081b689fda242d6> and <https://github.com/genshinsim/gcsim/pull/1148/commits/c8eded4f619e4c3abf79a51af0df567b15be1e3f>

## Remaining todo (tracked in #1650)
- [ ] add `null` type
- [ ] add parsing unit tests
- [ ] add comparison support for diff variable types (currently only support comparing numbers)
```
let a string = "blah";
let b string = "boo";
return a == b; //should return 0
```
- [ ] add `character` type; should support various methods such as `.cd`, `.energy`, etc
- [ ] add `action` type; should also support method (tbd)
- [ ] add custom functions re pre-delay
- [ ] parse existing fields into system calls, utilizing `character` and `action` type where appropriate
- [ ] separate warnings and errors in parsed results; warnings are more linting related; errors = parse/compile errors
- [ ] add support for ast validators to be loaded at parse time
- [ ] add default validator that does type checking
```
let a = "blah"
return a == 1; //this should fail parse

fn x(a string) string {
   return a + " is cool";
}
print(fn(1)); //this should fail as 1 is not a string
```
- [ ] refactor sysfunc into regular functions but injected into env at init; error handled user unintended use of system function name (identifier already exists error) 

## Specific to making wait useable in character actions

To make it delay to write `xingqiu attack, wait(10), attack, delay(5), attack;`, the parsing needs to be changed such as the entire line is treated as a statement (vs currently it is carved out the same way character stats parsing is carved out) and each action can be parsed as part of the recursive parser

- [ ] rewrite action parsing into parseStmt; action should be parsed recursively same as all other AST
- [ ] allow for `wait` in character actions (shelved for now because this is very difficult to implement)

## Optional

- [ ] add deprecation warning for `wait`; suggest user to use `sleep`
- [ ] add read-only variables (add `const` keyword?).
```
const val = 5;
val = 4; # error
```
- [ ] rewrite sys_func into a switch statement for better performance instead of using map

## Done
- [x] add support of different types of variables for let/assign statements.
```
let a = "test";
a = "string";
```
- [x] add `type` function to determine the type of variable.
```
print(type("asd")); # string
print(type(123)); # number
```
- [x] move functions from `fnMap` to `varMap`.
```
fn test() {}
print(test) # func
print(type(test)) # function
```
- [x] add anonymous functions.
```
let test = fn() { return 0; };
print(test())
```
- [x] add support a `test()()` record.
```
fn test() {
  return fn() { return 0; }
}
test()();
```
- [x] move standart functions (print/f/set_player_pos/etc...) to env (add `typBuintin` for built-in functions?).
```
print(print); # func
```
- [x] add `true` (1)/`false` (0) constants
- [x] clear some comments
- [x] clean up `roll` syntax
- [x] add `sleep` as alias for `wait`
- [x] add `delay` command
- [x] add parsing for explicit function/variable type declaration (undeclared defaults to `number` to keep backward compat)
- [x] fix functions with no return causing error (breaking change)

## Shelved for now

- [ ] if the variable/field doesn't exist, return `null` instead of an error.
```
print(unkval); # null
```
- [ ] add new syntax for indexing (just delete the dot at the beginning).
```
.obj.field1 # old syntax
obj.field1 # new syntax
```
- [ ] add the `object` type.
```
let obj = {
  field1 = 1,
  field2 = "test"
};
```


srliao:
@imring forgot where I had left off with this but I'll take a look again some late next week

unleashurgeek (approved):

PR #1156, opened 2022-12-23
fix damage and elemental ICD index wrapping

### LINKED ISSUE #1153: icd arrays should not wrap around
Both the elemental and damage sequence shouldn't be considered to wrap around but instead considered 0 if trying to access an entry outside of it.

skippi:
Fixes #1153 .

### DIFF
--- a/pkg/target/icd.go
+++ b/pkg/target/icd.go
@@ -21,10 +21,8 @@ func (t *Target) WillApplyEle(tag combat.ICDTag, grp combat.ICDGroup, char int) 
 
 	val := t.icdTagCounter[char][tag]
 	//increment the counter
-	//if counter > length, then reset back to 0
-	t.icdTagCounter[char][tag]++
-	if t.icdTagCounter[char][tag] == len(combat.ICDGroupEleApplicationSequence[grp]) {
-		t.icdTagCounter[char][tag] = 0
+	if t.icdTagCounter[char][tag] < len(combat.ICDGroupEleApplicationSequence[grp]) - 1 {
+		t.icdTagCounter[char][tag]++
 	}
 
 	t.Core.Log.NewEvent("ele icd check", glog.LogICDEvent, char).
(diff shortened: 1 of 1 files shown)


pkg/target/icd.go
```diff
@@ -21,10 +21,8 @@ func (t *Target) WillApplyEle(tag combat.ICDTag, grp combat.ICDGroup, char int)
 
 	val := t.icdTagCounter[char][tag]
 	//increment the counter
-	//if counter > length, then reset back to 0
-	t.icdTagCounter[char][tag]++
-	if t.icdTagCounter[char][tag] == len(combat.ICDGroupEleApplicationSequence[grp]) {
-		t.icdTagCounter[char][tag] = 0
+	if t.icdTagCounter[char][tag] < len(combat.ICDGroupEleApplicationSequence[grp]) - 1 {
+		t.icdTagCounter[char][tag]++
 	}
```
unleashurgeek, line 26: If I understand this issue correctly, we want `combat.ICDGroupEleApplicationSequence[grp][val]` to be `0` when `t.icdTagCounter[char][tag] >= len(combat.ICDGroupEleApplicationSequence[grp])`. Current logic will have last value in the sequence be used, which in some cases is not `0`.

pkg/target/icd.go
```diff
@@ -6,61 +6,60 @@ import (
...
 	val := t.icdTagCounter[char][tag]
-	//increment the counter
-	//if counter > length, then reset back to 0
 	t.icdTagCounter[char][tag]++
-	if t.icdTagCounter[char][tag] == len(combat.ICDGroupEleApplicationSequence[grp]) {
-		t.icdTagCounter[char][tag] = 0
+
+	// if counter > length, then use 0 for group seq
+	groupSeq := 0.0
+	if val < len(combat.ICDGroupEleApplicationSequence) {
```
k0l11, line 26: Shouldn't this be` if val < len(combat.ICDGroupEleApplicationSequence[grp])? (missing [grp])

k0l11 (approved):

PR #1159, opened 2022-12-24
add traveleranemo a4

### LINKED ISSUE #1103: Anemo MC A4 is not implemented
[Palm Vortex](https://genshin-impact.fandom.com/wiki/Palm_Vortex) kills regenerate 2% HP for 5s. This effect can only occur once every 5s

(should only proc on enemy kills and not on any other kills)

skippi:
Fixes #1103 .

EDIT: sample config https://gcsim.app/v3/viewer/share/9213b206-08b0-4efb-a4d9-ec6caae63b15.

### DIFF
--- a/internal/characters/traveleranemo/asc.go
+++ b/internal/characters/traveleranemo/asc.go
@@ -28,7 +28,7 @@ func (c *char) a4() {
 		c.AddStatus(a4ICDKey, 300, true)
 
 		for i := 0; i < 5; i = i + 1 {
-			c.Core.Tasks.Add(func() {
+			c.QueueCharTask(func() {
 				c.Core.Player.Heal(player.HealInfo{
 					Caller:  c.Index,
 					Target:  c.Index,
--- a/internal/characters/traveleranemo/asc.go
+++ b/internal/characters/traveleranemo/asc.go
@@ -1,0 +1,44 @@
+package traveleranemo
+
+import (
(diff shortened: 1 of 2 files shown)


unleashurgeek (approved):

internal/characters/traveleranemo/asc.go
```diff
@@ -0,0 +1,44 @@
...
+			return false
+		}
+		if c.StatusIsActive(a4ICDKey) {
+			return false
+		}
+
+		c.AddStatus(a4ICDKey, 300, true)
+
+		for i := 0; i < 5; i = i + 1 {
+			c.Core.Tasks.Add(func() {
```
k0l11, line 31: should be affected by hitlag

internal/characters/traveleranemo/asc.go
```diff
@@ -0,0 +1,44 @@
...
+			return false
+		}
+		if c.StatusIsActive(a4ICDKey) {
+			return false
+		}
+
+		c.AddStatus(a4ICDKey, 300, true)
+
+		for i := 0; i < 5; i = i + 1 {
+			c.QueueCharTask(func() {
```
k0l11, line 31: This isn't enough, because now you have hitlag affecting later ticks even though they shouldn't even be queued yet.

k0l11 (approved):

Issue #892, opened 2022-09-01
dash cooldown is not implemented 


k0l11:
"Sprinting enters a short cooldown if two dashes are used in quick succession. However, this does not apply to Alternative Sprints." (Also doesn't apply to Wanderer Dash inside E.)

exact mechanics:
- length of the cooldown -> 1.5s
- cooldown shared or per character? -> it's per character
- does the cooldown tick down when char with cooldown is off-field? -> no
- exact condition for the dash cooldown to be triggered -> second dash starts within 0.8s after first dash starts (insert link to KQM ticket later)
- affected by hitlag? -> yes, see https://youtu.be/sswkQCaTqzY?t=16 (dash 1 starts at ~15.2s, dash 2 starts at ~16.2s, hitlag inbetween dash 1 and dash 2 was 3+3=6f in total, which is 6/30 = 0.2s at 30fps, 1s - 0.2s = 0.8s)
- affected by ping? -> no 
- affected by cooldown reduction? -> most likely not
- affected by movement speed buffs? -> outside of decreasing the time between dashes, most likely not

Charlie-Zheng:
Link to KQM ticket:
https://library.keqingmains.com/evidence/general-mechanics/movement-and-physics#consecutive-dash-cooldown-trigger-is-08s

PR #1160, opened 2022-12-24
add dash cooldown mechanic

skippi:
Implements #892.

- [ ] optimize off-field tick implementation.
- [x] decide whether to use statuses or cooldown queue to implement. (decision is to use statuses)


k0l11:
> decide whether to use statuses or cooldown queue to implement.
Since it's most likely not affected by cooldown reduction, it should probably be a status.

k0l11:
FYI: Wanderer's Dash also has to be adjusted to account for this change, assuming he gets merged before this.

Issue #1151, opened 2022-12-20
XQ ult trigger is too early in NA animation


Charlie-Zheng:
Xingqiu Q rainswords trigger some frames into an NA, and not immediately at the start. This is unlike Yelan who appears to trigger immediately. On some characters this can be so late it happens after the hitbox.

https://user-images.githubusercontent.com/16330291/208778965-838c87a0-8931-4ee2-b10a-2705df90b2d6.mp4

Issue #989, opened 2022-10-04
This config has a moment where Yelan burst triggers but Xingqiu burst does not, exactly 60 frames after the previous trigger


soloxcx:
https://gcsim.app/v3/viewer/share/64352098-8a58-42b4-ae17-5047fb4901e1

so in these configs, yelan and xq bursts are both active. there is 1 moment where yelan's burst triggers but xingqiu's does not, and this moment happens to be exactly 60 frames after the previous trigger, maybe unrelated.

at frame 596 hutao triggers yelan and xingqiu bursts, then at frame 656 she tries to do it again. 

at frame 656 only yelan burst fires again while xingqiu's does not.

![image](https://user-images.githubusercontent.com/58927267/194109204-19d4ee3d-e1a4-4bb5-818f-f5470c556491.png)

if I add a wait(1) just before frame 656 then xingqiu burst successfully triggers again: https://gcsim.app/v3/viewer/share/39eeeb4d-5a0d-41ab-8312-c233c852a40d

soloxcx:
actually, this issue is not what I thought originally. hu tao triggers yelan and xingqiu bursts in 60 frame intervals for her entire uptime, it's just that specific timing at 656f that is still "not in a normal state" unlike the others. why, i'm not sure.

soloxcx:
cant replicate again, as far as i can tell

PR #1167, opened 2022-12-30
Added N0 timing implementation framework

Charlie-Zheng:
Frame timings collected and implemented for:
Everyone except for Nilou in E

https://docs.google.com/spreadsheets/d/1DkFeG8fEkmM84XW2jtYuNpNxyE1W7-XXMj0V6Bn6wrA/edit?usp=sharing


k0l11:
LGTM after a short look over it aside from a few comments. @unleashurgeek you might also want to check this out.

internal/characters/xingqiu/burst.go
```diff
@@ -154,6 +182,40 @@ func (c *char) summonSwordWave() {
...
+	} else {
+		return XQ_N0_delays[activeCharKey]
+	}
+}
+
+func (c *char) burstStateDelayFunc() {
+	//ignore if on ICD
+	if c.StatusIsActive(burstICDKey) {
+		return
+	}
```
unleashurgeek, line 204: should add a check for if the character is the same as when it was scheduled. ideally this check should instead be if the current NA is the same as the one that was scheduled, but I can't think of a good way to do that today. Should probably just add a TODO for this and create an issue

internal/characters/xingqiu/burst.go
```diff
@@ -25,6 +29,30 @@ func init() {
...
+	XQ_N0_delays_alt_forms[keys.Raiden] = 13
+	alt_form_status_keys[keys.Raiden] = "raidenburst"
+
+	XQ_N0_delays[keys.Bennett] = 7
+	XQ_N0_delays[keys.Diluc] = 15
+	XQ_N0_delays[keys.Kazuha] = 10
+	XQ_N0_delays[keys.Keqing] = 8
+	XQ_N0_delays[keys.Xiangling] = 7
+
+	// jumping/dashing during the NA windup for some catalysts modifies their frames - said by koli
```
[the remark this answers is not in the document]
Charlie-Zheng: Hmm but holding the button for a certain duration definitely affects it. Kokomi spawns her N1 in like 2f, but you can hold for 4 frames then release and still do a N1.

internal/characters/xingqiu/burst.go
```diff
@@ -154,6 +182,40 @@ func (c *char) summonSwordWave() {
...
+
+func (c *char) burstStateDelayFunc() {
+	//ignore if on ICD
+	if c.StatusIsActive(burstICDKey) {
+		return
+	}
+
+	state := c.Core.Player.CurrentState()
+	//this should start a new ticker if not on ICD and state is correct
+	if state == action.NormalAttackState {
```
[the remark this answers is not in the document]
Charlie-Zheng: I've used @unleashurgeek's suggestion of combining the conditions

internal/characters/xingqiu/burst.go
Charlie-Zheng: I added a source frame check, hopefully that ensure it works properly?

internal/characters/xingqiu/burst.go
```diff
@@ -25,6 +29,30 @@ func init() {
...
+		XQ_N0_delays_alt_forms[i] = -1
+	}
+
+	XQ_N0_delays[keys.Nahida] = 9
+	XQ_N0_delays[keys.Xingqiu] = 7
+	XQ_N0_delays[keys.Yelan] = 9
+
+	XQ_N0_delays[keys.Raiden] = 13
+	XQ_N0_delays_alt_forms[keys.Raiden] = 13
+	alt_form_status_keys[keys.Raiden] = "raidenburst"
```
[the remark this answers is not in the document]
Charlie-Zheng: I've made the burst key and melee key for Raiden and Tartaglia, respectively, public.

internal/characters/xingqiu/burst.go
```diff
@@ -25,6 +31,80 @@ func init() {
...
+		XQ_N0_delays_alt_forms[i] = -1
+	}
+
+	XQ_N0_delays[keys.Nahida] = 9
+	XQ_N0_delays[keys.Xingqiu] = 7
+	XQ_N0_delays[keys.Yelan] = 9
+
+	XQ_N0_delays[keys.Raiden] = 13
+	XQ_N0_delays_alt_forms[keys.Raiden] = 13
+	alt_form_status_keys[keys.Raiden] = raiden.BurstKey
```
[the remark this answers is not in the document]
Charlie-Zheng: It's not, I can remove it. Previous this was the only alt form I had counted so I had it included to show the difference.

internal/data/NAstateframes.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"github.com/genshinsim/gcsim/internal/characters/ayato"
+	"github.com/genshinsim/gcsim/internal/characters/cyno"
+	"github.com/genshinsim/gcsim/internal/characters/tartaglia"
+	"github.com/genshinsim/gcsim/pkg/core/keys"
+)
+
+var PercentDelay5 = make([]int, keys.EndCharKeys)
+var PercentDelay5AltForms = make([]int, keys.EndCharKeys)
+var AltFormStatusKeys = make([]string, keys.EndCharKeys)
```
unleashurgeek, line 12: NIT: improve naming

also change file name to all lowercase name

internal/characters/thoma/burst.go
```diff
@@ -78,56 +79,100 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
...
+			Write("state", action.NormalAttackState).
+			Write("icd", c.StatusExpiry(burstICDKey))
+		c.burstTickSrc = c.Core.F
+		//use the hitlag affected queue for this
+		c.QueueCharTask(c.burstTickerFunc(c.Core.F), 60) //check every 1sec
+	}
+}
+
+func (c *char) burstStateHook() {
 	c.Core.Events.Subscribe(event.OnStateChange, func(args ...interface{}) bool {
```
k0l11, line 115: This hook can be changed to OnAttack. Then you can skip the animation state check in the subscription here. Same applies to Xingqiu.

OnAttack gets emitted here:
https://github.com/genshinsim/gcsim/blob/97e8e28c0e21004127ac33dddc62faa91331fff7/pkg/core/player/exec.go#L157
OnStateChange gets emitted here:
https://github.com/genshinsim/gcsim/blob/c184fb9382670f65c852e6a568bd51819fe0fe65/pkg/core/player/animation/animation.go#L87

As you can see, OnAttack gets emitted right before SetActionUsed gets called, which is where OnStateChange is emitted. You don't need info about the previous state/action, so OnAttack should also work and reduce the amount times this function gets called.

internal/common/normalattackstate.go
```diff
@@ -134,3 +140,146 @@ func Get5PercentN0Delay(activeChar *character.CharWrapper) int {
...
+		abilICDKey:  abilICDKey,
+		abilTickSrc: 0,
+		abilHookSrc: 0,
+		delayFunc:   delayFunc,
+		summonFunc:  summonFunc,
+	}
+
+}
+
+func (w *NAHook) NAStateHook() {
```
unleashurgeek, line 178: NIT: I'd name this something like `Attach` or `Enable` to make it clear that it is performing a subscription

internal/common/normalattackstate.go
```diff
@@ -134,3 +140,146 @@ func Get5PercentN0Delay(activeChar *character.CharWrapper) int {
...
+		w.abilHookSrc = w.Core.F
+		delay := w.delayFunc(w.Core.Player.ActiveChar())
+		w.Core.Log.NewEvent(fmt.Sprintf("%v delay on state change", w.abilName), glog.LogCharacterEvent, w.c.Index).
+			Write("delay", delay)
+		// This accounts for the delay in n0 timing needed to trigger
+
+		if delay > 0 {
+
+			w.Core.Tasks.Add(w.naStateDelayFuncGen(w.Core.F), delay)
+		} else {
```
unleashurgeek, line 193: return inside this `if` and remove `else` to reduce code depth

NIT: remove blank line

internal/common/normalattackstate.go
```diff
@@ -134,3 +140,146 @@ func Get5PercentN0Delay(activeChar *character.CharWrapper) int {
...
+			} else {
+				w.summonFunc()
+				w.c.AddStatus(w.abilICDKey, 60, true)
+				w.Core.Log.NewEvent(fmt.Sprintf("%v on state change", w.abilName), glog.LogCharacterEvent, w.c.Index).
+					Write("state", action.NormalAttackState).
+					Write("icd", w.c.StatusExpiry(w.abilICDKey))
+				w.abilTickSrc = w.Core.F
+				//use the hitlag affected queue for this
+				w.c.QueueCharTask(w.naTickerFunc(w.Core.F), w.abilProcICD) //check every `ICD`` frames
+			}
```
unleashurgeek, line 210: IMO I think it'd be cleaner if we flip this condition and then have a single log line. Something like:

```go
isActive := w.c.StatusIsActive(w.abilICDKey)
if !isActive {
  w.summonFunc()
  w.c.AddStatus(w.abilICDKey, 60, true)
  w.c.QueueCharTask(w.naTickerFunc(w.Core.F), w.abilProcICD)
  w.abilTickSrc = w.Core.F
}

w.Core.Log.NewEvent(
    fmt.Sprintf("%v on state change [triggered=%v]", w.abilName, !isActive),
    glog.LogCharacterEvent,
    w.c.Index)
  .Write("state", w.Core.Player.CurrentState())
  .Write("icd", w.c.StatusExpiry(w.abilICDKey))
```

internal/common/normalattackstate.go
```diff
@@ -134,3 +140,146 @@ func Get5PercentN0Delay(activeChar *character.CharWrapper) int {
...
+		w.summonFunc()
+		w.c.AddStatus(w.abilICDKey, 60, true)
+		w.Core.Log.NewEvent(fmt.Sprintf("%v on state change", w.abilName), glog.LogCharacterEvent, w.c.Index).
+			Write("state", action.NormalAttackState).
+			Write("icd", w.c.StatusExpiry(w.abilICDKey))
+		w.abilTickSrc = w.Core.F
+		//use the hitlag affected queue for this
+		w.c.QueueCharTask(w.naTickerFunc(w.Core.F), w.abilProcICD) //check every `ICD` frames
+	}
+}
```
unleashurgeek, line 240: This logic seems very similar/nearly identical to the `delay = 0` case in the subscription? can we reuse it in some way (have the different delayed check -> function that does status check + all the other necessary calls?)

internal/common/normalattackstate.go
```diff
@@ -134,3 +140,146 @@ func Get5PercentN0Delay(activeChar *character.CharWrapper) int {
...
+			Write("state", state).
+			Write("icd", w.c.StatusExpiry(w.abilICDKey))
+
+		//we can trigger here b/c we're in normal state still and src is still the same
+		w.summonFunc()
+		w.c.AddStatus(w.abilICDKey, 60, true)
+		//in theory this should not hit an icd?
+		//use the hitlag affected queue for this
+		w.abilTickSrc = w.Core.F
+		w.c.QueueCharTask(w.naTickerFunc(w.Core.F), w.abilProcICD) //check every 1sec
```
unleashurgeek, line 283: Same comment with the other two places. Seems like this log + 4 operations can be reused

internal/characters/xingqiu/xingqiu.go
```diff
@@ -38,7 +37,15 @@ func NewChar(s *core.Core, w *character.CharWrapper, _ profile.CharacterProfile)
...
+	common.NewNAHook(
+		c.CharWrapper,
+		c.Core,
+		"xingqiu burst",
+		burstKey,
+		60,
+		burstICDKey,
+		common.Get5PercentN0Delay,
+		c.summonSwordWave).
+		NAStateHook()
```
unleashurgeek, line 49: super nit: does `gofmt` do this? IMO `.` should always start the next line when you are splitting between multiple (but also go syntax weird so whatever lol)

internal/characters/thoma/thoma.go
```diff
@@ -42,6 +42,17 @@ func (c *char) Init() error {
 		c.c6buff = make([]float64, attributes.EndStatType)
 		c.c6buff[attributes.DmgP] = .15
 	}
+
+	common.NewNAHook(
```
k0l11, line 46: Please consider passing a `common.NAHook` struct directly to make it easier to read/harder to mix up the params.

k0l11:
closes #1151 
need to check whether #989 is still a problem

internal/common/normalattackstate.go
```diff
@@ -147,139 +147,128 @@ func Get0PercentN0Delay(activeChar *character.CharWrapper) int {
...
+	icd := w.C.StatusIsActive(w.AbilICDKey)
+	if icd {
+		w.Core.Log.NewEvent(fmt.Sprintf("%v not triggered, on proc ICD", w.AbilName), glog.LogCharacterEvent, w.C.Index).
+			Write("icd", w.C.StatusExpiry(w.AbilICDKey))
+	}
+	return !icd
+}
+
+func (w *NAHook) checkState() bool {
+	state := w.Core.Player.CurrentState()
```
unleashurgeek, line 249: NIT: use `state` variable in logs or remove

unleashurgeek (approved):

PR #1203, opened 2023-01-03
fix ICD not using last sequence value for default

skippi:
If the ICD counter exceeds the length of the array, the last value should be used.


Charlie-Zheng (commented):
Using 
```
group = combat.ICDGroupEleApplicationSequence[grp]
groupSeq := group[len(group)-1]
```
Might be better?

k0l11 (approved):

unleashurgeek (approved):

Charlie-Zheng (approved):
Looks good for hot fix

Issue #1204, opened 2023-01-03
Ganyu C1 should only proc once per arrow


k0l11:
At the moment it procs twice (once on Frostflake Arrow and once on Frostflake Arrow Bloom).
Relevant: https://github.com/genshinsim/gcsim/blob/1afe4568d285361d3b61543cc9384dc8b1b88770/internal/characters/ganyu/cons.go#L19

PR #1213, opened 2023-01-07
Made Ganyu C1 a call back and fixed doubled energy gain

RarePossum:
Made Ganyu C1 a callback.

In addition, the bug where her C1 restored energy on every CA hit is fixed. This currently uses a status with a 25f delay to prevent energy gain. This should work for all practical scenarios outside of travel time shenanigans.


k0l11:
closes #1204

PR #1214, opened 2023-01-07
switched ganyu c1 to a callback and fixed doubled energy gain

RarePossum:
Address #1204 

Makes Ganyu C1 Call back instead of an on event subscribe. 

Sets a 24f icd to Ganyu's energy gain from C1 to prevent it from occuring on both hits of a CA.

### DIFF
--- a/internal/characters/ganyu/aimed.go
+++ b/internal/characters/ganyu/aimed.go
@@ -80,6 +80,7 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 				1,
 			),
 			travel,
+			c.c1(),
 		)
 
 		ai.Abil = "Frost Flake Bloom"
@@ -90,6 +91,7 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 			snap,
 			combat.NewCircleHitOnTarget(c.Core.Combat.PrimaryTarget(), nil, 5),
 			travel+bloom,
+			c.c1(),
 		)
 
 		// first shot/bloom do not benefit from a1
--- a/internal/characters/ganyu/aimed.go
+++ b/internal/characters/ganyu/aimed.go
@@ -17,7 +17,6 @@ func init() {
 	aimedFrames = frames.InitAbilSlice(113)
 	aimedFrames[action.ActionDash] = aimedHitmark
 	aimedFrames[action.ActionJump] = aimedHitmark
-	
 }
 
 func (c *char) Aimed(p map[string]int) action.ActionInfo {
@@ -46,11 +45,7 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 		HitlagOnHeadshotOnly: true,
 		IsDeployable:         true,
 	}
-	var c1cb combat.AttackCBFunc
-	if c.Base.Cons >= 1 {
-		c.c1done = false
-		c1cb = c.c1()
-	}
+	c1cb := c.c1()
 	// TODO: not sure if this works as intended
 	skip := 0
 	if c.Core.Status.Duration(c6Key) > 0 {
--- a/internal/characters/ganyu/aimed.go
+++ b/internal/characters/ganyu/aimed.go
@@ -17,6 +17,7 @@ func init() {
 	aimedFrames = frames.InitAbilSlice(113)
 	aimedFrames[action.ActionDash] = aimedHitmark
 	aimedFrames[action.ActionJump] = aimedHitmark
+	
 }
 
 func (c *char) Aimed(p map[string]int) action.ActionInfo {
@@ -45,7 +46,11 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 		HitlagOnHeadshotOnly: true,
 		IsDeployable:         true,
 	}
-
+	var c1cb combat.AttackCBFunc
+	if c.Base.Cons >= 1 {
+		c.c1done = false
+		c1cb = c.c1()
+	}
 	// TODO: not sure if this works as intended
 	skip := 0
 	if c.Core.Status.Duration(c6Key) > 0 {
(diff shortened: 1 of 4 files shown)


internal/characters/ganyu/cons.go
```diff
@@ -13,31 +13,30 @@ const (
...
 
-func (c *char) c1() {
-	c.Core.Events.Subscribe(event.OnEnemyDamage, func(args ...interface{}) bool {
-		atk := args[1].(*combat.AttackEvent)
-		e, ok := args[0].(*enemy.Enemy)
-		if !ok {
-			return false
+//Ganyu C1: Taking DMG from a Charge Level 2 Frostflake Arrow or Frostflake Arrow Bloom decreases opponents' Cryo RES by 15% for 6s.
+//A hit regenerates 2 Energy for Ganyu. This effect can only occur once per Charge Level 2 Frostflake Arrow, regardless if Frostflake Arrow itself or its Bloom hit the target.
+func (c *char) c1() combat.AttackCBFunc {
```
k0l11, line 21: There's no need to introduce a custom icd for this if you use the same callback instance for both arrows. So in aimed.go just define something like `c1cb := c.c1()` there.
Also missing constellation check:
```
if c.Base.Cons < 1 {
    return nil 
}
done := false 
return func(a combat.AttackCB) {
    e:= a.Target.(*enemy.Enemy)
    if e.Type() != combat.TargettableEnemy {
        return
    }
    if done {
        return
    }
    done = true
    ...
}
```

internal/characters/ganyu/cons.go
```diff
@@ -13,31 +13,30 @@ const (
...
-			return false
-		}
-		if !(atk.Info.Abil == "Frost Flake Arrow" || atk.Info.Abil == "Frost Flake Bloom") {
-			return false
-		}
-
-		c.AddEnergy(c1Key, 2)
 		e.AddResistMod(enemy.ResistMod{
-			Base:  modifier.NewBase(c1Key, 5*60),
+			Base:  modifier.NewBaseWithHitlag("ganyu-c1", 300),
```
k0l11, line 28: should use c1Key here if you're already defining a variable for it

internal/characters/ganyu/cons.go
```diff
@@ -13,31 +13,30 @@ const (
...
 		})
-
-		return false
-	}, c1Key)
+		//Uses ICD to simulate per arrow. 25f has it be restored on the same frame that the bloom hits. There should be no practical way to circumvent this
+		if !c.StatusIsActive(c1ICD) {
+			c.AddEnergy(c1Key, 2)
+			c.AddStatus(c1ICD, 24, false)
+		}
+		c.Core.Log.NewEvent("Rosaria A1 activation", glog.LogCharacterEvent, c.Index).
```
k0l11, line 37: Please remove this log entirely.

k0l11:
closes #1142

internal/characters/ganyu/aimed.go
```diff
@@ -17,6 +17,7 @@ func init() {
 	aimedFrames = frames.InitAbilSlice(113)
 	aimedFrames[action.ActionDash] = aimedHitmark
 	aimedFrames[action.ActionJump] = aimedHitmark
+
```
k0l11, line 20: nit: pls remove the newline

internal/characters/ganyu/cons.go
```diff
@@ -13,31 +12,32 @@ const (
 	c1Key = "ganyu-c1"
 	c4Key = "ganyu-c4"
 	c6Key = "ganyu-c6"
+	c1ICD = "ganyu-c1-energy-icd"
```
k0l11, line 15: pls remove this icd key since it's no longer necessary

internal/characters/ganyu/aimed.go
```diff
@@ -45,7 +46,11 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 		HitlagOnHeadshotOnly: true,
 		IsDeployable:         true,
 	}
-
+	var c1cb combat.AttackCBFunc
+	if c.Base.Cons >= 1 {
+		c.c1done = false
```
k0l11, line 51: This doesn't need to be a char variable, please stick to the way I described it in the very first comment.

internal/characters/ganyu/aimed.go
```diff
@@ -45,7 +46,11 @@ func (c *char) Aimed(p map[string]int) action.ActionInfo {
 		HitlagOnHeadshotOnly: true,
 		IsDeployable:         true,
 	}
-
+	var c1cb combat.AttackCBFunc
```
k0l11, line 49: This can be condensed to: 
```
c1cb := c.c1()
```
c.c1() returns nil if Ganyu isn't c1+.

internal/characters/ganyu/cons.go
```diff
@@ -13,31 +12,32 @@ const (
...
 
-func (c *char) c1() {
-	c.Core.Events.Subscribe(event.OnEnemyDamage, func(args ...interface{}) bool {
-		atk := args[1].(*combat.AttackEvent)
-		e, ok := args[0].(*enemy.Enemy)
-		if !ok {
-			return false
+//Ganyu C1: Taking DMG from a Charge Level 2 Frostflake Arrow or Frostflake Arrow Bloom decreases opponents' Cryo RES by 15% for 6s.
+//A hit regenerates 2 Energy for Ganyu. This effect can only occur once per Charge Level 2 Frostflake Arrow, regardless if Frostflake Arrow itself or its Bloom hit the target.
+func (c *char) c1() combat.AttackCBFunc {
```
k0l11, line 20: Please stick to the way I described it in the very first comment. That would simplify the logic here.

k0l11 (approved):

Issue #1217, opened 2023-01-10
At C2+, Hu Tao should apply blood blossoms with her burst while she's not in paramita


ScarletSeal:
"active hutao;
hutao burst;
wait(1000)" does not apply her mark ![Immagine](https://user-images.githubusercontent.com/118117643/211552499-70f3f66b-2293-439c-a65d-119e08e2481c.png)

"active hutao;
hutao skill, burst;
wait(1000)" applies the blossom as intended ![Immagine](https://user-images.githubusercontent.com/118117643/211552652-f2034833-498f-448d-ae2e-a94447224707.png)

Proof that she does in game: https://www.youtube.com/watch?v=uYOCgwXFpig

k0l11:
Thanks for creating this issue.
This is a rather simple fix, so feel free to submit a PR, where you go here:
https://github.com/genshinsim/gcsim/blob/489b5758af968026ab1772386d5096d4ee76ab13/internal/characters/hutao/burst.go#L49
and change it to:
```
if c.Base.Cons >= 2 {
    ...
}
```

PR #1218, opened 2023-01-10
C2 Hutao burst fix

ScarletSeal:
now the only condition for her to apply Blood Blossoms with her burst is "if c.Base.Cons >= 2"

"c.StatModIsActive(paramitaBuff)" condition removed


k0l11:
closes #1217

k0l11 (approved):

Issue #968, opened 2022-10-02
hu tao pp slide is not implemented


k0l11:
Hu Tao's E state (and most likely all the buffs she has at that time) cannot expire during her CA animation, so cancelling her CA into Q should make the Q always have the buffs from the start of her CA.

k0l11:
closed by #1290

PR #1227, opened 2023-01-16
Added Hutao's PPSlide mechanic.

Charlie-Zheng:
A status controls whether the buff stays up or not, but the buff is only removed on state changes or on swaps.

Addresses Issue #968

### DIFF
--- a/internal/characters/hutao/asc.go
+++ b/internal/characters/hutao/asc.go
@@ -11,11 +11,6 @@ const (
 )
 
 func (c *char) a1() {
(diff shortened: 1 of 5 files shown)


k0l11:
PP Slide for Hu Tao in particular should work like this:
- at the start of CA: add 1.6s to e status if still up
- at the end of CA: remove 1.6s from e status (can use OnRemoved in the return ActionInfo

unleashurgeek:
Added in #1290

PR #1235, opened 2023-01-18
alhaitham implementation

IvanHdzF:
TODO:
 Add frames
 Check hitlag data
 Check mirror timings
 Check hitlag extension
 Check ICDs
 Check snapshot timings
 Cleanup code (this also involves deleting the TODOs )
 Make a milkshake


internal/characters/alhaitham/alhaitham.go
```diff
@@ -0,0 +1,88 @@
...
+func (c *char) Init() error {
+	c.onExitField()
+	c.a4()
+
+	return nil
+}
+func (c *char) onExitField() {
+	c.Core.Log.NewEvent("sub yes", glog.LogCharacterEvent, c.Index)
+	c.Core.Events.Subscribe(event.OnCharacterSwap, func(args ...interface{}) bool {
+		// do nothing if previous char wasn't ayato
```
imring, line 54: alhaitham*

internal/characters/alhaitham/asc.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	m := make([]float64, attributes.EndStatType)
+	c.AddAttackMod(character.AttackMod{
+		Base: modifier.NewBase("alhaitham-a4", -1),
+		Amount: func(atk *combat.AttackEvent, _ combat.Target) ([]float64, bool) {
+			// only trigger on projection attack and burst damage
+			if atk.Info.Abil != "Chisel-Light Mirror: Projection Attack 1" &&
+				atk.Info.Abil != "Chisel-Light Mirror: Projection Attack 2" &&
+				atk.Info.Abil != "Chisel-Light Mirror: Projection Attack 3" &&
+				atk.Info.AttackTag != combat.AttackTagElementalBurst {
```
imring, line 43: what about using?
```go
atk.Info.AttackTag != combat.AttackTagElementalBurst && atk.Info.ICDTag != combat.ICDTagAlhaithamProjectionAttack
```

imring:
it's also worth adding a `mirrors` field:
```go
func (c *char) Condition(fields []string) (any, error) {
	switch fields[0] {
	case "mirrors":
		return c.mirrorCount, nil
	default:
		return c.Character.Condition(fields)
	}
}
```

IvanHdzF:
> it's also worth adding a `mirrors` field:
> 
> ```go
> func (c *char) Condition(fields []string) (any, error) {
> 	switch fields[0] {
> 	case "mirrors":
> 		return c.mirrorCount, nil
> 	default:
> 		return c.Character.Condition(fields)
> 	}
> }
> ```

Took care of feedback received, thanks!

k0l11:
some rough snapshot timings assuming action start as in my [sheet](https://docs.google.com/spreadsheets/d/1qznqZMkFRRFKpzEXvQ_bJAXoleI2BrV-nKNAusJoYrU/edit#gid=408326586):
- na/ca: hitmark
- low plunge: presumably hitmark
- 1 mirror left/right: 20/21f
- 2 mirror left/right: 22/21f
- 3 mirror: no clue, like 6f before hitmark of the first tick might be fine?
- e/hold e: hitmark
- q: 67f

IvanHdzF:
@k0l11 Updated the snapshot timings and other observations in your sheet notes
I think we do still need values for attackHitlagHaltFrame and attackDefHalt? I put it on the TODOs
Also I did some assumptions with constellation 4 and 6, added those in the TODOs
I will be waiting for further observations, thanks!

internal/characters/alhaitham/cons.go
```diff
@@ -0,0 +1,110 @@
...
+// his CRIT Rate and CRIT DMG will increase by 10% and 70% respectively for 6s.
+// If this effect is triggered again during its initial duration, the duration remaining will be increased by 6s.
+const c6key = "alhaitham-c6"
+
+func (c *char) c6() {
+	m := make([]float64, attributes.EndStatType)
+	m[attributes.CR] = 0.1
+	m[attributes.CD] = 0.7
+
+	if c.StatModIsActive(c6key) { //TODO: Needs a check to work only during the first 6s of the buff?
```
k0l11, line 96: No, I don't think it needs a check like that. Checking if C6 is already up is correct.

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,259 @@
...
+		CanBeDefenseHalted: false,
+	}
+	c.Core.QueueAttack(ai, combat.NewCircleHit(c.Core.Combat.Player(), c.Core.Combat.PrimaryTarget(), combat.Point{Y: 2}, 2.25), skillHoldHitmark, skillHoldHitmark)
+
+	c.SetCDWithDelay(action.ActionSkill, 18*60, 23)
+
+	return action.ActionInfo{
+		Frames:          frames.NewAbilFunc(skillHoldFrames),
+		AnimationLength: skillHoldFrames[action.InvalidAction],
+		CanQueueAfter:   skillHoldFrames[action.ActionDash], // earliest cancel
```
k0l11, line 122: Should be `skillHoldFrames[action.ActionLowPlunge]`.

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,259 @@
...
+		// queue up again if we still have mirrors
+		if c.mirrorCount > 0 {
+			if c.recentlyMirrorGain { //if mirror has been gained recently, mirror is lost after 234f
+				c.Core.Tasks.Add(c.mirrorLoss(src), 234) //not affected by hitlag
+				return
+			}
+			c.Core.Tasks.Add(c.mirrorLoss(src), 214) //not affected by hitlag, 448-234
+
+		}
+		c.recentlyMirrorGain = false
```
k0l11, line 187: I think you can simplify this to removing a mirror every 234f. If you keep it, then the 234f is supposed to be only for the very first mirror loss instead of the other around.

internal/characters/alhaitham/cons.go
```diff
@@ -0,0 +1,110 @@
...
+				AffectedStat: attributes.EM,
+				Amount: func() ([]float64, bool) {
+					return m, true
+				},
+			})
+		}
+		c.Core.Log.NewEvent("c4 buff on loss", glog.LogCharacterEvent, c.Index)
+	}
+
+}
```
k0l11, line 80: Please look at the mirrorGain/mirrorLoss refactor comment in skill.go first. Please consider passing the amount of mirrors that were generated and consumed to this function and splitting this function into two, because the consumption and generation of mirrors on burst happens at different points in time. Also, the stacks are not counted separately. It could look like this:
```
func (c *char c4Loss(consumed int) {
        if consumed <= 0 {
                return
        }
        m := make([]float64, attributes.EndStatType)
        m[attributes.EM] = 30 * consumed
        for i, char := range c.Core.Player.Chars() {
                // skip Alhaitham
                if i == c.Index {
                        continue
                }
                        Base:         modifier.NewBaseWithHitlag("alhaitham-c4-loss", 900),
                        AffectedStat: attributes.EM,
                                return m, true
                        },
                })
        }
}

func (c *char) c4Gain(generated int) {
        if generated <= 0 {
                return
        }
        m := make([]float64, attributes.EndStatType)
        m[attributes.DendroP] = 0.1 * generated
        c.AddStatMod(character.StatMod{
                Base:         modifier.NewBaseWithHitlag("alhaitham-c4-gain", 900),
                AffectedStat: attributes.DendroP,
                        return m, true
                },
        })
}
```

internal/characters/alhaitham/alhaitham.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type char struct {
+	*tmpl.Character
+	recentlyMirrorGain bool
+	mirrorCount        int
+	lastInfusionSrc    int
+	a1ICD              int
+	c1ICD              int
+	c2counter          int
```
k0l11, line 27: nit: c2counter -> c2**C**ounter to be consistent

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
+	"github.com/genshinsim/gcsim/pkg/enemy"
+)
+
+var skillTapFrames []int
+var skillHoldFrames []int
+
+//TODO: defhalt values are needed
```
k0l11, line 17: TODO can be removed after comments below have been addressed.

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,259 @@
...
+func (c *char) mirrorLoss(src int) func() {
+	return func() {
+		if c.lastInfusionSrc != src {
+			c.Core.Log.NewEvent("mirror decrease ignored, src diff", glog.LogCharacterEvent, c.Index).
+				Write("src", src).
+				Write("new src", c.lastInfusionSrc)
+			return
+		}
+		if c.mirrorCount == 0 { //for case when you swap out and have mirrorloss queue'd. T
+			//TODO:change the lastinfusionsrc on swap event perhaps?
```
k0l11, line 168: Please set c.lastInfusionSrc to something like -1 in onExitField and remove the TODO.

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,293 @@
...
+	if hold == 1 {
+		return c.SkillHold()
+	}
+
+	c.Core.Tasks.Add(func() {
+		if c.mirrorCount == 0 { //extra mirror if 0 when cast
+			c.mirrorGain(2)
+			return
+		}
+		c.mirrorGain(1)
```
k0l11, line 64: Please consider moving this part out to its own little function to remove some duplication between `Skill` and `SkillHold`.

internal/characters/alhaitham/skill.go
```diff
@@ -0,0 +1,293 @@
...
+	if c.mirrorCount > 3 { //max 3 mirrors at a time.
+		c.mirrorCount = 3
+		queueOnFrame := false              //var tracks if this is the first overflowing in this frame
+		if c.Core.F == c.lastInfusionSrc { //check if c.lastinfusion has already been called on this frame
+			queueOnFrame = true
+		}
+		if !queueOnFrame { //this avoids multiple queues of mirror loss if mirror overflow multiple times in same frame
+			c.lastInfusionSrc = c.Core.F
+			c.Core.Tasks.Add(c.mirrorLoss(c.Core.F, 1), 234)
+		}
```
k0l11, line 158: You can simplify this part to just:
```
if c.Core.F != c.lastInfusionSrc { // this avoids multiple queues of mirror loss if mirror overflow multiple times in same frame
        c.lastInfusionSrc = c.Core.F
	c.Core.Tasks.Add(c.mirrorLoss(c.Core.F, 1), 234)
}
```

internal/characters/alhaitham/alhaitham.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/player/character/profile"
+)
+
+func init() {
+	core.RegisterCharFunc(keys.Alhaitham, NewChar)
+}
+
+type char struct {
+	*tmpl.Character
+	recentlyMirrorGain bool
```
k0l11, line 21: Please remove this since it's no longer used.

k0l11 (approved):

PR #1236, opened 2023-01-18
Added Light of Foliar Incision

RarePossum:


internal/weapons/sword/foliar/foliar.go
```diff
@@ -0,0 +1,88 @@
...
+		Base:         modifier.NewBase("foliar-crit-rate", -1),
+		AffectedStat: attributes.NoStat,
+		Amount: func() ([]float64, bool) {
+			return m, true
+		},
+	})
+
+	caBoost := 0.09 + 0.3*float64(r)
+	procCount := 0
+	c.Events.Subscribe(event.OnEnemyHit, func(args ...interface{}) bool {
```
k0l11, line 53: I think it would be cleaner to separate the activation of the effect and the actual effect itself. On that split you can then make the subscription that does the activation happen on OnEnemyDamage instead of OnEnemyHit. The flat dmg should stay on OnEnemyHit.

internal/weapons/sword/foliar/foliar.go
```diff
@@ -0,0 +1,88 @@
...
+		if atk.Info.ActorIndex != char.Index {
+			return false
+		}
+		if !(atk.Info.AttackTag == combat.AttackTagNormal || atk.Info.AttackTag == combat.AttackTagElementalArt || atk.Info.AttackTag == combat.AttackTagElementalArtHold) {
+			return false
+		}
+		// The buff is a ping dependent action, we're assuming the first hit won't
+		// have extra damage.
+		if !char.StatusIsActive(icdKey) {
+			if !(atk.Info.AttackTag == combat.AttackTagNormal) || atk.Info.Element == attributes.Physical {
```
k0l11, line 64: Please consider flipping the conditions and moving them into the if check above:
```
if !char.StatusIsActive(icdKey) && atk.Info.AttackTag == combat.AttackTagNormal && atk.Info.Element != attributes.Physical
```

internal/weapons/sword/foliar/foliar.go
```diff
@@ -0,0 +1,88 @@
...
+				return false
+			}
+			char.AddStatus(buffKey, 600, true)
```
k0l11, line 67: The effect also lasts for 12s = 720f

internal/weapons/sword/foliar/foliar.go
```diff
@@ -0,0 +1,88 @@
...
+		Amount: func() ([]float64, bool) {
+			return m, true
+		},
+	})
+
+	caBoost := 0.09 + 0.3*float64(r)
+	procCount := 0
+	c.Events.Subscribe(event.OnEnemyHit, func(args ...interface{}) bool {
+		atk := args[1].(*combat.AttackEvent)
+		if atk.Info.ActorIndex != char.Index {
```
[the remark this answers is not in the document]
k0l11: could be wrong on this one

internal/weapons/sword/foliar/foliar.go
RarePossum: It might still work if infusion times out but I'll do that anyway and can change if evidence to the contrary arises

k0l11 (approved):

PR #1261, opened 2023-01-28
adjust alhaitham low plunge default length

k0l11:
It is very unrealistic to consistently do the short variant of low plunge. After this PR it will default to the shortest long version frames and you can go back to the old behaviour via a param called `short` (has to be >0).

### DIFF
--- a/internal/characters/alhaitham/plunge.go
+++ b/internal/characters/alhaitham/plunge.go
@@ -10,15 +10,15 @@ import (
 
 var lowPlungeFrames []int
 
-const lowPlungeHitmark = 18
+const lowPlungeHitmark = 38
 
 func init() {
-	lowPlungeFrames = frames.InitAbilSlice(50)
-	lowPlungeFrames[action.ActionAttack] = 29
-	lowPlungeFrames[action.ActionSkill] = 30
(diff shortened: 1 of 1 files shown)


internal/characters/alhaitham/plunge.go
```diff
@@ -36,6 +36,12 @@ func (c *char) LowPlungeAttack(p map[string]int) action.ActionInfo {
 		}
 	}
 
+	short := p["short"]
+	skip := 0
+	if short > 0 {
+		skip = 20
```
unleashurgeek, line 42: why not have this param be a variable amount for folks to choose how much delay/how high the plunge is? Maybe rename this to `height` where 0 is the old behavior and we add more delay as `height` increases?

k0l11: because this is supposed to only give you a quick option between the super short variant and the shortest long variant. if you need any more frames on top you should be using the pre-action function/waits.

k0l11: The range of frames between the two variants above is what I would consider to be "impossible" to do in-game. Either you do the short variant or you're already at the long variant frame count at the very least.

unleashurgeek (approved):

PR #1265, opened 2023-01-29
fix constructs and zhongli stele

k0l11:
- fix construct not getting re-added on refresh
- fix zhongli stele
  - stele tick shouldn't also do resonance routine on the other stele
  - initial dmg should be able to generate energy
  - energy gen refactored into callback
  - energy gen check should not happen per stele/resonance attack
  - energy gen icd should apply to zhongli and be affected by hitlag 
https://gcsim.app/v3/viewer/share/be7ad1a6-df09-4a46-b198-4d8144dfeaae

### DIFF
--- a/internal/characters/zhongli/stele.go
+++ b/internal/characters/zhongli/stele.go
@@ -132,24 +132,18 @@ func (c *char) resonance(src int) func() {
 }
 
 func (c *char) steleEnergyCB() combat.AttackCBFunc {
-	done := false
 	return func(a combat.AttackCB) {
 		if a.Target.Type() != combat.TargettableEnemy {
 			return
 		}
-		// can only proc once per skill cast / stele tick
-		if done {
-			return
-		}
-		done = true
 		if c.StatusIsActive(particleICDKey) {
 			return
 		}
+		c.AddStatus(particleICDKey, 90, true)
 		// 50% chance
 		if c.Core.Rand.Float64() > 0.5 {
 			return
 		}
-		c.AddStatus(particleICDKey, 90, true)
 		c.Core.QueueParticle("zhongli", 1, attributes.Geo, 20+c.ParticleDelay)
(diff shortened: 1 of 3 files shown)


internal/characters/zhongli/stele.go
```diff
@@ -113,23 +119,37 @@ func (c *char) resonance(src int) func() {
...
-		c.Core.Tasks.Add(c.resonance(src), 120)
+		done = true
+		if c.StatusIsActive(particleICDKey) {
+			return
+		}
+		// 50% chance
+		if c.Core.Rand.Float64() > 0.5 {
+			return
+		}
+		c.AddStatus(particleICDKey, 90, true)
```
unleashurgeek, line 152: status should be added before the 50% check

internal/characters/zhongli/stele.go
```diff
@@ -113,23 +119,37 @@ func (c *char) resonance(src int) func() {
...
+func (c *char) steleEnergyCB() combat.AttackCBFunc {
+	done := false
+	return func(a combat.AttackCB) {
+		if a.Target.Type() != combat.TargettableEnemy {
+			return
 		}
-		if orb {
-			c.energyICD = c.Core.F + 90
-			c.Core.QueueParticle("zhongli", 1, attributes.Geo, 20+c.ParticleDelay)
+		// can only proc once per skill cast / stele tick
```
unleashurgeek, line 140: is done redundant since status will be added after first call (causing it to only proc once)?

k0l11: No, it's not because otherwise the other attacks in the stele tick will be able to take a shot at proccing that 50% chance.

k0l11: Nevermind.

unleashurgeek (approved):

PR #1269, opened 2023-02-01
fix sara q stormcluster initial offset

k0l11:
After further investigation, 3.6m seems to be more accurate than 6m.

### DIFF
--- a/internal/characters/sara/burst.go
+++ b/internal/characters/sara/burst.go
@@ -80,8 +80,8 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 		direction := combat.DegreesToDirection(i * stepSize).Rotate(burstInitialDirection)
 		// 6 ticks per stormcluster
 		for j := 0; j < 6; j++ {
-			// start at 6 + 1.35 = 7.35 m offset, move 1.35m per tick
-			stormClusterPos := combat.CalcOffsetPoint(burstInitialPos, combat.Point{Y: 6 + 1.35*float64(j+1)}, direction)
+			// start at 3.6 m offset, move 1.35m per tick
+			stormClusterPos := combat.CalcOffsetPoint(burstInitialPos, combat.Point{Y: 3.6 + 1.35*float64(j)}, direction)
 			stormClusterAp := combat.NewCircleHitOnTarget(stormClusterPos, nil, stormClusterRadius)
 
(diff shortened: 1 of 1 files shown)


internal/characters/sara/burst.go
```diff
@@ -80,8 +80,8 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
 		direction := combat.DegreesToDirection(i * stepSize).Rotate(burstInitialDirection)
 		// 6 ticks per stormcluster
 		for j := 0; j < 6; j++ {
```
unleashurgeek, line 82: I think this is 5 ticks and not 6 :/

The `AttackWave_DamageHandler` modifier has a duration of 1.4s and think interval of 0.3s. If `T` is the time when the think interval starts (`T = initial_hit_time + 0.8s`): 
1. tick at `T`
2. tick at `T + 0.3`
3. tick at `T + 0.6`
4. tick at `T + 0.9`
5. tick at `T + 1.2`

There is no 6th tick since it would be at `T + 1.5` which is after the modifier expires/gets removed (removed at `T + 1.4`).

5 ticks also matches the information in the [KQM TCL](https://library.keqingmains.com/evidence/characters/electro/kujou-sara#sara-burst-mechanics-update):
image

internal/characters/sara/burst.go
```diff
@@ -80,8 +80,8 @@ func (c *char) Burst(p map[string]int) action.ActionInfo {
...
-			// start at 6 + 1.35 = 7.35 m offset, move 1.35m per tick
-			stormClusterPos := combat.CalcOffsetPoint(burstInitialPos, combat.Point{Y: 6 + 1.35*float64(j+1)}, direction)
+			// start at 3.6 m offset, move 1.35m per tick
+			stormClusterPos := combat.CalcOffsetPoint(burstInitialPos, combat.Point{Y: 3.6 + 1.35*float64(j)}, direction)
 			stormClusterAp := combat.NewCircleHitOnTarget(stormClusterPos, nil, stormClusterRadius)
 
 			c.Core.QueueAttack(ai, stormClusterAp, burstStart, burstClusterHitmark+18*j, c1cb)
```
unleashurgeek, line 87: super nit?: First hit is at `burstInitialHitmark + 0.8s`, thus `burstClusterHitmark` should equal `99` (`burstInitialHitmark + 60 * 0.8`)

unleashurgeek (approved):

Issue #1131, opened 2022-12-06
player taking dmg is not implemented


k0l11:
related: #985 
Need to do a check for weapons/characters/artifacts that have to do with player dmg like Serpent Spine, Thoma C1, ...

k0l11:
initial implementation done in #1655, so I'm closing this issue

Issue #836, opened 2022-08-22
add hurt syntax from prehitlag


k0l11:
tagging #564 and #565

PR #1286, opened 2023-02-08
refactor some char events into callbacks

### LINKED ISSUE #1143: rewrite some OnEnemyDamage subscriptions to callbacks
In cases where it's feasible to attach a callback to a specific attack's dmg instead of using an OnEnemyDamage subscription we should opt to use a callback, because OnEnemyDamage subscriptions get executed on every single dmg instance.

Examples: Ganyu C1, Rosaria A1, ...

k0l11:
closes #1143

### DIFF
--- a/internal/characters/beidou/cons.go
+++ b/internal/characters/beidou/cons.go
@@ -8,9 +8,10 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/player"
 )
 
-const c4key = "beidou-c4"
+const c4Key = "beidou-c4"
 
-func (c *char) c4() {
+// Upon being attacked, Beidou's Normal Attacks gain an additional instance of 20% Electro DMG for 10s.
+func (c *char) c4Init() {
 	c.Core.Events.Subscribe(event.OnPlayerHPDrain, func(args ...interface{}) bool {
 		di := args[0].(player.DrainInfo)
 		if !di.External {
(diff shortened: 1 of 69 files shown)


internal/characters/beidou/cons.go
```diff
@@ -22,23 +23,26 @@ func (c *char) c4() {
...
 
-	c.Core.Events.Subscribe(event.OnEnemyDamage, func(args ...interface{}) bool {
-		t := args[0].(combat.Target)
-		ae := args[1].(*combat.AttackEvent)
-		if ae.Info.ActorIndex != c.Index {
-			return false
-		}
-		if ae.Info.AttackTag != combat.AttackTagNormal && ae.Info.AttackTag != combat.AttackTagExtra {
-			return false
+// TODO: this should also be added to her CA
```
unleashurgeek, line 33: alternative/addition to this change: Feel like this is something that should be available via param since we don't have player damage simmed?

IE: have a `attack[c4=1]` and `charge[c4=1]` param that will go and add the status as part of the attack routine?

internal/characters/eula/burst.go
```diff
@@ -117,44 +118,42 @@ func (c *char) triggerBurst() {
...
-		atk := args[1].(*combat.AttackEvent)
-		dmg := args[2].(float64)
-		if c.Core.Status.Duration(burstKey) == 0 {
-			return false
+// When Eula's own Normal Attack, Elemental Skill, and Elemental Burst deal DMG to opponents,
+// they will charge the Lightfall Sword, which can gain an energy stack once every 0.1s.
+func (c *char) makeBurstStackCB() combat.AttackCBFunc {
+	if c.Core.Status.Duration(burstKey) == 0 {
+		return nil
+	}
```
unleashurgeek, line 126: Should this be pushed into the callback? In the current state not 1:1 to old behavior iff you burst while skill/attack is in-flight -- (old behavior check happens when attack lands, new behavior when attack gets queued)

internal/characters/beidou/cons.go
k0l11: I don't feel comfortable with putting in a param for something that we could just simulate in some shape or form, see #836 and #1131. Also, we don't have params for the other on-damage procs.

unleashurgeek (approved):

PR #1292, opened 2023-02-18
Implements Dash CD

### LINKED ISSUE #892: dash cooldown is not implemented

unleashurgeek:
Closes #892

Example sim with dash CD and lockout being retained between swaps
https://gcsim.app/v3/viewer/share/7ea7f12d-4a2f-4ce9-bd1e-e79e75ba7616

**Note:** in current state:
* in-game 3N2CD (using macros) hits dash lockout on 3rd dash (so 2nd happens while first is in 0.8s CD w/ hitlag extension). Video: https://www.youtube.com/watch?v=IdBaKWc3-5Q
* In sim, dash lockout does not happen (with hitlag extension, first dash cd ends on frame 127 [see extension on frame 117], second dash starts on frame 128): https://gcsim.app/v3/viewer/share/5ac1e788-f3cb-4d22-807f-c8c6f334b08a
* in-game 3N2CD (using macros) will **only** hit lockout with hitlag extension

### DIFF
--- a/internal/characters/wanderer/dash.go
+++ b/internal/characters/wanderer/dash.go
@@ -23,10 +23,26 @@ func init() {
 }
 
 func (c *char) Dash(p map[string]int) action.ActionInfo {
+	delay := c.checkForSkillEnd()
+
 	if c.StatusIsActive(skillKey) {
 		return c.WindfavoredDash(p)
 	}
-	return c.Character.Dash(p)
+
+	// defer dash cd and stamina consumption til after delay (delay simulates falling)
+	c.Core.Tasks.Add(func() {
+		c.ApplyDashCD()
+		c.QueueDashStaminaConsumption(p)
+	}, delay)
+
+	// length is standard dash length + skill end delay (to simulate falling)
+	length := c.DashLength() + delay
+	return action.ActionInfo{
+		Frames:          func(action.Action) int { return length },
+		AnimationLength: length,
+		CanQueueAfter:   length,
+		State:           action.DashState,
+	}
 }
 
 func (c *char) WindfavoredDash(p map[string]int) action.ActionInfo {
--- a/internal/characters/wanderer/dash.go
+++ b/internal/characters/wanderer/dash.go
@@ -23,18 +23,10 @@ func init() {
 }
 
 func (c *char) Dash(p map[string]int) action.ActionInfo {
-	delay := c.checkForSkillEnd()
-
 	if c.StatusIsActive(skillKey) {
 		return c.WindfavoredDash(p)
 	}
-
-	ai := c.Character.Dash(p)
-	ai.Frames = func(action action.Action) int { return delay + ai.Frames(action) }
-	ai.AnimationLength = delay + ai.AnimationLength
-	ai.CanQueueAfter = delay + ai.CanQueueAfter
-
-	return ai
+	return c.Character.Dash(p)
 }
 
 func (c *char) WindfavoredDash(p map[string]int) action.ActionInfo {
--- a/internal/template/character/stam.go
+++ b/internal/template/character/stam.go
@@ -57,18 +57,21 @@ func (c *Character) Dash(p map[string]int) action.ActionInfo {
 	}
 }
 
+// set the dash CD. If the dash was on CD when this dash executes, lockout dash
 func (c *Character) ApplyDashCD() {
-	// set the dash CD. If the dash was on CD when this dash executes, lockout dash
+	var evt glog.Event
+
 	if c.Core.Player.DashCDExpirationFrame > c.Core.F {
 		c.Core.Player.DashLockout = true
 		c.Core.Player.DashCDExpirationFrame = c.Core.F + 1.5*60
+		evt = c.Core.Log.NewEvent("dash cooldown triggered", glog.LogCooldownEvent, c.Index)
 	} else {
 		c.Core.Player.DashLockout = false
 		c.Core.Player.DashCDExpirationFrame = c.Core.F + 0.8*60
+		evt = c.Core.Log.NewEvent("dash lockout evaluation started", glog.LogCooldownEvent, c.Index)
 	}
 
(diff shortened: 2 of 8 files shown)


internal/characters/wanderer/dash.go
```diff
@@ -23,18 +23,10 @@ func init() {
...
 	if c.StatusIsActive(skillKey) {
 		return c.WindfavoredDash(p)
 	}
-
-	ai := c.Character.Dash(p)
-	ai.Frames = func(action action.Action) int { return delay + ai.Frames(action) }
-	ai.AnimationLength = delay + ai.AnimationLength
-	ai.CanQueueAfter = delay + ai.CanQueueAfter
-
-	return ai
```
unleashurgeek, line 37: In existing implementation, `c.checkForSkillEnd()` will always return 0 if `c.StatusIsActive(skillKey)`. The modifying `ai` is useless since delay will always be 0 in these cases

`c.WindfavoredDash` will not run the dash CD logic as desired

k0l11: You have to at least keep the `c.checkForSkillEnd()` call in order to have the E state end properly here.

k0l11: It's possible that `c.StatusIsActive(skillKey)` is true, but `c.skydwellerPoints` are < 0. In this case the normal dash should be extended by `delay` (Wanderer falls down and only then dashes). So please revert this change.

internal/template/character/stam.go
```diff
@@ -42,18 +42,35 @@ func (c *Character) ActionStam(a action.Action, p map[string]int) float64 {
...
+	} else {
+		c.Core.Player.DashLockout = false
+		c.Core.Player.DashCDExpirationFrame = c.Core.F + 0.8*60
+	}
+
+	c.Core.Log.NewEventBuildMsg(glog.LogCooldownEvent, c.Index, "dash cooldown triggered").
+		Write("lockout", c.Core.Player.DashLockout).
+		Write("expiry", c.Core.Player.DashCDExpirationFrame-c.Core.F).
+		Write("expiry_frame", c.Core.Player.DashCDExpirationFrame).
+		Write("dash_length", length)
```
k0l11, line 60: nit: Please move this to the branch with the actual dash cooldown above or make this into two separate logs with the 0.8*60 one informing the user that dash cd will be triggered if they dash before now + 0.8s and the 1.5*60 one actually saying "dash cooldown triggered". Having the same exact log message for both of these cases is confusing.

internal/template/character/stam.go
```diff
@@ -42,18 +42,38 @@ func (c *Character) ActionStam(a action.Action, p map[string]int) float64 {
...
+		c.Core.Player.DashCDExpirationFrame = c.Core.F + 1.5*60
+	} else {
+		c.Core.Player.DashLockout = false
+		c.Core.Player.DashCDExpirationFrame = c.Core.F + 0.8*60
 	}
 
+	c.Core.Log.NewEventBuildMsg(glog.LogCooldownEvent, c.Index, "dash cooldown triggered").
+		Write("lockout", c.Core.Player.DashLockout).
+		Write("expiry", c.Core.Player.DashCDExpirationFrame-c.Core.F).
+		Write("expiry_frame", c.Core.Player.DashCDExpirationFrame)
```
k0l11, line 73: nit: Please move this to the branch with the actual dash cooldown above or make this into two separate logs with the `0.8*60` one informing the user that dash cd will be triggered if they dash before `now + 0.8s` and the `1.5*60` one actually saying "dash cooldown triggered". Having the same exact log message for both of these cases is confusing.

pkg/core/player/player.go
```diff
@@ -93,11 +99,35 @@ func (h *Handler) swap(to keys.Char) func() {
 	return func() {
 		prev := h.active
 		h.active = h.charPos[to]
-		h.Log.NewEvent("executed swap", glog.LogActionEvent, h.active).
-			Write("action", "swap").
-			Write("target", to.String())
+
+		// still have remaining frames left on dash CD, save in char for when the go on-field again
```
k0l11, line 103: nit: the -> the**y**

pkg/core/player/player.go
```diff
@@ -185,10 +215,23 @@ func (h *Handler) ApplyHitlag(char int, factor, dur float64) {
...
+
+	// extend the dash cd by the hitlag extension amount
+	if h.DashCDExpirationFrame > *h.F {
+		ext := int(math.Ceil(dur * (1 - factor)))
+		h.DashCDExpirationFrame += ext
+		h.Log.NewEvent("dash cd hitlag extended", glog.LogHitlagEvent, char).
+			Write("extension", ext).
+			Write("expiry", h.DashCDExpirationFrame-*h.F).
+			Write("expiry_frame", h.DashCDExpirationFrame).
+			Write("lockout", h.DashLockout)
```
k0l11, line 233: nit: Similarly to the comment in stam.go, the log message should probably be different depending on whether it's the 0.8s or 1.5s timer.

k0l11 (approved):

PR #1296, opened 2023-02-20
fix the attack speed buff for some characters

imring:


internal/characters/diluc/skill.go
```diff
@@ -76,6 +75,22 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
...
+			Base:         modifier.NewBaseWithHitlag("diluc-c6-speed", 360),
+			AffectedStat: attributes.AtkSpd,
+			Amount: func() ([]float64, bool) {
+				if c.Core.Player.CurrentState() != action.NormalAttackState {
+					return nil, false
+				}
+				if count > 1 {
+					return nil, false
+				}
+				return mAtkSpd, true
```
k0l11, line 91: The instance of `count` isn't shared between the two mods so you'd have to make it into a char variable if you want to share it.

k0l11 (approved):

Issue #461, opened 2022-04-12
Kazuha using plunge without skill should exit gracefully


srliao:
![image](https://user-images.githubusercontent.com/906239/163030649-548df084-0ae6-425f-9cc1-4098a9d79356.png)

Bug caused by Kazuha using plunge without skill so frame is 0 (and queuing f-1 causes negative).

Kazuha plunge action should have a check at start for skill usage and if skill not used should just exit (return 0,0) doing nothing

k0l11:
in rewrite, the sim does not crash if you try to use high_plunge without skill, so maybe this issue should be closed? the sim doesn't stop users from input invalid actions anyways (except setting frames to 500 if you try to use the last NA in a combo followed by a charge attack on some characters)

PR #1297, opened 2023-02-20
implement kazuha glide cancel

k0l11:
- glide cancel is now possible via the `glide_cancel` param on Kazuha's `skill` (>0 means glide cancel)
- small refactoring of skill.go
- restricts his plunge to only be allowed after a Skill without Glide Cancel

### DIFF
--- a/internal/characters/kazuha/plunge.go
+++ b/internal/characters/kazuha/plunge.go
@@ -5,6 +5,7 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/action"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 	"github.com/genshinsim/gcsim/pkg/core/combat"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
 )
 
 var plungePressFrames []int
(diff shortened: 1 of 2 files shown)


internal/characters/kazuha/skill.go
```diff
@@ -7,50 +7,60 @@ import (
...
-
-	if hold >= 1 {
-		radius = 9
+	glide := p["glide"]
+	if glide < 0 {
+		glide = 0
+	}
+	if glide > 1 {
+		glide = 1
 	}
```
unleashurgeek, line 56: nit: `glide_cancel` or similar for param name to make it clear that this is a glide cancel and not just gliding in sim (my impression of `glide` would be acting like a wait between the skill and plunge where you hover in air)

internal/characters/kazuha/skill.go
```diff
@@ -7,50 +7,60 @@ import (
...
 
 const (
 	skillPressHitmark = 10
+	skillPressRadius  = 5
 	skillPressCDStart = 8
 	skillHoldHitmark  = 33
+	skillHoldRadius   = 9
 	skillHoldCDStart  = 31
 	particleICDKey    = "kazuha-particle-icd"
 )
```
unleashurgeek, line 21: super nit: would be best if we reduced `skillPress` and `skillHold` to a single function since the same logic is being duplicated between them (just with different values).

Naive thought on this would be you need to pass through:
* `radius`
* stats skill multiplier array
* `hitmark`
* `particles`
* `cd` (change C1 to subtract 10%)
* `cdStartDelay`
* frames (for shortest action, either separate variable or do loop to find shortest)
* `glide`

Since all of the information (except `glide`) is completely static, rather than having many `skillHold` and `skillPress` variables, can define something like a `SkillInfo` struct that you create a `hold` and `press` instance within `init` (has to be init because golang very cool and doesnt allow constant structs).

Result would be your generic function with a header like:
```go
func (c *char) skillExecute(info *SkillInfo, glide int) action.ActionInfo {
  // ...
}
```

unleashurgeek (approved):

internal/characters/kazuha/plunge.go
```diff
@@ -32,10 +33,17 @@ func init() {
...
+	if c.Core.Player.LastAction.Type != action.ActionSkill ||
+		c.Core.Player.LastAction.Param["glide"] != 0 {
+		c.Core.Log.NewEvent("only plunge after skill without glide cancel", glog.LogActionEvent, c.Index).
+			Write("action", action.ActionLowPlunge)
+		return action.ActionInfo{
+			Frames:          func(action.Action) int { return 1200 },
+			AnimationLength: 1200,
+			CanQueueAfter:   1200,
+			State:           action.Idle,
+		}
```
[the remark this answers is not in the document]
k0l11: Kind of tracked here #461.

internal/characters/kazuha/skill.go
k0l11: Out of scope of the PR. If I was to do this then I would also take a look at all of the other skillTap/skillHold implementations.

PR #1303, opened 2023-02-23
combat package refactor

k0l11:


unleashurgeek (commented):
Why not move all of `pkg/core/combat/target.go` to `targets` package (keep all target stuff in one place)?

unleashurgeek (approved):

PR #1305, opened 2023-02-26
add yaoyao

imring:
continued #1244


internal/characters/yaoyao/yuegui.go
```diff
@@ -0,0 +1,179 @@
...
+	yg.Gadget = gadget.New(c.Core, pos, 0.5, combat.GadgetTypYueguiThrowing)
+
+	yg.Gadget.Duration = 600
+	yg.Gadget.OnThinkInterval = yg.throw
+
+	// they start throwing 29f after being spawned
+	yg.Gadget.ThinkInterval = 29
+
+	yg.Gadget.OnKill = func() {
+		yg.Core.Log.NewEvent("Yuegui (Throwing) expiry", glog.LogCharacterEvent, yg.c.Index)
```
k0l11, line 49: This is still not consistent with the log in `newYueguiJump`.

internal/characters/yaoyao/yuegui.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (yg *yuegui) getHeal() player.HealInfo {
+	if yg.GadgetTyp() == combat.GadgetTypYueguiThrowing {
+		return yg.c.getSkillHealInfo(&yg.snap)
+	}
+	return yg.c.getBurstHealInfo(&yg.snap)
+}
+
+// TODO: Confirm if yueguis can infuse cryo
```
k0l11, line 167: Yes, but only via attacks from the enemy/environment, which is not relevant in the sim.

internal/characters/yaoyao/yuegui.go
```diff
@@ -0,0 +1,179 @@
...
+
+	yg.Core.QueueAttackWithSnap(
+		yg.ai,
+		yg.snap,
+		radishExplodeAoE,
+		travelDelay,
+		yg.c.makeHealCB(radishExplodeAoE, hi),
+		yg.makeParticleCB(),
+		yg.c.makeC2CB(),
+	)
```
k0l11, line 153: You're missing the mechanic that the ai and hi is decided on explosion, so if the Yuegui (Throwing) radish hits an enemy when the burst is up, then the radish is burst dmg, has the burst icd and burst healing scaling. Snapshot doesn't change here.

k0l11: The same mechanic should apply to the A1 radish.

imring: ???

internal/characters/yaoyao/yuegui.go
```diff
@@ -0,0 +1,186 @@
...
+
+	yg.Core.QueueAttackWithSnap(
+		ai,
+		yg.snap,
+		radishExplodeAoE,
+		travelDelay,
+		yg.c.makeHealCB(radishExplodeAoE, hi),
+		yg.makeParticleCB(),
+		yg.c.makeC2CB(),
+	)
```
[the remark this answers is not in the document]
k0l11: The `radishExplodeAoE` should be moved above the task.

internal/characters/yaoyao/cons.go
```diff
@@ -69,20 +69,31 @@ func (c *char) c4() {
...
+	ai := combat.AttackInfo{
+		ActorIndex: yg.c.Index,
+		Abil:       "Mega Radish",
+		AttackTag:  attacks.AttackTagNone,
+		ICDTag:     attacks.ICDTagNone,
+		ICDGroup:   attacks.ICDGroupDefault,
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Dendro,
+		Durability: 25,
+		Mult:       0.75,
```
k0l11, line 82: Missing hitlag info from the other radishes.

internal/characters/yaoyao/skill.go
```diff
@@ -0,0 +1,74 @@
...
+		State:           action.SkillState,
+	}
+}
+
+func (c *char) getSkillHealInfo(snap *combat.Snapshot) player.HealInfo {
+	maxhp := snap.BaseHP*(1+snap.Stats[attributes.HPP]) + snap.Stats[attributes.HP]
+	heal := skillRadishHealing[0][c.TalentLvlSkill()]*maxhp + skillRadishHealing[1][c.TalentLvlSkill()]
+	return player.HealInfo{
+		Caller:  c.Index,
+		Target:  c.Core.Player.Active(),
```
k0l11, line 69: That doesn't work since the active char might be different by the time the heal cb is executed.

internal/characters/yaoyao/cons.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	hi := player.HealInfo{
+		Caller:  yg.c.Index,
+		Target:  yg.c.Core.Player.Active(),
```
k0l11, line 86: That doesn't work since the active char might be different by the time the heal cb is executed.

internal/characters/yaoyao/yuegui.go
```diff
@@ -0,0 +1,189 @@
...
+	if currHPPerc > 0.7 && enemy != nil {
+		target = enemy.Pos()
+	} else {
+		// really it should be random if no targets are in range and the character's HP is full but we aren't really simming that
+		target = yg.Core.Combat.Player().Pos()
+	}
+	ai, hi := yg.getInfos()
+	yg.c.QueueCharTask(func() {
+		radishExplodeAoE := combat.NewCircleHitOnTarget(target, nil, radishRad)
+		radishExplodeAoE.SkipTargets[targets.TargettablePlayer] = false
```
k0l11, line 145: Please move radishExplodeAoE out and the getInfos() inside the task.

internal/characters/yaoyao/asc.go
```diff
@@ -0,0 +1,72 @@
...
+
+	c.Core.QueueAttack(
+		ai,
+		radishExplodeAoE,
+		0,
+		travelDelay,
+		c.makeHealCB(radishExplodeAoE, hi),
+		c.makeC2CB(),
+	)
+}
```
k0l11, line 49: This is also subject to the same mechanic as used in `yuegui.go`'s `throw()`.

imring: why? `a1Ticker` checks if burst active

internal/characters/yaoyao/yuegui.go
```diff
@@ -98,6 +98,7 @@ func (c *char) makeHealCB(area combat.AttackPattern, hi player.HealInfo) func(co
 			return
 		}
 		if c.Core.Combat.Player().IsWithinArea(area) {
+			hi.Target = c.Core.Player.Active()
```
k0l11, line 101: Now this will also overwrite the hi of the burst radish which is supposed to heal everyone.

internal/characters/yaoyao/asc.go
k0l11: It could expire for the very last radish.

imring: where do i get the skill radish if, for example, i use the burst only?

k0l11: Put the skillRadishAI into a char variable on Init? Could also move burstRadishAI there.

k0l11: Alternatively you could put it in a function just like the heal info.

k0l11 (approved):

PR #1308, opened 2023-03-01
beacon initial implementation

IvanHdzF:
needs the on damage mechanic
has a parameter for stacks as a workaround

### DIFF
--- a/internal/weapons/claymore/beacon/beacon.go
+++ b/internal/weapons/claymore/beacon/beacon.go
@@ -45,7 +45,7 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
 	mHP := make([]float64, attributes.EndStatType)
 	mHP[attributes.HPP] = 0.24 + float64(r)*.08
 	char.AddStatMod(character.StatMod{
-		Base:         modifier.NewBase("beacon-of-the-reed-sea-atk", stackDuration),
+		Base:         modifier.NewBase("beacon-of-the-reed-sea-atk", -1),
 		AffectedStat: attributes.ATKP,
 		Amount: func() ([]float64, bool) {
 			count := 0
--- a/internal/weapons/claymore/beacon/beacon.go
+++ b/internal/weapons/claymore/beacon/beacon.go
@@ -45,7 +45,7 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
 	mHP := make([]float64, attributes.EndStatType)
 	mHP[attributes.HPP] = 0.24 + float64(r)*.08
 	char.AddStatMod(character.StatMod{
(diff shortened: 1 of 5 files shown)


internal/weapons/claymore/beacon/beacon.go
```diff
@@ -0,0 +1,68 @@
...
+type Weapon struct {
+	Index  int
+	stacks int
+}
+
+func (w *Weapon) SetIndex(idx int) { w.Index = idx }
+func (w *Weapon) Init() error      { return nil }
+
+func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile) (weapon.Weapon, error) {
+	//After the character's Elemental Skill hits an opponent, their ATK will be increased by 20% for 8s.
```
imring, line 25: where is atk% buff after using skill?

internal/weapons/claymore/beacon/beacon.go
```diff
@@ -0,0 +1,68 @@
...
+			return mATK, true
+		},
+	})
+	char.AddStatMod(character.StatMod{
+		Base:         modifier.NewBase("beacon-hp", -1),
+		AffectedStat: attributes.HPP,
+		Amount: func() ([]float64, bool) {
+			if !c.Player.Shields.PlayerIsShielded() {
+				mHP[attributes.HPP] = 0.24 + float64(r)*.08
+			}
```
imring, line 62: add attribute clearing `mHP[attributes.HPP] = 0` if the player has the shield

imring: or use `return nil, false` if the player has the shield:
```go
	mHP := make([]float64, attributes.EndStatType)
	mHP[attributes.HPP] = 0.24 + float64(r)*.08
		Base:         modifier.NewBase("beacon-hp", -1),
		AffectedStat: attributes.HPP,
			if c.Player.Shields.PlayerIsShielded() {
				return nil, false
			}
			return mHP, true
		},
	})
```

IvanHdzF:
Anything else?

internal/weapons/claymore/beacon/beacon.go
```diff
@@ -0,0 +1,95 @@
...
+
+	stackDuration := 480 //8s * 60
+	const skillKey = "beacon-of-the-reed-sea-skill"
+	const damagedKey = "beacon-of-the-reed-sea-damaged"
+
+	mATK := make([]float64, attributes.EndStatType)
+	mHP := make([]float64, attributes.EndStatType)
+	mHP[attributes.HPP] = 0.24 + float64(r)*.08
+	char.AddStatMod(character.StatMod{
+		Base:         modifier.NewBase("beacon-of-the-reed-sea-atk", stackDuration),
```
imring, line 48: `-1` instead of `stackDuration`

imring (approved):

PR #1338, opened 2023-04-19
Added delay action to be used for intuitive delays on actions.

Charlie-Zheng:
Delay[f=5] will delay the execution of the next action by 5 frames. It does not affect the timing of conditional checks.
Delay is intended to be used in combination with other actions, and the syntax allows it to be used easier with other action.

Example:
`ayaka dash, delay[f=3], attack` will delay the attack by 3 frames.
While `ayaka dash; wait(3); ayaka attack;` will not delay the attack at all.


srliao:
I'm of the opinion that we shouldn't add/change the syntax for now until #1148 is done. This may or may not be necessary depending on how hooks ends up being used

k0l11:
implemented as part of #1148

PR #1341, opened 2023-04-22
Delayed chongyun's skill field until hitmark

### LINKED ISSUE #1344: Chonyun's E field starts on Cast
Chonyun's E field currently starts on cast. It should start around the hitmark instead.

RarePossum:
Delayed Chongyun's Field creation to the same frame as hit hitmark.
fixes #1344

### DIFF
--- a/internal/characters/chongyun/skill.go
+++ b/internal/characters/chongyun/skill.go
@@ -133,11 +133,12 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 			c.Core.QueueAttackEvent(c.a4Snap, 0)
 		}, 665)
 	}
-
-	c.Core.Status.Add("chongyunfield", 600)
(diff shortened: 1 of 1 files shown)


internal/characters/chongyun/skill.go
```diff
@@ -133,11 +133,12 @@ func (c *char) Skill(p map[string]int) action.ActionInfo {
 			c.Core.QueueAttackEvent(c.a4Snap, 0)
 		}, 665)
 	}
-
-	c.Core.Status.Add("chongyunfield", 600)
+	c.QueueCharTask(func() {
```
k0l11, line 136: I don't think this task should be affected by hitlag.

k0l11: In this case it doesn't matter for now

k0l11 (approved):

PR #1351, opened 2023-05-06
baizhu bugfixes and cleanup

k0l11:
- fix baizhu e heal scaling
- fix baizhu q dmg scaling
- fix baizhu e healing only active char and not entire party
- fix baizhu e, q and c2 heal calculation (flat heal amount was being used as healing bonus instead of getting added to the heal amount)
- fix baizhu e detection radius
- fix baizhu e, ca, q and c2 hitbox
- fix baizhu c2 elemental icd
- fix baizhu particle icd (previously 2s icd, but you should be able to get 2x particles doing EE)


internal/characters/baizhu/skill.go
```diff
@@ -101,29 +97,33 @@ func (c *char) chain(src int, count int) combat.AttackCBFunc {
...
-	c.AddStatus(particleICDKey, 2*60, true)
-	c.Core.QueueParticle(c.Base.Key.String(), count, attributes.Dendro, c.ParticleDelay)
 }
 
 func (c *char) skillHealing() {
 	c.Core.Tasks.Add(func() {
 		c.Core.Player.Heal(player.HealInfo{
 			Caller:  c.Index,
-			Target:  c.Core.Player.Active(),
+			Target:  -1,
```
unleashurgeek, line 123: From skill descriptions: `the Sprite will return, healing all nearby party members based on Baizhu's Max HP.` Sure this should be all and not active only (not clear from dm comparing the logic between other heals)?

k0l11: yes I'm sure. `nearby party members` = everyone

unleashurgeek (approved):

PR #1357, opened 2023-05-06
baizhu frames + bugfixes

k0l11:
- fixes baizhu e not ticking more than once in single target
- fixes baizhu q healing on very first tick

### DIFF
--- a/internal/characters/baizhu/burst.go
+++ b/internal/characters/baizhu/burst.go
@@ -23,8 +23,12 @@ const (
 )
 
 func init() {
-	burstFrames = frames.InitAbilSlice(115)
-	burstFrames[action.ActionSwap] = 114
+	burstFrames = frames.InitAbilSlice(105) // Q -> CA/D
+	burstFrames[action.ActionAttack] = 104
+	burstFrames[action.ActionSkill] = 104
+	burstFrames[action.ActionJump] = 104
+	burstFrames[action.ActionWalk] = 104
+	burstFrames[action.ActionSwap] = 102
 }
 
 func (c *char) Burst(p map[string]int) action.ActionInfo {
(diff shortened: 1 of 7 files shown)


internal/characters/baizhu/skill.go
```diff
@@ -77,13 +78,15 @@ func (c *char) chain(src int, count int) combat.AttackCBFunc {
...
-		})
+		next := c.Core.Combat.RandomEnemyWithinArea(combat.NewCircleHitOnTarget(a.Target, nil, 10), nil)
 		if next == nil {
 			c.skillHealing()
 			return
 		}
+		delay := skillTickInterval
+		if next.Key() == a.Target.Key() {
+			delay += 6 // add some delay in case it's a different target
+		}
```
skippi, line 89: - The comment here doesn't seem to match what's happening with the `if` check. It looks like we're adding delay if the target stays the same.
- Should we document this 6f delay in the baizhu frame sheet?

internal/characters/baizhu/burst.go
```diff
@@ -12,27 +12,41 @@ import (
...
+	burstRefreshRate  = 146
+	burstShieldExpiry = 152
+	// subject to change
+	burstTickRelease = 21
+	burstTickTravel  = 8
+)
 
 func init() {
 	burstFrames = frames.InitAbilSlice(115)
 	burstFrames[action.ActionSwap] = 114
```
skippi, line 27: Missing burst cancels here from the sheet.

skippi (approved):

skippi (changes requested):
Meant for request changes.

internal/characters/baizhu/skill.go
k0l11: I wouldn't. All the values here are dependent on distance so I just picked some reasonable value from a quick test in-game

skippi (approved):

PR #1369, opened 2023-06-07
Make Nahida hold duration more intuitive

RarePossum:
Make Nahida's Hold Duration be the amount from frames its held longer than the minimum rather than the exact number of frame until hitmark. This is more intuitive to use


internal/characters/nahida/skill.go
```diff
@@ -90,13 +90,14 @@ func (c *char) skillHold(p map[string]int) action.ActionInfo {
...
-	if hold > 317 {
-		hold = 317
+	if hold > 300 {
+		hold = 300
 	}
-	if hold < 17 {
-		hold = 17
+	if hold < 1 {
+		hold = 1
 	}
```
unleashurgeek, line 98: can you also update the comment to reflect this change

unleashurgeek (approved):

PR #1373, opened 2023-06-13
add gadget selection and fix ayaka ca, lisa q at c4, yae e

### LINKED ISSUE #1342: Yae Fischl and Lisa do not target cores
Yae E, Oz and Lisa Q never target cores, Even if they target at a lower priority, currently they will never hit a core.

Seperate issue is bloom hitbox size but thats a big question mark.

k0l11:
- move enemy selection methods from enemy.go to select.go
- add gadget equivalents for enemy selection methods in an area (needed for ayaka ca and lisa q at c4, yae e)
- fix ayaka ca (spawns up to 5 attacks, prioritises all enemies first and then all gadgets)
- fix lisa q at c4 (spawns up to 3 attacks, uses enemy + gadget count for dischargeCount decision making, prioritises all enemies first and then all gadgets)
- fix yae e (spawns 1 attack, prioritises enemy over gadget)

closes #1342


pkg/core/combat/select.go
```diff
@@ -0,0 +1,279 @@
...
+		if !v.IsAlive() {
+			continue
+		}
+		if !v.IsWithinArea(a) {
+			continue
+		}
+		gadgets = append(gadgets, v)
+	}
+	return gadgets
+}
```
unleashurgeek, line 52: nit: can simplify this down to a single helper function with generics (I think, not a go expert)
```go
func withinAreaFiltered[T Gadget | Enemy](a AttackPattern, filter func(t T) bool, original []Target) []T {
	var out []T
	hasFilter := filter != nil
	for _, v := range original {
		if v == nil {
			continue
		}
		if hasFilter && !filter(v) {
			continue
		}
		if !v.IsAlive() {
			continue
		}
		if !v.IsWithinArea(a) {
			continue
		}
		out= append(out, v)
	}
	return out
}
```

you could technically do the same for the handler methods, but you'd need to if/else based on `T` type since you need to call `c.gadgets` or `c.enemies`, thought it would be good if we can minimize the amount of copy/pasted logic in the other functions

internal/characters/yaemiko/kitsune.go
```diff
@@ -142,17 +143,32 @@ func (c *char) kitsuneTick(totem *kitsune) func() {
...
+		// try to target an enemy first
+		enemy := c.Core.Combat.RandomEnemyWithinArea(totem.kitsuneArea, nil)
+		if enemy != nil {
+			pos = enemy.Pos()
+		} else {
+			// target gadget if no enemy was targeted
+			gadget := c.Core.Combat.RandomGadgetWithinArea(totem.kitsuneArea, nil)
+			pos = gadget.Pos()
+		}
+		tick(pos)
```
unleashurgeek, line 170: Need a `nil` check for case of no gadget is found and `pos` remained nil/unspecified

pkg/core/combat/select.go
k0l11: you are not allowed to use interfaces that contain methods in a union, so `Gadget | Enemy` is not legal

unleashurgeek (approved):

PR #1378, opened 2023-06-14
change icd var initialization

### LINKED ISSUE #1245: Change ICD var initialization
ICD variable initialization should be moved into an `init()` function:

https://github.com/genshinsim/gcsim/blob/dd5eccb70baaaabe8b99ef8431b4919584d34330/pkg/core/combat/icd.go#L136-L140

They should be set like:

```go
func init() {
   ICDGroupEleApplicationSequence[ICDGroupDefault] = {1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0}
}
```

This avoids any issues that would be caused by const reordering

k0l11:
closes #1245

### DIFF
--- a/pkg/core/attacks/icd.go
+++ b/pkg/core/attacks/icd.go
@@ -86,148 +86,97 @@ const (
 	ICDGroupLength
 )
 
-var ICDGroupResetTimer = []int{
-	150, //default
-	60,  //amber
-	60,  //venti
-	300, //fischl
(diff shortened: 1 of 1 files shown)


k0l11:
see #1245
> This avoids any issues that would be caused by const reordering

PR #1387, opened 2023-06-17
change some mods from expiry -1

k0l11:
working on #203 

- done
	- crimson witch 4pc -> 2pc increase (also adjust mod name: cw-4pc -> crimson-4pc-stacks)
	- flower of paradise lost 4pc increase
	- mika a1
	- razor q speed burst (also adjust mod name: speed-burst -> razor-q)
	- razor e energy sigil (also adjust mod name: er-sigil/razor-sigil-duration -> razor-sigil)
	- mappa mare (also adjust mod name: mappa/mappa-mare-stacks -> mappa-mare)
	- beacon
- TODO (part of separate PR after mods and stats array rewrite since these require new enemy mods, namely ones to influence dmg bonus/crit rate buffs tied to an enemy):
	- ganyu c4 damage bonus
	- mona burst damage bonus
	- mona c4 crit rate
	- toukabou shigure damage bonus
- considered changing these but I don't think it's feasible (**person reviewing the PR should double-check**):
	- husk 4pc
	- nymph's dream 4pc
	- aloy coil mod
	- ayaka c6
	- ayato c2
	- candace a4
	- candace burst init
	- heizou c6
	- sayu c2
	- shenhe c4
	- polar star
	- thundering pulse
	- fruit of fulfillment 
	- serpent spine 
	- alley flash 
	- mistsplitter 
	- lost prayer

### DIFF
--- a/internal/weapons/catalyst/mappa/mappa.go
+++ b/internal/weapons/catalyst/mappa/mappa.go
@@ -17,7 +17,8 @@ func init() {
 }
 
 type Weapon struct {
-	Index int
+	stacks int
+	Index  int
 }
 
 func (w *Weapon) SetIndex(idx int) { w.Index = idx }
@@ -27,9 +28,10 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
 	w := &Weapon{}
 	r := p.Refine
 
-	stacks := 0
-	const stackKey = "mappa-mare-stacks"
-	stackDuration := 600 // 10s * 60
+	dmg := 0.06 + float64(r)*0.02
+
+	const buffKey = "mappa-mare"
+	buffDuration := 600 // 10s * 60
 
 	addStack := func(args ...interface{}) bool {
 		atk := args[1].(*combat.AttackEvent)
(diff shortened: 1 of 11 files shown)


internal/weapons/catalyst/mappa/mappa.go
```diff
@@ -40,46 +42,39 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
...
 		}
+
+		m := make([]float64, attributes.EndStatType)
+		m[attributes.PyroP] = dmg * float64(w.stacks)
+		m[attributes.HydroP] = dmg * float64(w.stacks)
+		m[attributes.CryoP] = dmg * float64(w.stacks)
+		m[attributes.ElectroP] = dmg * float64(w.stacks)
+		m[attributes.AnemoP] = dmg * float64(w.stacks)
+		m[attributes.GeoP] = dmg * float64(w.stacks)
+		m[attributes.DendroP] = dmg * float64(w.stacks)
```
unleashurgeek, line 59: nit: cleaner to make list of attributes to set and loop over:
```go
for _, attr := range elementAttributes {
  m[attr] = dmg * float64(w.stacks)
}

unleashurgeek (commented):
Currently going through the list of other mods in description. Will comment thoughts on each as I go through

unleashurgeek:
husk 4pc: No clean solution imo. Only way I can think of:
  - create a stat mod per stack (husk-4pc-stack-#)
  - create an "add stack" helper function. When adding a stack you reset the durations of all stacks
  - durations should be set such that 1 stack is lost every 6s. IE:
    - `husk-4pc-stack-1`: 6s
    - `husk-4pc-stack-2`: 12s
    - `husk-4pc-stack-3`: 18s
    - `husk-4pc-stack-4`: 24s
  - delete stack loss logic since it now happens naturally with the above durations

unleashurgeek:
nymph's dream 4pc (to me is looks like current implementation is incorrect? In the current state, you'll get a status effect even when active stacks > 3. So then when 1 status effect expires, another will immediately take over without the character doing anything):
* stat mod per type
* in damage event, loop over all possible stat mods. If 3 are active, do nothing
* else, add a new stat mod of that attack type

internal/weapons/catalyst/mappa/mappa.go
k0l11: leaving as is (many other places do it like this and I don't feel like changing them in this PR)

k0l11:
> to me is looks like current implementation is incorrect? In the current state, you'll get a status effect even when active stacks > 3.
implementation should be correct, see the [artifact description](https://genshin-impact.fandom.com/wiki/Nymph%27s_Dream)

k0l11:
> husk 4pc: No clean solution imo.
> nymph's dream 4pc

fun fact: they also have it as an expiry -1 mod

unleashurgeek:
candace burst init:
* Add attack mod to all chars when burst happens with the given duration
* keep the conditional logic in the amount func

unleashurgeek:
shenhe C4:
* keep the conditional logic in the amount func
* Replace C4BuffKey with adding this mod w/ the 60s duration (and deleting in reset CB)

unleashurgeek (approved):

Issue #124, opened 2021-12-31
OnReaction hook for all reactions


srliao:
Add some sort of wrapper to simplify if we need to subscribe to all reactions at once

srliao:
@HanSeoulOh this is not done btw

k0l11:
Do we really need this?

k0l11:
closing in favor of #1300

PR #1388, opened 2023-06-17
fix shatter in all reaction event subs

k0l11:
related to #124

As of 3.7, only Nahida E and Mailed Flower consider Shatter as an elemental reaction when it comes to triggering on "all" reactions.


unleashurgeek (approved):
RE https://github.com/genshinsim/gcsim/issues/124. Would probably just close in favor of an overall event refactor

PR #1389, opened 2023-06-17
fix shenhe c4

k0l11:
- change from expiry -1 to modifying skill snapshot instead 
- only applied to first enemy in AoE before
- stack gain via E was only +1 in AoE before

live: https://gcsim.app/v3/viewer/share/dc0df522-8415-465e-a89c-7d4c66680c31
pr: https://gcsim.app/v3/viewer/share/86295336-c2d7-4b42-9e59-3b4225127751

### DIFF
--- a/internal/characters/shenhe/cons.go
+++ b/internal/characters/shenhe/cons.go
@@ -1,7 +1,6 @@
 package shenhe
 
 import (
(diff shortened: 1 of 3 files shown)


k0l11:
modifying the dmg% of the attack is also how *they* do it

unleashurgeek (approved):

PR #1390, opened 2023-06-17
change heizou c6 from expiry -1

k0l11:
slight change in snapshot timing is fine (~2f)


k0l11:
this is also how *they* do it

PR #1405, opened 2023-07-04
add substatOptimFull flag to cli

### LINKED ISSUE #1398: add cli option for running substat optimizer and normal sim in one command
At the moment, the user first has to run the sim with their config in substat optimizer mode, creating a new text file for the optimized config. Then they have to run the sim again, but this time without the substat optimizer using the optimized config as input. This workflow could be simplified by allowing these two steps to be done one after another in the same command via a new flag for the cli tool.

k0l11:
closes #1398 

Workflow to get a .gz viewer file for a given config in combination with the substat optimizer:

before:
```
./gcsim.exe -c="test.txt" -substatOptim -out="test_opt.txt"
./gcsim.exe -c="test_opt.txt" -gz -out=test
```

after:
```
./gcsim.exe -c="test.txt" -substatOptimFull -gz -out=test
```

key differences:
- before: create a file holding the optimized config under "test_opt.txt"
- after: overwrite the content of the original config file "text.txt" to contain the optimized config

If the user wants to adjust substats after running the new workflow, they can simply adjust the config file and run the sim again, but this time without the `-substatOptimFull` flag.

The old workflow is still available.

### DIFF
--- a/cmd/gcsim/main.go
+++ b/cmd/gcsim/main.go
@@ -28,18 +28,19 @@ var (
 )
 
 type opts struct {
-	config       string
-	out          string //file result name
-	gz           bool
-	serve        bool
-	nobrowser    bool
-	keepserving  bool
-	substatOptim bool
-	verbose      bool
-	options      string
-	debugMinMax  bool
-	cpuprofile   string
-	memprofile   string
+	config           string
+	out              string //file result name
+	gz               bool
+	serve            bool
+	nobrowser        bool
+	keepserving      bool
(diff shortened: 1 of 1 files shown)


unleashurgeek:
imo we should look into cleaning up and rebuilding our cli from scratch. On the srsim side, using this to build the cli for a more consistent and extendable experience (code not yet pushed): https://cli.urfave.org/v2/examples/full-api-example/

unleashurgeek (approved):

PR #1410, opened 2023-07-05
database infinite scroll

srliao:
change pagination to infinite scroll on db

TODO: really server should respond back if there are more data... right now this relies on sort of a hacky solution


srliao:
forgot to check the tag view... will convert back to pr when ready

srliao:
fixed infinite scroll on tag bag, but also temporary disabled the route for now until we are sure of what route we want

unleashurgeek (approved):

PR #1414, opened 2023-07-05
the better date format

srliao:
image


srliao:
nevermind forget it. realized this doesnt translate pepela

PR #1419, opened 2023-07-06
lumine db temp fix

srliao:


unleashurgeek:
I do not know if I would consider this a fix since it goes from lumine breaking searches to it being impossible to search the traveler

srliao:
> I do not know if I would consider this a fix since it goes from lumine breaking searches to it being impossible to search the traveler

traveler search is broken regardless for a different reason. the config file for traveler currently specifies `traveler<element>` as key, but in reality we use both `lumine<element>` and `aether<element>`, which is what is stored in db

iirc i need to move traveler logic to a common folder (similar to how we handle weapons) and then have individual folders for lumine and aether, with respective keys. 

imo this should be merged to fix the crashing and traveler search disabled for now

unleashurgeek:
We should disable by removing traveler as an option from search then? In current state, going to confuse users in thinking that there are no sims with traveler since those searches will return 0 results?

srliao:
> We should disable by removing traveler as an option from search then? In current state, going to confuse users in thinking that there are no sims with traveler since those searches will return 0 results?

Yes should do that. I can't get to it this weekend though. Maybe sometimes next mid next week

PR #1420, opened 2023-07-06
Add possibility for normal cons

k0l11:
- fix ayato c3 and c5 being swapped
- from this point onwards, skill cons being 3 and burst cons being 5 is no longer default and char implementations have to specify NormalCon/SkillCon/BurstCon if they should be enabled at a certain constellation, default value is -1 which means that it doesn't apply at any constellation

### DIFF
--- a/pkg/simulation/run.go
+++ b/pkg/simulation/run.go
@@ -17,9 +17,9 @@ import (
 func (s *Simulation) Run() (res stats.Result, err error) {
 	defer func() {
 		// recover from panic if one occured. Set err to nil otherwise.
-		if recover() != nil {
+		if r := recover(); r != nil {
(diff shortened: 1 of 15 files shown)


srliao:
wouldn't it be better to panic (and thereby hopefully fail tests or something lik that) if cons are not set? ideally every char must have it set?

k0l11:
@srliao added

pkg/simulation/run.go
```diff
@@ -17,9 +17,9 @@ import (
 func (s *Simulation) Run() (res stats.Result, err error) {
 	defer func() {
 		// recover from panic if one occured. Set err to nil otherwise.
-		if recover() != nil {
+		if r := recover(); r != nil {
```
srliao, line 20: nit: technically `r` should be `err`

srliao (approved):

PR #1422, opened 2023-07-06
add docs for heizou

k0l11:
Also update the checksum for our Prism repo since I was getting "The remote archive doesn't match the expected checksum"

### DIFF
--- a/ui/packages/docs/docs/reference/characters/heizou.md
+++ b/ui/packages/docs/docs/reference/characters/heizou.md
@@ -1,0 +1,29 @@
+---
+title: Heizou
+---
+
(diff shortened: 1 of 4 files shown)


srliao:
checksum change seems to be causing tests to fail. not entirely sure why

srliao:
@unleashurgeek can you please take a look at if yarn change ok?

unleashurgeek:
Both the checksum and yarn config should remain unmodified (`68f79` is the correct checksum). Unclear why first commit changed this checksum

unleashurgeek (approved):

PR #1531, opened 2023-07-09
remove unused blog folder from docs

k0l11:


k0l11:
closing this because who knows, we might want to have a blog page in the future

PR #1574, opened 2023-07-19
fix character and weapon select

srliao:
character and weapon select should now use pipeline data and stay up to date with core changes (as long as `go generate` is ran)


ui/packages/ui/src/Components/Select/weapons.tsx
```diff
@@ -2,140 +2,9 @@ import { MenuItem } from "@blueprintjs/core";
...
-  "wastergreatsword",
-  "wavebreakersfin",
-  "whiteblind",
-  "whiteirongreatsword",
-  "whitetassel",
-  "windblumeode",
-  "wineandsong",
-  "wolfsgravestone",
-];
+export const weapons: IWeapon[] = valid_weapons.map(e => e)
```
k0l11, line 7: why the map?

ui/packages/ui/src/Data/index.ts
```diff
@@ -8,23 +9,60 @@ export interface CharDataMap {
...
+    case "WEAPON_CATALYST":
+      return "catalyst";
+    default:
+      return "unknown";
+  }
 }
 
-export function TravelerCheck(key: string) {
-  return key.startsWith("aether") || key.startsWith("lumine") || key.startsWith("traveler");
+export const valid_weapons: string[] = Object.keys(weaponPipelineData.data).map(
```
k0l11, line 50: why the map?

ui/packages/ui/src/Components/Select/weapons.tsx
srliao: fixed

ui/packages/ui/src/Data/index.ts
srliao: fixed

k0l11 (approved):

PR #1576, opened 2023-07-19
add artifact names to pipeline

srliao:
this should resolve everything that's blocking for #1520


pipeline/cmd/pipeline/main.go
```diff
@@ -59,9 +63,24 @@ func main() {
...
+	//generate artifact data
+	log.Println("running pipeline for artifacts...")
+	ga, err := artifact.NewGenerator(artifact.GeneratorConfig{
+		Root: cfg.artifactPath,
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	log.Println("generate weapon data for ui...")
```
k0l11, line 81: should say artifact instead of weapon

k0l11 (approved):

PR #1577, opened 2023-07-19
change pipeline to output pretty json

srliao:


k0l11:
missing commit of the pretty json pipeline data

PR #1589, opened 2023-07-30
add back character actions statistic

### LINKED ISSUE #1564: add back character ability usage to viewer page
min, max, average

k0l11:
closes #1564 

- actions are always sorted according to an (arbitrary) pre-defined order that is similar to docs
- consistency with docs is not ensured and color choice may not be perfect
- legend order + mapping between color and action is consistent across sims

Example images: 
grafik
grafik


ui/packages/ui/src/Pages/Viewer/Components/Miscellaneous/CharacterActionsBarChart/BarChart.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+type Props = {
+  width: number;
+  height: number;
+  names?: string[];
+  actions?: SourceStats[];
+  actionNames?: string[] | null;
+}
+
+const margin = { top: 0, left: 300, right: 20, bottom: 40 };
```
unleashurgeek, line 15: Should set the left margin similar to the Character DPS bar graphs (I need to write something that is more dynmaic and based on max number of characters in label or something).

unleashurgeek:
this graph also probably doesn't need full width and can split with the field time pie 50/50?

k0l11:
new images:
grafik
grafik

unleashurgeek (approved):
nit: for added clarity, might be nice for this graph to also have an x-axis label of something like "count"

PR #1591, opened 2023-07-31
add back reaction/non-reaction source dps split

### LINKED ISSUE #1563: add way to split source dps chart into non-reaction/reaction dmg
for example: no reaction vs. melt version of ability

k0l11:
closes #1563 

Splits the abilities of the source dps graph into reaction/non-reaction with no option to group them together again for now.

Example image:
grafik

### DIFF
--- a/pkg/agg/damage/damage.go
+++ b/pkg/agg/damage/damage.go
@@ -105,7 +105,12 @@ func (b *buffer) Add(result stats.Result) {
 			charTargetDPS[ev.Target] += ev.Damage
 			charElementDPS[ev.Element] += ev.Damage
 			charDPS += ev.Damage
-			sourceDPS[ev.Source] += ev.Damage
+			sourceDPSKey := ev.Source
+			reactModifier := string(ev.ReactionModifier)
+			if reactModifier != "" {
+				sourceDPSKey += " (" + reactModifier + ")"
+			}
+			sourceDPS[sourceDPSKey] += ev.Damage
 		}
 
 		b.characterDPS[i].Add(charDPS * time)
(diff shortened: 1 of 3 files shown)


pkg/agg/damage/damage.go
```diff
@@ -105,7 +105,12 @@ func (b *buffer) Add(result stats.Result) {
 			charTargetDPS[ev.Target] += ev.Damage
 			charElementDPS[ev.Element] += ev.Damage
 			charDPS += ev.Damage
-			sourceDPS[ev.Source] += ev.Damage
+			sourceDPSKey := ev.Source
+			reactModifier := string(ev.ReactionModifier)
+			if reactModifier != "" {
+				sourceDPSKey += " (" + reactModifier + ")"
```
unleashurgeek, line 111: I'm not a fan of this approach. If in the future we ever add proper game mappings for keys, we will be unable to easily leverage them in UI for localization since we would need to string parse to get the original sourceDPSKey. Ideally, data model should be changed to add react modifier as a new dimension and then UI can decide if it wants to merge like so or add this as a toggle on the graph or what. The main problem with this is if you change the data structure for this data, you make the new UI incompatible with the older version on source dps data unless you also program backwards compatibility which is a pain in the ass and imo not worth it unless you have a clear idea on how you'd leverage this extra information without concatenating in UI like you're doing here.

At minimum: you should make the react modifier more human readable and have some hand crafted map of what the user-facing keys here should be

k0l11:
new: 
grafik

pkg/agg/damage/damage.go
k0l11: did the minimum without adding a handcrafted map since I think the string representation of ReactionModifier is good enough (vaporize/melt/spread/aggravate)

unleashurgeek (approved):

PR #1603, opened 2023-08-03
adjust noelle extend param default

k0l11:
https://discord.com/channels/845087716541595668/1034835649762689105
- no longer assumes max extension/c6 extension bug by default

Breaking change for all noelle sims that currently do not explicitly specify the `extend` param on `burst`.
I think it's worth it, because the max extension via c6 bug is probably not reasonable to assume for noelle sims since it depends on the rotation whether it can be achieved or not.


k0l11:
closed following discussion here https://discord.com/channels/845087716541595668/983391844631212112/1136729431671578715

PR #1616, opened 2023-08-11
fix icd reset timer to be >t instead of >t-2

k0l11:
ICD timer should reset *after* the timer duration has passed, which means > timer. Current implementation is > timer - 2.


k0l11:
Closing in favor of #1617 after discussing it internally.

PR #1618, opened 2023-08-12
Make enemy example sensical

Charlie-Zheng:
Somewhat based on this guy: https://genshin-impact.fandom.com/wiki/Consecrated_Horned_Crocodile
I applied 2.5x abyss multiplier on the HP values.

Instead of dropping 2 Hydro orbs, I've adjusted it to drop 1 clear orb, which gives the same amount of energy to non Hydro units.

### DIFF
--- a/ui/packages/docs/docs/reference/config.md
+++ b/ui/packages/docs/docs/reference/config.md
@@ -98,7 +98,7 @@ If you set this to a negative number or a really large number, behaviour is unde
 
 Example:
 ```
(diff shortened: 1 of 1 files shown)


k0l11:
I don't see the value of this change without an added explanation of where to find the enemy lvl, calc the enemy hp, etc. The example is literally just to show the syntax.

Charlie-Zheng:
What is a realistic example a *problem*

srliao:
Closed per discussion here: https://discord.com/channels/845087716541595668/983391844631212112/1141024714052075640

PR #1626, opened 2023-08-16
add lyney

k0l11:
- [X] New character package
- [X] Config in character package
- [X] Run pipeline with added config
- [X] Character key
- [X] Shortcuts for character key
- [X] Add Character package to imports
- [X] Character talent stats
- [X] Character curve
- [X] Normal Attack
- [X] Charge Attack / Aimed Shot
- [X] Skill
- [X] Burst 
- [X] A1
- [X] A4
- [X] C0
- [X] C1
- [X] C2
- [X] C3
- [X] C4
- [X] C5
- [X] C6
- [X] Other necessary talents (custom dash/jump, low/high plunge, ...)
- [X] Hitlag
- [X] ICD
- [X] StrikeType
- [X] Hitboxes
- [X] Attack durability
- [X] Particles
- [x] Frames
- [X] Update documentation


soloxcx:
Looks like C0 Lyney is trying to summon two hats as if he is C1, and the first hat is destroyed instantly.

[sample.gz](https://github.com/genshinsim/gcsim/files/12359641/serve_sample.json.gz)

internal/characters/lyney/aimed.go
```diff
@@ -0,0 +1,240 @@
...
+			combat.NewBoxHit(
+				c.Core.Combat.Player(),
+				c.Core.Combat.PrimaryTarget(),
+				geometry.Point{Y: -0.5},
+				0.1,
+				1,
+			),
+			0,
+			travel,
+			c.skillAlignedCB,
```
k0l11, line 141: Should work even when not hitting an enemy

internal/characters/lyney/burst.go
```diff
@@ -0,0 +1,142 @@
...
+		if prev != c.Index {
+			return false
+		}
+		if !c.StatusIsActive(burstKey) {
+			return false
+		}
+		c.explosiveFirework()
+		return false
+	}, "lyney-exit")
+}
```
k0l11, line 72: Canceling Q is not possible via Swap. It's only possible via E while Q is active, E should not do anything except cancel Q state if done during Q.

internal/characters/lyney/skill.go
```diff
@@ -0,0 +1,91 @@
...
+		Durability: 25,
+		Mult:       skill[c.TalentLvlSkill()] + skillBonus[c.TalentLvlSkill()]*float64(c.propSurplusStacks),
+	}
+	c.Core.Player.Heal(player.HealInfo{
+		Caller:  c.Index,
+		Target:  c.Core.Player.Active(),
+		Message: "Bewildering Lights",
+		Src:     0.2 * c.MaxHP() * float64(c.propSurplusStacks),
+		Bonus:   c.Stat(attributes.Heal),
+	})
```
k0l11, line 57: Should be moved inside the char task since this should also happen at hitmark

internal/characters/lyney/aimed.go
```diff
@@ -0,0 +1,228 @@
...
+		Durability:           25,
+		Mult:                 prop[c.TalentLvlAttack()],
+		HitWeakPoint:         weakspot == 1,
+		HitlagHaltFrames:     0.12 * 60,
+		HitlagFactor:         0.01,
+		HitlagOnHeadshotOnly: true,
+		IsDeployable:         true,
+	}
+	c.QueueCharTask(func() {
+		hpDrained := c.propSurplus()
```
k0l11, line 126: should drain right after the prop arrow snaps so it doesn't get the mh stack

internal/characters/lyney/cons.go
```diff
@@ -0,0 +1,161 @@
...
+// Additionally, Prop Arrows will summon 2 Grin-Malkin Hats and grant Lyney 1 extra stack of Prop Surplus.
+// This effect can occur once every 15s.
+func (c *char) c1() {
+	if c.Base.Cons < 1 {
+		return
+	}
+	c.maxHatCount = 2
+}
+
+func (c *char) c1StackIncrease() int {
```
imring, line 29: why is it called `c1StackIncrease`?

internal/characters/lyney/cons.go
```diff
@@ -0,0 +1,161 @@
...
+	addCount := 1
+	if  c.Base.Cons >= 1 && !c.StatusIsActive(c1ICDKey) {
+		addCount = 2
+	}
+	return addCount
+}
+
+func (c *char) c1HatIncrease() int {
```
imring, line 37: same above

internal/characters/lyney/burst.go
```diff
@@ -0,0 +1,121 @@
...
+	burstFrames[action.ActionAttack] = 297
+	burstFrames[action.ActionAim] = 297
+	burstFrames[action.ActionSkill] = 101
+	burstFrames[action.ActionDash] = 287
+	burstFrames[action.ActionJump] = 285
+	burstFrames[action.ActionSwap] = 283
+}
+
+// Burst attack damage queue generator
+func (c *char) Burst(p map[string]int) action.ActionInfo {
```
imring, line 36: without extra duration? idk if it's necessary

internal/characters/lyney/cons.go
k0l11: See the C1 text. At C1, every 15s the prop stack increase via prop arrow becomes 2 instead 1. This function determines that.

internal/characters/lyney/cons.go
k0l11: See the C1 text. At C1, every 15s the hat spawn via prop arrow becomes 2 instead 1. This function determines that.

internal/characters/lyney/burst.go
k0l11: Wdym by extra duration?

internal/characters/lyney/cons.go
imring: but there's not just c1 stack here.

internal/characters/lyney/burst.go
imring: oh i see canceling the burst using the skill

imring (approved):

PR #1644, opened 2023-08-21
fix hp bond weapons atk calc and active unit check

soloxcx:
From [Discord](https://discord.com/channels/845087716541595668/1143010182562062377/1143010182562062377):

Sample before commits: https://gcsim.app/sh/FWJWTC7CgTfk#sample=10236862637200687826&tab=sample

HP bonds were previously: 
1) activating on any ally skill (47f)
2) giving a flat atk buff lower than intended by a factor of 1000 (531f, atk gain should be 233 not 0.23)

### DIFF
--- a/internal/weapons/catalyst/flowingpurity/flowingpurity.go
+++ b/internal/weapons/catalyst/flowingpurity/flowingpurity.go
@@ -48,6 +48,9 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
 	bondDMGPCap := 0.09 + float64(r)*0.03
 
 	c.Events.Subscribe(event.OnSkill, func(args ...interface{}) bool {
+		if c.Player.Active() != char.Index {
+			return false
+		}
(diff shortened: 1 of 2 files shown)


k0l11 (dismissed):

internal/weapons/catalyst/flowingpurity/flowingpurity.go
```diff
@@ -65,7 +68,7 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p weapon.WeaponProfile
 
 		char.SetHPDebtByRatio(hp)
 		debt := char.CurrentHPDebt()
-		bondDMGP := (debt / 1000) * bondPercentage // use hp debt since you only get the buff after clearing bond anyway
+		bondDMGP := debt * bondPercentage // use hp debt since you only get the buff after clearing bond anyway
```
k0l11, line 71: This was correct, undo this change.

k0l11 (approved):

PR #1654, opened 2023-08-24
fix mh triggering on heal if hp already max

k0l11:
don't want to bloat on heal event sub with prev hp, so doing some floating point comparison


unleashurgeek (approved):
outside of the scope of this PR, but for cases like this I can see a parameter being valuable to enable/allow it on overheals since we do not simulate damage to heal

k0l11:
there are some sources of "damage" that we are already simming, example: hu tao e. also bloom explosion/burgeon soon which should cover most of the reliable damage sources

PR #1657, opened 2023-08-25
Only skip results parsing if there are no results

Charlie-Zheng:


soloxcx:
Context: This PR fixes `+params=[...]` being ignored as of release v.2.4.1

srliao (approved):

PR #1665, opened 2023-08-27
empty map

imring:
- [removing memory allocation for an empty map](https://github.com/genshinsim/gcsim/commit/4b61682998c759ce9dbc8ca806a1a85317a1c691);
- [add a required third parameter for execute_action](https://github.com/genshinsim/gcsim/commit/1ea6a34a8d141ef8f3e477f39ad09a07fdd0c3c9).

### DIFF
--- a/pkg/gcs/ast/parseCharacterAction.go
+++ b/pkg/gcs/ast/parseCharacterAction.go
@@ -59,9 +59,13 @@ Loop:
 			if err != nil {
 				return nil, err
 			}
-			if param != nil {
-				expr.Args = append(expr.Args, param)
+			if param == nil {
+				param = &MapExpr{
+					Pos:    n.pos,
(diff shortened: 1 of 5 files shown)


pkg/gcs/ast/parseCharacterAction.go
```diff
@@ -59,9 +59,13 @@ Loop:
 			if err != nil {
 				return nil, err
 			}
-			if param != nil {
-				expr.Args = append(expr.Args, param)
+			if param == nil {
+				param = &MapExpr{
+					Pos:    n.pos,
+					Fields: make(map[string]Expr),
```
srliao, line 65: remove the alloc here

srliao (changes requested):

srliao (approved):

PR #1686, opened 2023-09-02
new gcs evaluator - attempt at fixing performance issue introduced via gcsl rework

srliao:
This PR rewrites the gcs evaluator to remove the use of channel for blocking purposes (not truly a concurrent workload) which adds a lot of overhead as a result of context switching


srliao:
oh hold for now because it looks like this might actually be a performance decrease for wasm. needs more testing

also need to fix env scoping which is causing issues with closures not behaving properly

PR #1690, opened 2023-09-03
Add initial DB warning

soloxcx:
Adds a permanent (for now) disclaimer over the DB. It is simply placed below the ActionBar so user can scroll past it.

The intent is to clarify that the DB is not designed for robust, standardized comparisons. While some configs are written for this purpose, the user should exercise discretion on how they interpret the results, and they are freely able to open them, edit, and rerun.

Future updates: dismissible? translations?

The tanuki asset is official, from mhy <https://genshin-impact.fandom.com/wiki/HoYoLAB/Paimon%27s_Paintings> (Set 12), similar to the other assets.

Various screen sizes:
image
image
image


srliao:
Added hide/show and lined up warning with the db entries

image

image

srliao (approved):

PR #1700, opened 2023-09-10
Add back hurt syntax

### LINKED ISSUE #836: add hurt syntax from prehitlag

k0l11:
closes #836 

- no system function for generating hurt, tracked in #1703
- no stats collection or aggregation, tracked in #1701 

removed params in favor of only proccing via taking dmg:
- remove vourukashasglow stacks param
    - fix drain part triggering on non-external drain
- remove beaconofthereedsea damaged param
- remove alleyflash lambda param
    - fix drain part triggering if someone other than holder is drained
- adjust drain event sub active checks


k0l11:
will remove params that were crutches for this not existing

pkg/simulation/setup.go
```diff
@@ -343,3 +344,66 @@ func (s *Simulation) handleEnergy() {
...
+		ap.SkipTargets[targets.TargettableEnemy] = true
+		ap.SkipTargets[targets.TargettableGadget] = true
+
+		s.C.QueueAttack(ai, ap, -1, f) // -1 to avoid snapshot
+
+		s.C.Log.NewEventBuildMsg(glog.LogHurtEvent, -1, "hurt queued (once)").
+			Write("last", s.cfg.Hurt.LastHurt).
+			Write("cfg", s.cfg.Hurt).
+			Write("amt", amt).
+			Write("hurt_frame", s.C.F+f)
```
unleashurgeek, line 377: nit: should move to function since code between `every` and `once` is basically the same

pkg/gcs/ast/parseHurt.go
```diff
@@ -0,0 +1,158 @@
...
+			return parseRows, nil
+		default:
+			return nil, fmt.Errorf("ln%v: unrecognized token parsing hurt event: %v", n.line, n)
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nil, errors.New("unexpected end of line while parsing hurt event")
+}
```
unleashurgeek, line 120: nit: best to not have repeat code and simplify to one function and extend `interval` support being either a start and end or specific frame regardless of if `once` or `every`.

unleashurgeek (approved):

PR #1705, opened 2023-09-12
add go linting

### LINKED ISSUE #1226: Add linting to Go backend
To make code style consistent and avoid bad practices, we should introduce linting to our Go backend. I think it would be nice to run these ~~both as a pre-commit hook and~~ in CI (GitHub Actions).

VS Code related:
- It would also be nice if we can make it run on save for people using VS Code as their editor. In order to make this easy to adopt, we should create a `.vscode` folder with recommended extensions and settings. 
- Since VS Code has some [problems](https://github.com/Microsoft/vscode/issues/32693) related to this settings management, I propose only having a few top level folders in this repository like: `frontend` (gcsim.app), `backend` (all Go backend related files) and `docs` (docs.gcsim.app). After that we can place `.vscode` inside of `backend`. 
- In order for `.vscode` to be picked up, people using VS Code should open the `backend` folder only. Opening VS Code in a subfolder/higher level folder would make it unable to pick up the `.vscode` folder. This restriction should be communicated somewhere.

some relevant links:
- some talk about introducing golangci-lint: https://youtu.be/1h3c8r4Twlw?t=523
- https://github.com/golangci/golangci-lint
- https://golangci-lint.run/
  - https://golangci-lint.run/usage/integrations/#go-for-visual-studio-code
  - https://golangci-lint.run/usage/linters/
  - https://golangci-lint.run/usage/install#github-actions
- ~~https://bongnv.com/blog/2021-08-29-pre-commit-hooks-golang-projects/~~
- https://go.dev/doc/effective_go

k0l11:
closes #1226

- some linters are commented out (either impractical or highly controversial)
- exhaustruct should be enabled in a later PR
- includes significant fix for dori: after-sales rounds attacks never occurred before


.vscode/settings.json
```diff
@@ -0,0 +1,31 @@
...
+  "[javascript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+
```
unleashurgeek, line 16: May want to get rid of these since prettier is not configured atm in the ui side, but at the same time that can be a UI problem whenever someone makes a change next

unleashurgeek (approved):

.vscode/settings.json
k0l11: yeah will leave it for ui dev

unleashurgeek (dismissed):

unleashurgeek (dismissed):

unleashurgeek (approved):

PR #1709, opened 2023-09-17
Added sortby dps and create date options

ChinPTeo:
many changes because srl doesn't have prettier installed :smadge:


k0l11 (approved):
LGTM except for missing translation on the page in general ("Tags", "Date Added", "Sort By")

PR #1712, opened 2023-09-19
Changed db to open the config page instead of the results page

Charlie-Zheng:
I think that since the DB is primarily supposed to be used as a rotational database, and that we don't encourage people to blindly compare results of sims from the DB, it makes more sense for links to open to the configs first.


Charlie-Zheng:
Closing based on comments from Kyle

PR #1718, opened 2023-09-22
fix pipeline linting

srliao:


srliao:
redoing PR without yml snake case change

PR #1727, opened 2023-09-28
Neuvillette Implementation

Charlie-Zheng:
* [x]  New character package
* [x]  Config in character package
* [x]  Run pipeline with added config
* [x]  Character key
* [x]  Shortcuts for character key
* [x]  Add Character package to imports
* [x]  Character talent stats
* [x]  Character curve
* [x]  Normal Attack
* [x]  Charge Attack / Aimed Shot
* [x]  Skill
* [x]  Burst
* [x]  A1
* [x]  A4
* [x]  C0
* [x]  C1
* [x]  C2
* [x]  C3
* [x]  C4
* [x]  C5
* [x]  C6
* [x]  Other necessary talents (custom dash/jump, low/high plunge, ...)
* [x]  Hitlag
* [x]  ICD
* [x]  StrikeType
* [x]  Hitboxes
* [x]  Attack durability
* [x]  Particles
* [x]  Frames
* [x]  Update documentation
* [x] Update localization (IngameNames.json)


Charlie-Zheng:
The Spiritbreath Thorn currently scales on HP. It should scale on ATK instead.

Charlie-Zheng:
Status mod names for all A1 reactions seem to be "vaporize"

Charlie-Zheng:
Need to add filter to A1 event subscription to only do enemy reactions

Charlie-Zheng:
Also Neuv c6 is very very scuffed because it cannot pick up the droplets that c4 Neuv generates from healing

Charlie-Zheng:
I think neuv E CD doesn't start immediately, go check frame video to confirm

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,265 @@
...
+	endLag[action.ActionBurst] = 0
+	endLag[action.ActionSkill] = 0
+	endLag[action.ActionDash] = 0
+	endLag[action.ActionJump] = 0
+
+	earlyCancelEndLag = frames.InitAbilSlice(5000)
+	earlyCancelEndLag[action.ActionBurst] = 0
+	earlyCancelEndLag[action.ActionSkill] = 0
+	earlyCancelEndLag[action.ActionDash] = 0
+	earlyCancelEndLag[action.ActionJump] = 0
```
imring, line 52: nit: they're already 0

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,265 @@
...
+					return chargeJudgementStart + chargeJudgementDur + endLag[next]
+				},
+				AnimationLength: chargeJudgementStart + chargeJudgementDur + endLag[action.InvalidAction],
+				CanQueueAfter:   chargeJudgementStart + chargeJudgementDur + endLag[action.ActionDash],
+				State:           action.ChargeAttackState,
+			}, nil
+		}
+
+		return action.Info{}, fmt.Errorf("%v: Cannot execute %d CA Judgement Ticks. Max executed %d", c.CharWrapper.Base.Key, maxTicks, ticksDone)
+	}
```
imring, line 196: nit:
```go
	if p["ticks"] > 0 {
		// param for letting the user not do the full channel
		// calculate how long the judgement duration should be based on their tick count
		// additionally modify the frames so that only D/J/Q/E can follow. Otherwise sim errors on action
		// also need to verify it for c6
		maxTicks := p["ticks"]
		ticksDone := 0
		delay := getChargeJudgementHitmarkDelay(ticksDone)
		lag := endLag
		for delay < chargeJudgementDur {
			c.QueueCharTask(c.judgementWave, chargeJudgementStart+delay)
			ticksDone++
			delay = getChargeJudgementHitmarkDelay(ticksDone)

			if ticksDone == maxTicks {
				c.chargeEarlyCancelled = true
				lag = earlyCancelEndLag
				break
			}
		}
		if delay > chargeJudgementDur {
			delay = chargeJudgementDur
		}

		
		if maxTicks == ticksDone+1 {
			c.QueueCharTask(c.judgementWave, chargeJudgementStart+chargeJudgementDur)
		} else if maxTicks > ticksDone {
			return action.Info{}, fmt.Errorf("%v: Cannot execute %d CA Judgement Ticks. Max executed %d", c.CharWrapper.Base.Key, maxTicks, ticksDone)
		}

		// He drains 5 times in 3s, on frame 40, 70, 100, 130, 160
		for d := 40; d < delay; d += 30 {
			c.QueueCharTask(c.consumeHp, chargeJudgementStart+d)
		}
		return action.Info{
			Frames: func(next action.Action) int {
				return chargeJudgementStart + delay + lag[next]
			},
			AnimationLength: chargeJudgementStart + delay + lag[action.InvalidAction],
			CanQueueAfter:   chargeJudgementStart + delay + lag[action.ActionDash],
			State:           action.ChargeAttackState,
		}, nil
	}
```

internal/characters/neuvillette/skill.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		for i := 0; i < 3; i++ {
+			// TODO: find the actual sourcewater droplet spawn radius for Neuv E
+			pos := geometry.CalcRandomPointFromCenter(ac.Target.Pos(), circ.Radius()+1.0, circ.Radius()+4.0, c.Core.Rand)
+			common.NewSourcewaterDropletNeuv(c.Core, pos)
+		}
+		c.Core.Combat.Log.NewEvent("Spawned 3 droplets", glog.LogCharacterEvent, c.Index)
+	}
+
+	c.Core.QueueParticle(c.Base.Key.String(), 4, attributes.Hydro, c.ParticleDelay)
```
imring, line 97: add icd for particle generation (0.3sec i think)

internal/characters/neuvillette/skill.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return action.Info{
+		Frames:          frames.NewAbilFunc(skillFrames),
+		AnimationLength: skillFrames[action.InvalidAction],
+		CanQueueAfter:   skillFrames[action.ActionSwap], // earliest cancel
+		State:           action.SkillState,
+	}, nil
+}
+
+func (c *char) skillcb(ac combat.AttackCB) {
```
imring, line 81: add a condition if the enemy gets hit:
```go
	if ac.Target.Type() != targets.TargettableEnemy {
		return
	}
```

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (c *char) c4() {
+	c.Core.Events.Subscribe(event.OnHeal, func(args ...interface{}) bool {
+		target := args[1].(int)
+		if c.Core.Player.ActiveChar().Index == c.Index && c.Index == target {
+			// TODO: find the actual sourcewater droplet spawn shape for Neuv C4
+			player := c.Core.Combat.Player()
+			center := player.Pos().Add(player.Direction().Normalize().Mul(geometry.Point{X: 3.0, Y: 3.0}))
+			pos := geometry.CalcRandomPointFromCenter(center, 0, 2.5, c.Core.Rand)
+			common.NewSourcewaterDropletHydroTrav(c.Core, pos)
```
imring, line 56: NewSourcewaterDropletNeuv*, but see the issue above

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil, false
+		},
+	})
+}
+
+func (c *char) c4() {
+	c.Core.Events.Subscribe(event.OnHeal, func(args ...interface{}) bool {
+		target := args[1].(int)
+		if c.Core.Player.ActiveChar().Index == c.Index && c.Index == target {
+			// TODO: find the actual sourcewater droplet spawn shape for Neuv C4
```
imring, line 52: use `c.Core.Player.Active() == c.Index`

internal/characters/traveler/common/hydro/travelerhydro.go
```diff
@@ -36,15 +36,25 @@ func NewChar(gender int) core.NewCharacterFunc {
 }
 
 func (c *char) Init() error {
-	c.droplets = make([]*common.SourcewaterDroplet, 0)
-
 	return nil
 }
 
+func (c *char) getSourcewaterDroplets() []combat.Gadget {
```
imring, line 42: the function returns the number of droplets created by traveler. add the condition `g.GadgetTyp() == combat.GadgetTypSourcewaterDropletHydroTrav`.

ui/packages/docs/src/components/Fields/character_data.json
```diff
@@ -172,7 +172,15 @@
       "fields": [
         ".travelerhydro.droplets"
       ],
-      "desc": "Number of Sourcewater droplets created by Traveler."
+      "desc": "Number of Sourcewater droplets on the field."
```
imring, line 175: why?

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,265 @@
...
+	chargeFrames[action.ActionSwap] = 58
+
+	endLag = frames.InitAbilSlice(51)
+	endLag[action.ActionWalk] = 36
+	endLag[action.ActionCharge] = 30
+	endLag[action.ActionSwap] = 27
+	endLag[action.ActionBurst] = 0
+	endLag[action.ActionSkill] = 0
+	endLag[action.ActionDash] = 0
+	endLag[action.ActionJump] = 0
```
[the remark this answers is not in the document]
Charlie-Zheng: They aren't. That's for InitNormalCancelSlice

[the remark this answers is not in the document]
imring: oh i see. reverted the changes

internal/characters/traveler/common/hydro/travelerhydro.go
Charlie-Zheng: I modified the documentation to instead say "number of sourcewater droplets on the field"

ui/packages/docs/src/components/Fields/character_data.json
Charlie-Zheng: I don't quite understand the point of differentiating between those created by HydroMC vs anyone else. Besides the unshared 12 gadget limit, they behave identically

imring: i initially created it for testing a1, but for some reason i decided to keep it.
if you need the number of sourcewater droplets, create a field `.gadgets.sourcewater.count` (see dendrocore implementation)

Charlie-Zheng: Should I remove this param from the HMC then? And still keep the one from Neuvillette, since that one uses Neuvillettes position to check which ones are in range?

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,122 @@
...
+			return
+		}
+
+		if c.chargeJudgeStartF+c.chargeJudgeDur-c.Core.F <= 60 {
+			droplets := c.getSourcewaterDroplets()
+
+			// c6 only absorbs one droplet at a time
+			if len(droplets) > 0 {
+				c.consumeDroplet(droplets[c.Core.Combat.Rand.Intn(len(droplets))])
+				c.chargeJudgeDur += 60
```
k0l11, line 87: Please include a log under the character event category indicating that this happened which has chargeJudgeDur before and after extension.

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+
+		ticks, ok := p["ticks"]
+		if !ok {
+			ticks = -1
+		} else if ticks < 0 {
+			ticks = 0
+		}
+
+		c.QueueCharTask(c.chargeJudgementTick(c.chargeJudgeStartF, 0, ticks, false), chargeLegalEvalLeft+getChargeJudgementHitmarkDelay(0))
```
k0l11, line 130: change this to:
```
// cannot use hitlag affected queue because the logic just does not work then
// -> can't account for possible hitlag delaying the update of the anim length (sim moves on to next action, but ticks continue)
// start counting at 1 for correct number of ticks when supplying ticks param
c.Core.Tasks.Add(c.chargeJudgementTick(c.chargeJudgeStartF, 1, ticks, false), chargeLegalEvalLeft+getChargeJudgementHitmarkDelay(0))
```

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,289 @@
...
+			c.nextTickTime = nextF - c.chargeJudgeStartF
+
+			if nextF > c.chargeJudgeStartF+c.chargeJudgeDur {
+				c.QueueCharTask(c.chargeJudgementTick(src, tick+1, maxTick, true), c.chargeJudgeStartF+c.chargeJudgeDur-c.Core.F)
+				c.nextTickTime = c.chargeJudgeDur
+				c.nextTickTime2 = nextF - c.chargeJudgeStartF
+			}
+		}
+	}
+}
```
k0l11, line 254: Please change the implementation to:
```
func (c *char) chargeJudgementTick(src, tick, maxTick int, last bool) func() {
	return func() {
		if c.chargeJudgeStartF != src {
			return
		}
		// no longer in CA anim -> no tick
		if c.Core.F > c.chargeJudgeStartF+c.chargeJudgeDur {
			return
		}

		// last tick -> check for C6 extension
		if last {
			// C6 did not extend CA -> proc wave and stop queuing ticks
			if c.Core.F == c.chargeJudgeStartF+c.chargeJudgeDur {
				c.judgementWave()
			} else {
				// C6 extended the CA between when this tick was queued and when this tick was executed
				// -> allow the other non last queued task to execute by extend anim length to include that task
				c.tickAnimLength = c.tickAnimLengthC6Extend
			}
			return
		}

		// tick param supplied and hit the limit -> proc wave, enable early cancel flag for next action check and stop queuing ticks
		if tick == maxTick {
			c.judgementWave()
			c.chargeEarlyCancelled = true
			return
		}

		c.judgementWave()

		// next tick handling
		if maxTick == -1 || tick < maxTick {
			tickDelay := getChargeJudgementHitmarkDelay(tick + 1)
			// calc new animation length to be up until next tick happens
			nextTickAnimLength := c.Core.F - c.chargeJudgeStartF + tickDelay

			// always queue up non-final tick that will be executed in case C6 was proc'd
			c.Core.Tasks.Add(c.chargeJudgementTick(src, tick+1, maxTick, false), tickDelay)

			// queue up last tick if next tick would happen after CA duration ends
			if nextTickAnimLength > c.chargeJudgeDur {
				// queue up final tick to happen at end of CA duration
				c.Core.Tasks.Add(c.chargeJudgementTick(src, tick+1, maxTick, true), c.chargeJudgeDur-c.tickAnimLength)
				// update tickAnimLength to be equal to entire CA duration at the end
				c.tickAnimLength = c.chargeJudgeDur
				// if C6 is triggered, then tickAnimLength will be wrong so this var holds the actual tickAnimLength if ticks continued normally beyond the original final tick
				c.tickAnimLengthC6Extend = nextTickAnimLength
			} else {
				// next tick happens within CA duration -> update tickAnimLength as usual
				c.tickAnimLength = nextTickAnimLength
			}
		}
	}
}
```
This + the suggested change for the very first queuing of the task should fix the bug when ticks param < 8 the actual ticks are 1 more than supplied.
Includes:
- nextTickTime -> tickAnimLength
- nextTimeTime2 -> tickAnimLengthC6Extend
- bunch of comments
- less frame arithmetic / simpler conditions for the next tick part at the bottom
- no char queue because skill aligned hitlag leads to desync (see change suggested for very tick for reasoning)

I tested it out a bit (non-c6 and ticks < 8, non-c6 and no ticks param, c6 and ticks <8, c6 and no ticks param, c6 and ticks param > 8) and it seems to work well.

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,122 @@
...
+		if c.Index != target {
+			return false
+		}
+		if c.Core.Player.Active() == c.Index && c.Index == target {
+			// TODO: find the actual sourcewater droplet spawn shape for Neuv C4
+			player := c.Core.Combat.Player()
+			center := player.Pos().Add(player.Direction().Normalize().Mul(geometry.Point{X: 3.0, Y: 3.0}))
+			pos := geometry.CalcRandomPointFromCenter(center, 0, 2.5, c.Core.Rand)
+			common.NewSourcewaterDroplet(c.Core, pos, combat.GadgetTypSourcewaterDropletNeuv)
+			c.lastc4 = c.Core.F
```
k0l11, line 69: Similarly to C6, this could also use a character event log indicating that this proc'd.

ui/packages/docs/src/components/AoE/character_data.json
```diff
@@ -4422,6 +4422,83 @@
...
+        "ability": "Equitable-Judgement",
+        "shape": "Box",
+        "center": "PrimaryTarget",
+        "boxX": 3,
+        "boxY": 8
+      },
+      {
+        "ability": "CA",
+        "shape": "Box",
+        "center": "PrimaryTarget",
```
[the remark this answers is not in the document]
Charlie-Zheng: For this, does it apply to the Equitable Judgement as well as the CA? (Equitable Judgement is the charged CA)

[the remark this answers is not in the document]
k0l11: Yes

internal/characters/neuvillette/charge.go
Charlie-Zheng: I additionally updated the `getChargeJudgementHitmarkDelay` function to match the 1 indexed tick counting, and correspondingly updated the initial call to use getDelay(1)

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,289 @@
...
+		Frames:          func(next action.Action) int { return windup + chargeFrames[next] },
+		AnimationLength: windup + chargeFrames[action.InvalidAction],
+		CanQueueAfter:   windup + chargeFrames[action.ActionDash],
+		State:           action.ChargeAttackState,
+	}, nil
+}
+
+func (c *char) judgementWave() {
+	// calculated every hit since canqueueafter is after the first tick, so configs can change the primary target/entity positions while the CA happens
+	ap := combat.NewBoxHit(c.Core.Combat.Player(), c.Core.Combat.PrimaryTarget(), geometry.Point{}, 3.5, 15)
```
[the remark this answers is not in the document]
Charlie-Zheng: I also updated the hitbox calculation for the uncharged CA to use NewBoxHitOnTarget

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,305 @@
...
+				Abil:       "Charge Attack",
+				AttackTag:  attacks.AttackTagExtra,
+				ICDTag:     attacks.ICDTagNone,
+				ICDGroup:   attacks.ICDGroupDefault,
+				StrikeType: attacks.StrikeTypeDefault,
+				Element:    attributes.Hydro,
+				Durability: 25,
+				Mult:       charge[c.TalentLvlAttack()],
+			}
+			ap := combat.NewBoxHitOnTarget(c.Core.Combat.Player(), c.Core.Combat.PrimaryTarget(), 3, 8)
```
k0l11, line 174: offset should be nil, not PrimaryTarget

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,305 @@
...
+		Frames:          func(next action.Action) int { return windup + chargeFrames[next] },
+		AnimationLength: windup + chargeFrames[action.InvalidAction],
+		CanQueueAfter:   windup + chargeFrames[action.ActionDash],
+		State:           action.ChargeAttackState,
+	}, nil
+}
+
+func (c *char) judgementWave() {
+	// calculated every hit since canqueueafter is after the first tick, so configs can change the primary target/entity positions while the CA happens
+	ap := combat.NewBoxHitOnTarget(c.Core.Combat.Player(), c.Core.Combat.PrimaryTarget(), 3.5, 15)
```
k0l11, line 195: offset should be nil, not PrimaryTarget

internal/characters/neuvillette/cons.go
k0l11: Please prepend "C4: " just like you did for the C6 log.

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,132 @@
...
+			c.Core.Combat.Log.NewEvent("Spawned 1 droplet", glog.LogCharacterEvent, c.Index).
```
k0l11, line 71: Please prepend "C4: " just like you did for the C6 log.

ui/packages/ui/src/Translation/locales/IngameNames.json
```diff
@@ -119,6 +119,7 @@
       "lyney": "Lyney",
       "lynette": "Lynette",
       "freminet": "Freminet",
+      "neuvillette": "Neuvillette",
```
k0l11, line 122: Also add the other translations please (see wiki).

k0l11:
also missing N0 timing

internal/characters/neuvillette/neuvillette.go
```diff
@@ -0,0 +1,109 @@
...
+		droplet, ok := g.(*common.SourcewaterDroplet)
+		if !ok {
+			continue
+		}
+		if droplet.Pos().Distance(playerPos) <= 15 {
+			droplets = append(droplets, droplet)
+		}
+	}
+	return droplets
+}
```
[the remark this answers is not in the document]
Charlie-Zheng: How do we account for the camera not always the direction that the player faces? and/or being able to turn the camera while charging?

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,122 @@
...
+
+		if c.chargeJudgeStartF+c.chargeJudgeDur-c.Core.F <= 60 {
+			droplets := c.getSourcewaterDroplets()
+
+			// c6 only absorbs one droplet at a time
+			if len(droplets) > 0 {
+				c.consumeDroplet(droplets[c.Core.Combat.Rand.Intn(len(droplets))])
+				c.chargeJudgeDur += 60
+			}
+		}
```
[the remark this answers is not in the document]
Charlie-Zheng: Is that 60f CD from datamine? Visually when watching a c6 neuv do CAs, he only absorbs a droplet once the CA "circle" drops below the 1s remaining threshold. So for the first 2s, he doesn't absorb droplets.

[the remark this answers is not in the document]
k0l11: Oops, quickly glanced over this and thought this was another cooldown calc.

internal/characters/neuvillette/neuvillette.go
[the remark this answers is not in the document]
k0l11: > How do we account for the camera not always the direction that the player faces?
Assumption here is that the camera in CA = player direction for simplicity.
> Is the expected method that the config write will set the direction of the player before using CA?
No, it would use the player's direction. Adjusting would be done via teleporting the player (which updates the player's direction automatically). For example if you have 2 targets on the x-axis then you'd also want to move the player on the x-axis to hit both.
> And for being able to turn the camera while charging, we just assume that the player isn't turning the camera (no high DPI mouse shenanigans)?
I would not support that. Same thing can be achieved by stacking enemies on the same spot. We don't support missing CA dmg ticks by doing that anyways.

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,122 @@
...
+		// C6 projectile stops on first target hit, with 0.5 rad sphere hitbox.
+		// Because we don't simulate the projectile, it's just a circle hit
+		ap := combat.NewCircleHitOnTarget(c.Core.Combat.PrimaryTarget(), geometry.Point{}, 0.5)
+		// it looks like the c6 has 29 frames of delay but I didn't count it rigourously
+		c.Core.QueueAttack(ai, ap, 29, 29)
+		c.Core.QueueAttack(ai, ap, 29, 29)
+		c.QueueCharTask(c.c6(src), 120)
+	}
+}
```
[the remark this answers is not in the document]
Charlie-Zheng: I don't think the C6 hits are tied to CA hits, since the CA hits are every 25f, which puts the "actual" time between C6 hits to 25*5=125f instead of 120f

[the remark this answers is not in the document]
k0l11: But it is tied to the hits.

> Additionally, when Equitable Judgment hits opponents, it will fire off 2 additional currents [...]

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,145 @@
...
+		if c.Core.Player.Active() != c.Index {
+			return false
+		}
+		if c.Index != target {
+			return false
+		}
+		if c.StatusIsActive(c4ICDKey) {
+			return false
+		}
+		if c.Core.Player.Active() == c.Index && c.Index == target {
```
k0l11, line 65: remove the wrapping if-statement, condition is always true

internal/characters/neuvillette/neuvillette.go
```diff
@@ -0,0 +1,139 @@
...
+		droplets = append(droplets, droplet)
+	}
+
+	return droplets
+}
+
+func (c *char) getSourcewaterDropletsC6() []*common.SourcewaterDroplet {
+	player := c.Core.Combat.Player()
+
+	// TODO: this is an approximation based on an ongoing KQM ticket (faster-neuvi-balls)
```
k0l11, line 111: pls remove this, we are sure of the area for c6

internal/characters/neuvillette/cons.go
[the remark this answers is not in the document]
Charlie-Zheng: Hmmm but I think it's not tied to the actual damage hit, but rather the gadget "hitting" the enemy

internal/characters/neuvillette/skill.go
```diff
@@ -0,0 +1,152 @@
...
+		ActorIndex:         c.Index,
+		Abil:               "Spiritbreath Thorn (" + c.Base.Key.Pretty() + ")",
+		AttackTag:          attacks.AttackTagElementalArt,
+		ICDTag:             attacks.ICDTagNone,
+		ICDGroup:           attacks.ICDGroupDefault,
+		StrikeType:         attacks.StrikeTypeSpear,
+		Element:            attributes.Hydro,
+		Durability:         0,
+		Mult:               thorn[c.TalentLvlSkill()],
+		CanBeDefenseHalted: true,
```
k0l11, line 76: add HitlagFactor:     0.01 above this and adjust docs

internal/characters/neuvillette/cons.go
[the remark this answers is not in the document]
k0l11: It is literally tied to the damage, I checked dm

internal/characters/neuvillette/charge.go
```diff
@@ -0,0 +1,299 @@
...
+			StrikeType: attacks.StrikeTypePierce,
+			Element:    attributes.Hydro,
+			Durability: 25,
+			FlatDmg:    chargeJudgement[c.TalentLvlAttack()] * c.MaxHP(),
+		}
+
+		c.chargeJudgeStartF = c.Core.F + chargeLegalEvalLeft
+		c.chargeJudgeDur = 173
+
+		c.QueueCharTask(c.c6DropletCheck(c.chargeJudgeStartF), chargeLegalEvalLeft)
```
k0l11, line 110: now this will be queued even at < C6

internal/characters/neuvillette/cons.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func (c *char) c6cb(a combat.AttackCB) {
+	if a.Target.Type() != targets.TargettableEnemy {
+		return
+	}
+	if c.StatusIsActive(c6ICDKey) {
+		return
+	}
```
k0l11, line 126: missing c.AddStatus(c6ICDKey, 2*60, true)

k0l11 (approved):

PR #1748, opened 2023-11-11
Fix Fischl A4 Timing

RarePossum:
Fixes Fischl A4 timing. This should cause it to proc after TKP as seen in game. The frame count matches personal testing that I seem to have lost the footage of.


RarePossum:
I have not seen evidence to the contrary, and ti seem to be variable, don't merge yet

k0l11:
A better fix would be:
TKP: 4f -> 3f
A4: 3f -> 4f

PR #1771, opened 2023-12-16
Add song of days past

Siwat-Keng:
- [x] New artifact package
- [x] Run pipeline with added config
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation
- [x] Update localization (IngameNames.json)


internal/artifacts/songofdayspast/songofdayspast.go
```diff
@@ -0,0 +1,119 @@
...
+		default:
+			return false
+		}
+		char := s.core.Player.ByIndex(atk.Info.ActorIndex)
+		if s.core.Status.Duration(wavesOfDaysPastKey) == 0 {
+			return false
+		}
+		if char.Index != s.core.Player.Active() {
+			return false
+		}
```
k0l11, line 108: this can be just
```
if atk.Info.ActorIndex != s.core.Player.Active() {
    return false
}
```
no need for the `char := ...` above

internal/artifacts/songofdayspast/songofdayspast.go
```diff
@@ -0,0 +1,119 @@
...
+		s.healStacks += healAmt
+		if s.healStacks >= 15000 {
+			s.healStacks = 15000
+		}
+		if s.core.Status.Duration(yearningKey) == 0 {
+			s.core.Status.Add(yearningKey, 6*60)
+			s.core.Tasks.Add(func() {
+				s.core.Status.Add(wavesOfDaysPastKey, 10*60)
+				s.core.Flags.Custom[wavesOfDaysPastKey] = 5
```
k0l11, line 82: since another yearning instance can already start while waves effect is still up, the healStacks/collected healing needs to "snapshot" into another set var and then set healStacks to 0 in here

internal/artifacts/songofdayspast/songofdayspast.go
```diff
@@ -0,0 +1,119 @@
...
+		if char.Index != s.core.Player.Active() {
+			return false
+		}
+		if s.core.Flags.Custom[wavesOfDaysPastKey] > 0 {
+			s.core.Flags.Custom[wavesOfDaysPastKey]--
+			amt := s.healStacks * 0.08
+			atk.Info.FlatDmg += amt
+		}
+		if s.core.Flags.Custom[wavesOfDaysPastKey] == 0 {
+			s.healStacks = 0
```
k0l11, line 115: should be changed to
```
c.Core.Status.Delete(wavesOfDaysPastKey)
```
(following the previous comment there is no need to reset the collected healing nor snapshot healing here because the snapshot healing will be overwritten by the next add of the waves effect anyways)

PR #1778, opened 2023-12-16
fix hutao c6 check not starting at t=0s

### LINKED ISSUE #1734: Hu Tao C6 is checking even when she's off field
C6 checks every 2s but skips the first check at the start of the sim

See 120f https://gcsim.app/sh/PcPKw97bbdCN

k0l11:
closes #1734
https://discord.com/channels/845087716541595668/1167692515017760788

- first check was at t=2s before

### DIFF
--- a/internal/characters/hutao/cons.go
+++ b/internal/characters/hutao/cons.go
@@ -25,10 +25,8 @@ func (c *char) c6() {
 		c.checkc6(false)
 		return false
 	}, "hutao-c6")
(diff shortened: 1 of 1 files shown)


MGolub101:
This is kinda late but Hu Tao's c6 only activates when she takes damage. I.e. Casting her skill or standing in fire. She can sit at 1 hp forever and never activate it until one of those conditions are met.

PR #1794, opened 2023-12-17
remove doc on hidden system func

srliao:


srliao:
closed in favour of #1795

PR #1800, opened 2023-12-17
add back dps on embed

srliao:
adding back dps on embed based on discussion here: https://discord.com/channels/845087716541595668/983391844631212112/1185622407772381254

code is a bit hacky to be maybe fixed if issues comes up


srliao:
Looks like this:
image

dps will not show if there are errors (i.e. dev build or modified/dirty build)

PR #1826, opened 2023-12-22
Added a mode to ignore burst energy

Charlie-Zheng:
Added a mode so that energy is ignored when attempting to burst.


k0l11:
Like srl said here <https://discord.com/channels/845087716541595668/983391844631212112/1187633245920170054>, 2 things and then this can be merged:
- update docs over at `ui/packages/docs/docs/reference/config.md` to include this new flag in the Valid options section
- add a new warning to `pkg/agg/warnings.go` which is a bool that is set if the config has the new mode enabled and regen protos
- inside of `ui/packages/ui/src/Pages/Viewer/Components/Warnings.tsx`:
 - add a new component called `IgnoreEnergyWarning` to the bottom of the file (look at target overlap warning for inspiration)
 - make sure that the warning's intent is set to `Intent.DANGER` to make it red
 - add the new component to the warnings array towards the top of the file

k0l11:
reminder to self to add highlighting for new option in prism

PR #1854, opened 2023-12-30
add optional param to specify how many shard to try and pick up

srliao:


k0l11:
docs

k0l11:
denied, put that in a loop

PR #1869, opened 2024-01-03
Additional Energy Information in Results

Charlie-Zheng:
Added weighted er and flat/particle energy per burst to results.

This can be used with ignore_burst_energy=true to generate ER req graphs or be used in the substat optimizer to determine ER needs easier.


srliao:
My immediate thought is that this should be a separate collector, and not part of the energy collector, that is only enabled if we are ignoring burst cost. The OnBurst cb and the ER calc is expensive from when they are not needed

k0l11:
closing in favor of #1890

PR #1893, opened 2024-01-11
Bugfixes for sample logs

shizukayuki:
* Add highlight to animation and `mod extended` logs
* Fix mod logs to use hitlagged expiry
* Move `mod extended` logs to char column
* Remove event name from cooldown logs `cooldown: skill cooldown triggered` -> `skill cooldown triggered`


ui/packages/ui/src/Pages/Sample/Components/parsev2.ts
```diff
@@ -211,21 +211,16 @@ export function parseLogV2(
 
         e.icon = "queue";
         break;
+      case "cooldown":
```
k0l11, line 214: Is the stuff here supposed to be commented out?

PR #1900, opened 2024-01-16
Fix dehya

Aftermathrar:
visible changes:
- attack:
  - **fix normal attacks resetting the boolean for whether skill had been reused**
  - **fix dehya n1 to n4 and reposition lv.4 multipliers**
  - fix dehya n1 to n4 hitlag and small attack hitmark/action frame adjustments
- skill:
  - **implement skill damage mitigation and self DoT**
  - **fix skill cast/recast frames**
  - **fix skill damage DoT having icd (should have no icd)**
  - **fix field proc delay (10f -> 2f after dmg)**
  - **fix field proc hitlag not being deployable**
  - **fix field triggering on final burst hit damage**
  - **fix burst casting skill if starting with skill off cooldown**
  - **fix redeployed field not using saved DoT damage cooldown**
  - fix sac field pickup timing to happen right before redeploy instead of on action start
- burst:
  - **add dash -> jump interaction during burst leading to kick**
  - **fix first punch, kick, slow punch frames**
  - **fix burst interaction with attack/skill/jump during punches**
  - fix burst not being hitlag extendable (doesn't really matter right now because burst punch/kick do not have hitlag)
  - fix punch/kick state for iframe semantics 
  - min 6 punches idle, max 10 punches <C6, max 15 punches C6 applies now
- ascension:
  - **implement a1**
  - **implement a4**
- constellations:
  - **implement c2 dmg buff**
  - **fix c2 not extending field correctly**
  - **fix c2 extending initial cast**
  - fix c4 icd not being hitlag extended
  - adjust c2 flag reset to be a bit more future-proof

dev related changes:
- add new function to remove field since it is used in several scenarios
- skill recast function now uses character skill tracking variable for skill duration
- rename boolean for skill reuse to make more sense 
- rename remove field to pick up field
- move asc/cons checks into the functions
- move c2 to its own event sub
- remove old burst cast tracking variable
- remove old burstCast tracking variable
- replace burstFrames with constant
- refactor jump canceling burst punches
- refactor dehya c1, c2, c4, c6
- cannot cancel before first punch so simplified burst func
- remove unnecessary hitlag info for non-hitlag attacks in skill


k0l11:
reviewing later, will make release without this

PR #1914, opened 2024-01-20
add cumulative damage statistic and graph in web

k0l11:
- shown as a new card right below damage timeline
- from a quick check, run time doesn't seem to slow down much? (reviewer pls check)
- doing some target index <-> target key mapping that is very fragile (pls check)

example images:
image
image
image
image


protos/model/result.proto
```diff
@@ -169,6 +170,25 @@ message CharacterBuckets {
...
 
+message TargetBucketStats {
+  uint32 bucket_size = 1 [json_name="bucket_size"];
+  map<int32, TargetBuckets> targets = 2 [json_name="targets"];
+}
+
+message TargetBuckets {
+  TargetBucket overall = 1 [json_name="overall"];
+  TargetBucket target = 2 [json_name="target"];
+}
```
unleashurgeek, line 181: nit: may be cleaner if we can move `overall` to `TargetBucketStats` and do the cross-target sum in agg flush rather than deferring it to the UI

ui/packages/ui/src/Pages/Viewer/Components/Damage/CumulativeDamageCard/CumulativeDamageData.tsx
```diff
@@ -0,0 +1,131 @@
...
+  target: string,
+  input?: TargetBucketStats
+): OverTimeData {
+  return useMemo(() => {
+    if (input?.bucket_size == null || input.targets == null) {
+      return { data: [], duration: 1, maxValue: 1 };
+    }
+    let targetData: TargetBucket | undefined;
+    if (graph === "overall") {
+      console.log(input.targets);
```
unleashurgeek, line 51: nit: remove log

PR #1917, opened 2024-01-21
fix cumulative damage and metadata sorting

k0l11:
- cmp.Compare in cumu dmg and Less func in metadata did not do floating point equality via a delta making it unstable
- use slices.SortStableFunc for sorting
- also fix edge case of iterations = 1 not populating q1/q3 percentile
- move common logic as generic funcs to util package

### DIFF
--- a/pkg/agg/util.go
+++ b/pkg/agg/util.go
@@ -104,3 +104,32 @@ func (h *LinearHist) BinToValue(bin float64) float64 {
 func Ptr[T any](v T) *T {
 	return &v
 }
+
+// util for metadata and damage agg
+
+const FloatEqDelta = 0.00001
+
+// given a pre-sorted slice of values, returns the indexes of the percentiles
+func GetPercentileIndexes[T any](a []T) (int, int) {
+	l := len(a)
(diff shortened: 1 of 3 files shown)


pkg/agg/util.go
```diff
@@ -104,3 +104,32 @@ func (h *LinearHist) BinToValue(bin float64) float64 {
...
+	return (l - 1) / 2, (l-1)/2 + 1
+}
+
+// given a pre-sorted slice of values, returns the median element
+func Median[T any](a []T) T {
+	l := len(a)
+
+	if l == 0 {
+		var empty T
+		return empty
```
unleashurgeek, line 130: nit: can take empty case in as a param (thinking about how for seed selection this returning 0 DPS instead of -1 will be a different behavior, but I do not remember if that matters for UI)

unleashurgeek (approved):

pkg/agg/util.go
k0l11: does not matter for ui and if empty case customization is needed, can change signature later

PR #1920, opened 2024-01-22
add param for extending dash length

### LINKED ISSUE #1625: Add support for extending dash length
Generic character dash should take `f` as a param that gets added on-top of the base dash frames. Custom dash implementations already accept this param.

k0l11:
closes #1625
- `dash[extend=2]` to extend dash by 2f for a character


k0l11:
abandoning because the continuous stamina consumption is a pain

PR #1930, opened 2024-01-23
fix recovery buttons not working

k0l11:
- for shared sims / when running from simulator tab
- do not show buttons if no recoveryConfig available

### DIFF
--- a/ui/packages/ui/src/Stores/viewerSlice.ts
+++ b/ui/packages/ui/src/Stores/viewerSlice.ts
@@ -4,12 +4,14 @@ import { SimResults } from "@gcsim/types";
 export interface Viewer {
   data: SimResults | null;
   hash: string | null;
+  recoveryConfig: string | null;
   error: string | null;
 }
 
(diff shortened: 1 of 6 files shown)


ui/packages/ui/src/Stores/viewerSlice.ts
```diff
@@ -27,7 +29,8 @@ export const viewerSlice = createSlice({
       state.error = null;
       return state;
     },
-    setError: (state, action: PayloadAction<{ error: string }>) => {
+    setError: (state, action: PayloadAction<{ recoveryConfig: string | null; error: string }>) => {
+      state.recoveryConfig = action.payload.recoveryConfig;
```
unleashurgeek, line 33: In `start`, should set `recoveryConfig` to `null` to prevent the chance of a prior config bleeding over to a new run (case of error happening but for some reason `recoveryConfig` is not updated)

unleashurgeek (approved):

PR #1932, opened 2024-01-24
add fully localized search for gcsim char/artifact/weapon keys

### LINKED ISSUE #1847: add UI for "dictionary" to search gcsim keys from text input
Non-english users want to find character/weapon/artifact keys to put into their config using their native language.
Example: Chinese user enters 娜维娅 into a search bar and can copy "navia" to their clipboard.

k0l11:
closes #1847 
- on click it will copy the english key to clipboard and show a msg indicating as such on the top right
- fix artifact select not using pipeline data
- fix selects not using localization for the placeholder

example images:
- English
image
image
image
image
- Japanese
image
image
image
image


k0l11:
new omnibar width so that it fits on mobile
image

PR #1940, opened 2024-01-29
localize stat array, viewer page warnings, result tab (mostly), ... 

k0l11:
- rename dendro (fuel) in sim and ui to dendro-fuel for consistency
- use same error formatting for runtime error as for config error
- split DataColors into DataColorsConst and a useDataColors hook which uses translated keys
- remove console.log from landing page db entry/release text load
- adjust embed to make sure it's always using en localization
- adjust warnings to be more streamlined in terms of phrasing
- adjust graphs to be more readable on mobile by adjusting margin (especially Source ... and Total Energy)
- adjust tooltips to use max-content instead of min-content
- adjust db search to also consider gcsim keys in search (not just the localized name, same behaviour as name search on sim page)
- add action and stat name search on sim page

example images for result tab:
- English
3EswFIFbui
- Japanese
JWKK3nD5Xt


ui/packages/embed/src/Layout/index.tsx
```diff
@@ -11,6 +12,8 @@ type Props = {
 const Layout = ({ data }: Props) => {
   const [ready, setReady] = useState(false);
   const [loaded, setLoaded] = useState(0);
+  const { i18n } = useTranslation();
+  i18n.changeLanguage("en"); // make sure that embed is definitely using English localization, might not be necessary?
```
unleashurgeek, line 16: one thing we could do is add a language property to the share url (this was my original plan). IE: https://gcsim.app/sh/123?lang=ja would force the site to be in jp (there is a built-in one for this already but don't remember what it should be) and then the embed request can also pick up the language and generate accordingly to allow for localized embeds.

unleashurgeek (approved):

PR #772, opened 2022-08-16
bow ca feature update

k0l11:
closes #174 

sheet: https://docs.google.com/spreadsheets/d/187T-SngEZUUordjY_K_tF_DdvHjQju9CoBJdp2eJOis/edit?usp=sharing

main points:
- implements phys ca and arcc in all bow aimed.go files
- implements a hold param in all bow aimed.go files:
    - hold = 0 means ARCC
    - hold = 1 means Aimed Shot (phys ca)
    - hold = 2 means Fully-Charged Aimed Shot (element ca) or Fully-Charged Aimed Shot Lv. 1 (same as element ca, Ganyu only)
    - hold = 3 means Fully-Charged Aimed Shot Lv. 2 (Ganyu only) or Fully-Charged Aimed Shot (1 Kindling Arrow) (Yoimiya only)
    - hold = 4 means Fully-Charged Aimed Shot (2 Kindling Arrows) (Yoimiya only)
    - hold = 5 means Fully-Charged Aimed Shot (3 Kindling Arrows) (Yoimiya only)
- hold param defaults to:
    - 3 for Ganyu
    - 2 for everyone else
- in order for Diona's C4, Sara's Crowfeather mechanic and Yelan's Breakthrough Barb to be executed you have to have hold on 2 
- adjusts weakspot param:
    - it was not checked for < 0 and > max before
    - 0 <= weakspot <= 3 on venti
        - 1 = main arrow hits weakspot
        - 2/3 = c1 arrow 1/1+2 hit weakspot 
    - 0 <= weakspot <= 4 on yoimiya
        - 1 = main arrow hits weakspot
        - 2/3/4 = kindling arrow 1/1+2/1+2+3 hit weakspot
    - 0 <= weakspot <= 1 on everyone else
    - defaults to 0
 
also:
- implements ganyu charge lv. 1
- fix wrong yelan barb cancel frames
- fix wrong yelan fully aimed ca arrow scaling
- fix yelan breakthrough barb missing weakspot info
- replace kindling param with hold param on yoimiya
- adjust yoimiya kindling arrow weakspot

TODO:
- [ ] think about ARCC (currently only phys ARCC, maybe add element ARCC, maybe remove ARCC alltogether?)
- [ ] think about RR Cancel
- [ ] think about param syntax
- [ ] update docs after merge

PR #1959, opened 2024-02-01
add AimParam type and tartaglia aim hold param

k0l11:
AimParam:
- 0 = phys
- 1...4 different aimed shot levels (see code comments)
- no ARCC support intended

tartaglia:
- hold = 0: phys
- hold = 1: fully-charged (default)
- fix aimed shot riptide behaviour based on e state before hit:
  - phys, e state not active: no riptide related trigger
  - phys, e state active: apply riptide -> slash
  - fully-charged, e state not active: only apply riptide
  - fully-charged, e state active: apply riptide -> slash
  - fully-charged will not trigger flash if riptide does not exist beforehand


k0l11:
partially sets up continuation of #772

PR #1961, opened 2024-02-02
add custom monster targets

imring:
continue #1105 
- [x] Decide which mobs should be included in gcsim;
- [x] Finalize syntax;
- [x] Design storage for monster data;
- [x] Documentation.


ui/packages/ui/src/Translation/locales/German.json
```diff
@@ -365,5 +366,9 @@
         "dash_data_header": "Gesamtverzögerung des Rennens aufgrund der Abklingzeit pro Iteration",
         "ignore_burst_energy_title": "\"Spezialfähigkeit-Energiekosten ignorieren\"-Modus ist aktiviert",
         "ignore_burst_energy_body": "Die Simulation ignoriert den Energiebedarf für Spezialfähigkeiten. Dies kann zu ungenauen Simulationen führen."
+    },
+    "states": {
+        "true": "wahr",
+        "false": "unwahr"
```
k0l11, line 372: falsch

ui/packages/ui/src/Translation/locales/German.json
```diff
@@ -26,7 +26,8 @@
         "dendro%": "Dendro-SCH-Bonus",
         "anemo%": "Anemo-SCH-Bonus",
         "geo%": "Geo-SCH-Bonus",
-        "cryo%": "Kryo-SCH-Bonus"
+        "cryo%": "Kryo-SCH-Bonus",
+        "modified": "Geändert"
```
k0l11, line 30: Modifiziert

pkg/enemy/types.go
```diff
@@ -0,0 +1,282 @@
...
+
+import (
+	"fmt"
+
+	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/info"
+	"github.com/genshinsim/gcsim/pkg/shortcut"
+)
+
+var levelMultiplier = [][]float64{
```
k0l11, line 11: There is a new hp level multiplier curve that was added in 4.0, see https://genshin-impact.fandom.com/wiki/Enemy/Level_Scaling#HP_Level_Multiplier, so the HP values listed for enemies like Underwater Survey Mek (underwatersurveymek) are wrong.

pkg/enemy/types.go
```diff
@@ -0,0 +1,282 @@
...
+		3791.2686,
+		3934.9932,
+		4082.7458,
+	},
+}
+
+var abyssHpMultipliers = map[string]float64{
+	"ruinserpent":    2.0,
+	"goldenwolflord": 2.0,
+}
```
k0l11, line 221: Should mention in docs that for some enemies 2.5 is not the default multiplier (idk where to best put this list, since it is rather small it could just be stated in config file reference where you discuss what `hp_mult` does as it's not expected to grow very much?)

ui/packages/docs/docs/reference/config.md
```diff
@@ -150,11 +150,30 @@ target lvl=88 resist=0.1 pos=0,0 radius=2 freeze_resist=0.8 hp=9999 particle_thr
...
 
 :::danger
 All configs must have at least one enemy specified. Otherwise you will get an error. 
 :::
 
+:::info
+To use custom HP or resistances with `type`, include them after `type`.
+Example:
+```
+target lvl=100 type=eremitegalehunter[mult=1] radius=2 pos=0,2.4 hp=10000;
```
k0l11, line 163: hp_mult

ui/packages/docs/docs/reference/config.md
```diff
@@ -150,11 +150,30 @@ target lvl=88 resist=0.1 pos=0,0 radius=2 freeze_resist=0.8 hp=9999 particle_thr
...
+
+:::info
+Optimal params may be added to the target type via `[]` after the enemy name.
+:::
+
+#### Optional target type params
+| name | description | default |
+| --- | --- | --- |
+| `hp_mult` | HP multiplier for the enemy. The default is the Spiral Abyss multiplier. | 2.5 |
+| `particles` | Use particle theresholds from the enemy. Use `0` if you want to use a custom threshold (`pacticle_threshold`). | 1 |
```
k0l11, line 175: Btw, are you sure that you need this particles param? Shouldn't it just default to using the particle_threshold, if it is defined, over enemy particles (allowing overwrite by just defining particle_threshold instead of having to specify particle_threshold and particles=1)? Once this comment is resolved, I think you can mark syntax as finalized in the PR TODO (could also ask srl for his opinion on `target ... type=<name>[hp_mult=...]`)

k0l11 (commented):
not a full review yet, just some comments before i go to sleep

k0l11:
For the enemies included in sim part, idk what the best approach would be. Depending on the impact of all the enemy data on wasm size/binary size, we could just put in all enemies.

pkg/enemy/types.go
imring: maybe on their page under the HP table?

ui/packages/docs/docs/reference/config.md
imring: if the user doesn't want the energy generation from either `particle_threshold` or an enemy curve? e.g. use `every energy` instead of them? then they need to use `particle_threshold=99999999 particle_drop_count=0`?

k0l11: Ok, then leave as is.

pkg/enemy/types.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+
+	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/curves"
+	"github.com/genshinsim/gcsim/pkg/core/info"
+	"github.com/genshinsim/gcsim/pkg/shortcut"
+)
+
+var abyssHpMultipliers = map[string]float64{
```
k0l11, line 12: rimebiterbathysmalvishap / bolteaterbathysmalvishap should have 1.75 multiplier btw according to https://genshin-impact.fandom.com/wiki/Enemy/Level_Scaling#HP_Multipliers -> update here and in docs

scripts/enemystat/main.go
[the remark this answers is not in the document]
k0l11: can be done in a later pr, doesn't have to be now, just would recommend opening an issue so this is not forgotten. Also, what datamine version was this last run on?

pkg/core/info/enemy.go
```diff
@@ -4,6 +4,21 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 )
 
+// TODO: protobuf?
```
k0l11, line 7: something to consider when integrating into pipeline / creating issue for this integration (can be done in later PR)

pkg/enemy/types.go
```diff
@@ -0,0 +1,76 @@
...
+type TargetParams struct {
+	HpMultiplier float64
+	Particles    bool
+}
+
+func ConfigureTarget(profile *info.EnemyProfile, name string, params TargetParams) error {
+	if !(1 <= profile.Level && profile.Level <= 100) {
+		return fmt.Errorf("invalid target level: must be between 1 and 100")
+	}
+	if name == "dummy" {
```
k0l11, line 26: This is undocumented, please document it (part where you describe what type does) or remove.

pkg/enemy/types.go
```diff
@@ -0,0 +1,76 @@
...
+		profile.Modified = true
+		profile.ParticleDropThreshold = 0
+		profile.ParticleDropCount = 0
+		profile.ParticleElement = 0
+		profile.ParticleDrops = nil
+		profile.HP = 562949953421311
```
k0l11, line 32: 100000000 (100 million) please

pkg/enemy/types.go
```diff
@@ -0,0 +1,76 @@
...
+		enemyInfo.ParticleDropThreshold = profile.ParticleDropThreshold
+		enemyInfo.ParticleDropCount = profile.ParticleDropCount
+		enemyInfo.ParticleElement = profile.ParticleElement
+		enemyInfo.ParticleDrops = []info.HpDrop{}
+	}
+	*profile = enemyInfo
+	return nil
+}
+
+//go:generate go run github.com/genshinsim/gcsim/scripts/enemystat
```
k0l11, line 65: another thing to consider moving when integrating into pipeline / creating issue for this integration (can be done in later PR)

k0l11:
regarding the failing linters: please fix them by changing the loop to `for i := range a.Targets` instead of `for i, v := range a.Targets` and then replacing the v with `a.Targets[i]` (slice name just an example)

pkg/enemy/types.go
imring: is it necessary if the change was one time?
image

i think to delete them, because they are rarely cases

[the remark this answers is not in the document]
imring: answer below

k0l11: I didn't even notice they were 1 time only, so yeah please delete them.

pipeline/pkg/data/enemy/enemy.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: crutch to get enemy names
+	e.textMap, err = textmap.NewTextMapSource(root + "/../TextMap/TextMapEN.json")
+	if err != nil {
+		return nil, err
+	}
+	return e, nil
```
imring, line 45: is there a better way to get the EN textmap?

k0l11: Maybe something like here? https://github.com/genshinsim/gcsim/blob/8d62c1ec28bf6189fed4f1740fe7a90d3055ec6e/pipeline/cmd/generate/main.go#L120

imring: fixed a little, but the question is still open
maybe add a general textmap (en) to `cmd/generate/main.go` for cases like this?

k0l11: idk up to you

PR #1966, opened 2024-02-02
modify airborne check

srliao:
preparing for some old bird lady thinggy

### DIFF
--- a/internal/characters/keqing/jump.go
+++ b/internal/characters/keqing/jump.go
@@ -1,0 +1,17 @@
+package keqing
+
+import (
+	"errors"
+
+	"github.com/genshinsim/gcsim/pkg/core/action"
+	"github.com/genshinsim/gcsim/pkg/core/player"
+)
+
+func (c *char) Jump(p map[string]int) (action.Info, error) {
+	if c.Core.Status.Duration(player.XianyunAirborneBuff) > 0 {
+		// TODO: implementation to set airborne flag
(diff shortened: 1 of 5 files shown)


internal/characters/keqing/jump.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"errors"
+
+	"github.com/genshinsim/gcsim/pkg/core/action"
+	"github.com/genshinsim/gcsim/pkg/core/player"
+)
+
+func (c *char) Jump(p map[string]int) (action.Info, error) {
+	if c.Core.Status.Duration(player.XianyunAirborneBuff) > 0 {
```
k0l11, line 11: if c.StatusIsActive(player.XianyunAirborneBuff)

pkg/core/player/player.go
```diff
@@ -311,22 +305,23 @@ const (
 	AirborneXiao
 	AirborneVenti
 	AirborneKazuha
```
k0l11, line 307: won't be used i believe

internal/characters/keqing/jump.go
srliao: not fixing for now since this is an example and should be deleted when xianyun gets implemented. chances are we should be checking individual char mods anyways

pkg/core/player/player.go
srliao: pass for now. maybe someone will do in future

PR #1974, opened 2024-02-03
add stats to pipeline

srliao:


srliao:
Example of what to add to `config.yaml` to use this

```yaml
generate_skill_data: true
skill_data_mapping:
  attack:
    0:
      n1
    1: 
      n2 
    2:
      n3_1
    3:
      n3_2
    4:
      n4
    5:
      ca1
    6:
      ca2
```

The result will look something like:
image

You will probably need to declare another variable to consolidate all the attacks together, something like

```go
var attack = [][][]float64{
	{
		n1,
	},
	{
		n2,
	},
	{
		n3_1, n3_2,
	},
	{
		n4,
	},
}
```

PR #1988, opened 2024-02-07
add NextQueueItemIsValid to Character Interface

k0l11:
- refactor error if using charge without attack

### DIFF
--- a/internal/template/character/ready.go
+++ b/internal/template/character/ready.go
@@ -1,8 +1,12 @@
 package character
 
 import (
+	"fmt"
+
 	"github.com/genshinsim/gcsim/pkg/core/action"
 	"github.com/genshinsim/gcsim/pkg/core/glog"
+	"github.com/genshinsim/gcsim/pkg/core/info"
+	"github.com/genshinsim/gcsim/pkg/core/player"
 )
 
 func (c *Character) ActionReady(a action.Action, p map[string]int) (bool, action.Failure) {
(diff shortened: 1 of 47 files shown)


internal/template/character/ready.go
```diff
@@ -37,3 +41,17 @@ func (c *Character) ActionReady(a action.Action, p map[string]int) (bool, action
 	}
 	return true, action.NoFailure
 }
+
+func (c *Character) NextQueueItemIsValid(next action.Eval) error {
```
srliao, line 45: this implementation should be in `template/Character`

pkg/simulation/run.go
```diff
@@ -133,6 +133,12 @@ func queuePhase(s *Simulation) (stateFn, error) {
 		// we do the same skip here as if eval doesn't have any more ations
 		return s.advanceFrames(1, queuePhase)
 	}
+	// check if the next queue item is valid
+	// example: most sword characters can't do charge if the previous action was not attack
+	currentChar := s.C.Player.ActiveChar()
+	if err := currentChar.NextQueueItemIsValid(*next); err != nil {
```
srliao, line 139: this check here should be just right after the swap check

srliao: also probably ok to pass next *action.Eval? not sure if any benefit to having signature non pointer and having to dereference here

srliao (dismissed):

PR #2016, opened 2024-03-03
character plunge frames/attack implementation - airborne from xianyun

Charlie-Zheng:
- [x] albedo
- [x] alhaitham
- [x] ayaka
- [x] ayato
- [x] barbara
- [x] bennett
- [x] charlotte
- [x] cyno (out/in q state)
- [x] dehya
- [x] diluc
- [x] dori
- [x] eula
- [x] freminet
- [x] furina
- gaming **not in this pr**
- [x] heizou
- [x] hu tao (out/in e state)
- [x] itto
- [x] kaveh
- [x] kazuha
- [x] keqing
- [x] klee
- [x] lynette
- [x] navia
- [x] noelle
- [x] raiden (out/in q state)
- [x] sucrose
- [x] wriothesley
- [x] xiangling
- [x] xianyun
- [x] xiao (out/in q state)
- [x] yae
- [x] yanfei


internal/characters/bennett/plunge.go
```diff
@@ -0,0 +1,164 @@
...
+	collision, ok := p["collision"]
+	if !ok {
+		collision = 0 // Whether or not Xiao does a collision hit
+	}
+
+	if collision > 0 {
+		c.plungeCollision(collisionHitmark)
+	}
+
+	poiseDMG := 150.0
```
srliao, line 65: nit: i assume compiler optimizes this variable away. doesn't look like it gets changed anywhere. should probably get rid of the variable here

internal/characters/albedo/plunge.go
```diff
@@ -0,0 +1,164 @@
...
+	collision, ok := p["collision"]
+	if !ok {
+		collision = 0 // Whether or not collision hit
+	}
+
+	if collision > 0 {
+		c.plungeCollision(collisionHitmark)
+	}
+
+	poiseDMG := 150.0
```
srliao, line 65: nit: see comment on bennett

internal/characters/albedo/plunge.go
```diff
@@ -0,0 +1,164 @@
...
+		highPlungeHitmark,
+	)
+
+	return action.Info{
+		Frames:          frames.NewAbilFunc(highPlungeFrames),
+		AnimationLength: highPlungeFrames[action.InvalidAction],
+		CanQueueAfter:   highPlungeFrames[action.ActionDash],
+		State:           action.PlungeAttackState,
+	}, nil
+}
```
srliao, line 93: there needs to be something at the end of the plunge to reset the airborne flag back to grounded. otherwise the status remains `AirborneXianyun`

this is needed on every single plunge action. strictly speaking i guess even jump should have a reset as well to simulate slow falling and no action taken but i think it's fine to ignore that use case

srliao (dismissed):

srliao:
In addition to my comment re setting status, every jump action should also probably return an error (and therefore ending the sim) if already airborne (i.e. `jump, jump` should crash the sim). 

This is because if we don't simulate slow falling to ground, then you can't jump again while in mid air

Charlie-Zheng:
> In addition to my comment re setting status, every jump action should also probably return an error (and therefore ending the sim) if already airborne (i.e. `jump, jump` should crash the sim).
> 
> This is because if we don't simulate slow falling to ground, then you can't jump again while in mid air

Hmm I thought this would be like globally applied to all characters (no actions except plunge can be done while airborne), and it's up to the individual characters to override

srliao:
> In addition to my comment re setting status, every jump action should also probably return an error (and therefore ending the sim) if already airborne (i.e. `jump, jump` should crash the sim).
> 
> This is because if we don't simulate slow falling to ground, then you can't jump again while in mid air

retracted. already handled here. i forgot

https://github.com/genshinsim/gcsim/blob/cb8fccbc4ce4f45c48a55e9db081cc621dd3d538/pkg/core/player/exec.go#L127-L129

PR #2017, opened 2024-03-04
added fixed seed for gradients

Charlie-Zheng:
When calculating gradients, the seeds are all the same now for that one gradient. This reduces unstable and non-optimal results when optimizing sims with random delays.

Seeds are created by time.Now().UnixNanos()

### DIFF
--- a/pkg/optimization/optstats/simulator.go
+++ b/pkg/optimization/optstats/simulator.go
@@ -2,8 +2,8 @@ package optstats
 
 import (
 	"context"
+	"math/rand"
 	"slices"
-	"time"
 
 	"github.com/genshinsim/gcsim/pkg/agg"
 	"github.com/genshinsim/gcsim/pkg/core/info"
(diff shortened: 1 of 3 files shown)


pkg/optimization/optstats/simulator.go
```diff
@@ -38,6 +38,12 @@ func RunWithConfigCustomStats[T any](ctx context.Context, cfg string, simcfg *in
 	pool := WorkerNewWithCustomStats(simcfg.Settings.NumberOfWorkers, respCh, errCh, customCh)
 	pool.StopCh = make(chan bool)
 
+	seeds := make([]int64, simcfg.Settings.Iterations)
```
srliao, line 41: You should be able to simplify this code to
```go
    // spin off a go func that will queue jobs for as long as the total queued < iter
    // this should block as queue gets full
    go func() {
        src := rand.NewSource(seed)
        // make all the seeds
        wip := 0
        for wip < simcfg.Settings.Iterations {
            pool.QueueCh <- JobCustomStats[T]{
                Cfg:     simcfg.Copy(),
                Actions: gcsl.Copy(),
                Seed:    src.Int63(),
                Cstat:   cstat,
            }
            wip++
        }
    }()
    ```

srliao (approved):

PR #2031, opened 2024-03-09
Xianyun jump

Charlie-Zheng:
Frames for dash jumps here
https://docs.google.com/spreadsheets/d/1B0QP2iSruOdYVB2ouJ0BYi_jfn77_IzMHObkHb3Cp8w/edit#gid=0


internal/template/character/stam.go
```diff
@@ -102,18 +111,51 @@ func (c *Character) DashLength() int {
...
+		switch c.Weapon.Class {
+		case info.WeaponClassSword, info.WeaponClassSpear:
+			lowPlunge = 5
+		}
+
+		highPlunge := 8
+		switch c.Weapon.Class {
+		case info.WeaponClassSword, info.WeaponClassSpear:
+			highPlunge = 9
+		}
```
k0l11, line 141: Please combine these two switch statements because you don't need to check the weapon class twice in the same way and to save vertical space in an already long if block, so
```
// 4/8 for claymore/bow/catalyst and 5/9 for sword/polearm
lowPlunge := 4
highPlunge := 8
switch c.Weapon.Class {
case info.WeaponClassSword, info.WeaponClassSpear:
	lowPlunge = 5
	highPlunge = 9
}
```

internal/template/character/stam.go
```diff
@@ -102,18 +111,51 @@ func (c *Character) DashLength() int {
...
+				switch a {
+				case action.ActionLowPlunge:
+					return lowPlunge
+				case action.ActionHighPlunge:
+					return highPlunge
+				default:
+					return 60
+				}
+			},
+			AnimationLength: 60,
```
k0l11, line 153: Please put this in a variable and use it here and in the frames func. Also add a comment indicating what this actually represents (upper bound for jump for low/high plunge i assume).

internal/template/character/stam.go
```diff
@@ -102,18 +111,51 @@ func (c *Character) DashLength() int {
...
+		}
+		return action.Info{
+			Frames: func(a action.Action) int {
+				switch a {
+				case action.ActionLowPlunge:
+					return lowPlunge
+				case action.ActionHighPlunge:
+					return highPlunge
+				default:
+					return 60
```
k0l11, line 150: Please add a comment that you expect this case to be ineffectual because sim will error out on xianyun buffed jump into anything else than low plunge/high plunge.

internal/template/character/stam.go
```diff
@@ -102,18 +111,51 @@ func (c *Character) DashLength() int {
...
-	case info.BodyMale:
-		f = 28
-	case info.BodyLady:
-		f = 32
-	case info.BodyLoli:
-		f = 29
+	case info.BodyGirl, info.BodyLoli:
+		return 4
+	case info.BodyMale, info.BodyLady:
+		return 3
```
k0l11, line 119: Please remove this case as it is already covered by the default.

internal/template/character/stam.go
```diff
@@ -122,6 +164,31 @@ func (c *Character) Jump(p map[string]int) (action.Info, error) {
...
 
+func (c *Character) JumpLength() int {
+	if c.Core.Player.LastAction.Type == action.ActionDash {
+		switch c.CharBody {
+		case info.BodyLoli, info.BodyLady, info.BodyMale:
+			return 37
+		case info.BodyGirl, info.BodyBoy:
+			return 34
+		default:
+			return 36
```
k0l11, line 175: Please use one of the other two cases as default since the 36 is actually not a frame count for any of the models above.

PR #2038, opened 2024-03-10
add basic ending energy stats

srliao:
image

Simple ending energy graph. Localization not done (someone else can handle it)


k0l11:
fixing later

PR #2042, opened 2024-03-11
Changed Xianyun's burst stack timing

Charlie-Zheng:
Updated to reduce at the end of the frame that triggers, instead of the end of the next frame


k0l11:
closing in favor of #2043

PR #2089, opened 2024-03-18
Substatopt 4 star

Charlie-Zheng:
Fixed the stat output not using the 4* stat modifier. 
Added each 4* artifact reducing the total substat upper limit by 2. Currently no effect on individual substat cap.
Added each 4* artifact reducing the substat multiplier by 0.04 (4x 4* artifacts will have a substat modifier of 0.84x)
Added main stat value verification. Incorrect mainstat values will give a warning.


Charlie-Zheng:
Should I make incorrect mainstat values be a FATAL error instead of a WARNING?

Charlie-Zheng:
Converting to draft until KQM finalizes their 4* stat requirements

PR #2095, opened 2024-03-20
Add param validation to characters

selvuen:


internal/characters/xianyun/config.yml
```diff
@@ -1,6 +1,15 @@
 package_name: xianyun
 genshin_id: 10000093
 key: xianyun
+action_param_keys:
+  attack:
+    - param: "travel"
+  charged:
```
srliao, line 7: this should say `charge`

srliao: will fix in separate PR. there are other code issues unrelated to this so I'll just fix it together

PR #2161, opened 2024-04-26
Clam and SoDP use total healing before BoL reduction.

### LINKED ISSUE #2160: Clam and Sodp use actual heal value instead of pre BoL heal value
From a short conversation in KQM, apparently BoL doesn't reduce the healing accumulated by Clam or Sodp.

Charlie-Zheng:
Closes https://github.com/genshinsim/gcsim/issues/2160


srliao:
Blame Charlie if this is wrong

PR #2163, opened 2024-05-03
add Fragment of Harmonic Whimsy

Siwat-Keng:
Artifact PR checklist:

- [x] New artifact package
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation


internal/artifacts/fragmentofharmonicwhimsy/fragmentofharmonicwhimsy.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if count >= 4 {
+		m := make([]float64, attributes.EndStatType)
+		m[attributes.DmgP] = 0.18
+		c.Events.Subscribe(event.OnHPDebt, func(args ...interface{}) bool {
+			index := args[0].(int)
+			amount := args[1].(float64)
+			if char.Index == index && amount != 0 {
+				char.AddStatMod(character.StatMod{
+					Base: modifier.NewBaseWithHitlag(fmt.Sprintf("fragmentofharmonicwhimsy-4pc-%v-stack", s.counter+1), 6*60),
```
srliao, line 51: This would cause every stack to have independent duration. The description sounds like if you get a new stack every stack should be refreshed. This needs to be confirmed

srliao (changes requested):

internal/artifacts/fragmentofharmonicwhimsy/fragmentofharmonicwhimsy.go
```diff
@@ -0,0 +1,78 @@
...
+	if count >= 4 {
+		m := make([]float64, attributes.EndStatType)
+		c.Events.Subscribe(event.OnHPDebt, func(args ...interface{}) bool {
+			index := args[0].(int)
+			amount := args[1].(float64)
+			if char.Index != index || amount == 0 {
+				return false
+			}
+
+			if !char.StatusIsActive(fohw4pc) {
```
srliao, line 56: This should use `StatModIsActive` even though they both work. Just in case implementation changes in future this won't break

pkg/shortcut/artifacts.go
```diff
@@ -40,6 +40,9 @@ var SetNameToKey = map[string]keys.Set{
 	"flowerofparadiselost":               keys.FlowerOfParadiseLost,
 	"paradiselost":                       keys.FlowerOfParadiseLost,
 	"fopl":                               keys.FlowerOfParadiseLost,
+	"fragmentofharmonicwhimsy":           keys.FragmentOfHarmonicWhimsy,
+	"harmonicwhimsy":                     keys.FragmentOfHarmonicWhimsy,
```
srliao, line 44: Forgot to say just two names please (one shortcut + one actual)

srliao (changes requested):

srliao (approved):

PR #2164, opened 2024-05-03
add Crimson Moon's Semblance

Siwat-Keng:
Weapon PR checklist:

- [x] New weapon package
- [x] Run pipeline with added config (generates weapon curve, `.generated.json` files)
- [x] Weapon key
- [x] Update documentation


internal/weapons/spear/crimsonmoonssemblance/crimsonmoonssemblance.go
```diff
@@ -0,0 +1,87 @@
...
+			return false
+		}
+
+		char.AddStatus(icdKey, icdDuration, true)
+		char.ModifyHPDebtByRatio(0.25)
+
+		return false
+	}, fmt.Sprintf("crimsonmoonssemblance-hit-%v", char.Base.Key.String()))
+
+	c.Events.Subscribe(event.OnHPDebt, func(args ...interface{}) bool {
```
srliao, line 58: I *think* this entire block should just be an AddStatMod, with the current debt percent being a check inside the `Amount` func.

This code here would cause the mod to be updated only when hp debt changes. While the result is the same (since event would have to be triggered for BOL to reduce), it seems a little inefficient to re-create a slice and re-add a mod every time BOL changes. Better to just add a permanent mod and check the BOL amount each time ?

internal/weapons/spear/crimsonmoonssemblance/crimsonmoonssemblance.go
```diff
@@ -0,0 +1,77 @@
...
+		return false
+	}, fmt.Sprintf("crimsonmoonssemblance-hit-%v", char.Base.Key.String()))
+
+	m := make([]float64, attributes.EndStatType)
+
+	char.AddStatMod(character.StatMod{
+		Base:         modifier.NewBaseWithHitlag("crimsonmoonssemblance-bonus", -1),
+		AffectedStat: attributes.DmgP,
+		Amount: func() ([]float64, bool) {
+			maxhp := char.MaxHP()
```
srliao, line 64: nit: this `maxhp` store here seems unnecessary

internal/weapons/spear/crimsonmoonssemblance/crimsonmoonssemblance.go
```diff
@@ -0,0 +1,77 @@
...
+			m[attributes.DmgP] = 0.0
+			if char.CurrentHPDebt() > 0 {
+				m[attributes.DmgP] += 0.08 + 0.04*float64(refine)
```
srliao, line 67: nit: this calculation should probably just be done once and stored outside? although I'm actually curious which is better: storing this in a variable that I'm assuming is not on the stack or just redoing this simple calc

srliao (approved):

PR #2180, opened 2024-06-15
add unfinishedreverie

Siwat-Keng:
- [x] New artifact package
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation


pkg/core/attributes/element.go
```diff
@@ -17,6 +17,7 @@ const (
 	Hydro
 	Dendro
 	Quicken // or overdose
+	Burning
```
srliao, line 20: I don't believe Burning is an actual element but a reaction so we shouldn't add a new element here. See my comment in reactable for possible solution

pkg/reactable/reactable.go
```diff
@@ -293,6 +293,9 @@ func (r *Reactable) AuraContains(e ...attributes.Element) bool {
 		if v == attributes.Cryo && r.Durability[Frozen] > ZeroDur {
 			return true
 		}
+		if v == attributes.Burning && r.Durability[BurningFuel] > ZeroDur && r.Durability[Burning] > ZeroDur {
+			return true
+		}
```
srliao, line 298: Instead of having this check as part of AuraContains, more sensible to just add a new helper function call it `IsBurning() bool` or something and do the same check. This way you don't need the new element as well

srliao (changes requested):

srliao:
FYI I didn't check the actual artifact code yet, only commenting on the Burning element part of it

internal/artifacts/unfinishedreverie/unfinishedreverie.go
```diff
@@ -0,0 +1,90 @@
...
+				return m, true
+			},
+		})
+	}
+
+	return &s, nil
+}
+
+func (s *Set) enemyCheck(char *character.CharWrapper, c *core.Core) func() {
+	return func() {
```
srliao, line 68: is a function generator needed here? since this is already a method of set could just store `*character.CharWrapper` and `*core.Core` inside `Set`?

internal/artifacts/unfinishedreverie/unfinishedreverie.go
```diff
@@ -0,0 +1,90 @@
...
+			if e.IsBurning() {
+				s.lastF = c.F
+				break
+			}
+		}
+		if c.F-s.lastF > 6*60 && !char.StatusIsActive(IcdKey) && s.stacks > 0 {
+			s.stacks--
+			char.AddStatus(IcdKey, icd, false)
+		}
+		if c.F == s.lastF && !char.StatModIsActive(IcdKey) && s.stacks < 5 {
```
srliao, line 84: nit: should be else if here right?

internal/artifacts/unfinishedreverie/unfinishedreverie.go
```diff
@@ -0,0 +1,90 @@
...
+				continue
+			}
+			if e.IsBurning() {
+				s.lastF = c.F
+				break
+			}
+		}
+		if c.F-s.lastF > 6*60 && !char.StatusIsActive(IcdKey) && s.stacks > 0 {
+			s.stacks--
+			char.AddStatus(IcdKey, icd, false)
```
srliao, line 82: nit: probably slightly more performant to just use a variable to track the frame icd last triggered since you don't need hitlag here

srliao (dismissed):

PR #2192, opened 2024-08-10
Previous DB requests cancellation

Abdock:
Added feature that cancel previous requests if new was run


Abdock:
I ran into some problem, when I had a bad internet connection, queries `/api/db?q={query string}` were not always executed, or not in the order they should be, for example if I wanted to sort by DPS in descending order, the second click did not work, after researching I noticed that the problem may be due to the fact that the previous query that was sent on a given endpoint is not canceled.

unleashurgeek (approved):

PR #2212, opened 2024-08-31
add nightsoul

imring:


internal/template/nightsoul/nightsoul.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"github.com/genshinsim/gcsim/pkg/core"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
+	"github.com/genshinsim/gcsim/pkg/core/player/character"
+)
+
+const NightsoulBlessingStatus = "nightsoul-blessing"
+
+type Nightsoul struct {
```
srliao, line 12: nit: this should probably be called something that's not `Nightsoul`. there's a naming guide somewhere in go doc that says avoid names like `nightsoul.Nightsoul`. I'm bad with names though so... whatever works `nightsoul.Blessing`? `nightsoul.Tracker`?

srliao (dismissed):

PR #2246, opened 2024-10-14
Adjusted the ICD so that Mualani gains stacks every 0.7s

### LINKED ISSUE #2245: Mualani gains momentum stacks every 0.8s instead of 0.7s
https://gcsim.app/sh/9WFtM6WnnzzK
image

Charlie-Zheng:
Closes #2245


imring:
fixed #2248

PR #2250, opened 2024-10-20
clean out trash

srliao:
per popular user demands, gcsim will now only contain developers approved characters, starting by clearing out trash characters

TODO: aether needs to be deleted still


clevernt (approved):
LGTM

PR #2294, opened 2024-12-16
Swapped DEF and DEF% in substats to be the same order as HP and ATK stats

JohnnyToumieh:
This will fix my OCD 👍

### DIFF
--- a/pkg/core/attributes/stats.go
+++ b/pkg/core/attributes/stats.go
@@ -13,8 +13,8 @@ type (
 // stat types
 const (
 	NoStat Stat = iota
-	DEFP
 	DEF
+	DEFP
 	HP
 	HPP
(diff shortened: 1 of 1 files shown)


srliao:
I have a feeling is going to break a ton of stuff because there are hard coded ordering in UI and elsewhere... I could have remembered wrong though

JohnnyToumieh:
Hmmmmmm ok maybe ill unrequest then since its unlikely i can find all of this information with a simple search

JohnnyToumieh:
I made another one where i ctrl-f "def%" and "DEFP" and swapped them around. won't try to pull request tho cauz idk if there's anything hard coded that could be broken. Like if something is specifically checking if first substat is flat DEF

PR #2301, opened 2025-01-01
Initial mavuika implementation

Charlie-Zheng:
- [x] New character package
- [x] Character key
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation
- [x] Xianyun Plunge (optional)


internal/characters/mavuika/cons.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (c *char) c6RingCB() func(a combat.AttackCB) {
+	if c.Base.Cons < 6 {
+		return nil
+	}
+	return func(a combat.AttackCB) {
+		if a.Target.Type() != targets.TargettableEnemy {
+			return
+		}
+		if c.StatusIsActive(c6IcdKey) {
```
jwalk-511, line 140: Does this status have a function, since the ring can only trigger once every 2 seconds? Or is it just an extra sanity in case the ring hit interval is somehow decreased?

Or does the cb from QueueAttack get called for every target hit, so this is required to prevent multiple triggers from a single ring hit in aoe?

internal/characters/mavuika/jump.go
```diff
@@ -0,0 +1,33 @@
...
+
+var bikeJumpFrames []int
+
+func init() {
+	bikeJumpFrames = frames.InitAbilSlice(42)
+	bikeJumpFrames[action.ActionLowPlunge] = 16
+}
+
+func (c *char) Jump(p map[string]int) (action.Info, error) {
+	if !c.StatusIsActive(player.XianyunAirborneBuff) && c.armamentState == bike &&
```
jwalk-511, line 17: Leaving note for reference for plunge code review. If Mavuika has Xianyun Airborne Buff, Mavuika will do a Xianyun-buffed jump, no matter whether Mav has bike armament state or not.

internal/characters/mavuika/plunge.go
```diff
@@ -0,0 +1,213 @@
...
+	bikePlungeFrames[action.ActionSwap] = 75
+}
+
+// Low Plunge attack damage queue generator
+// Use the "collision" optional argument if you want to do a falling hit on the way down
+// Default = 0
+func (c *char) LowPlungeAttack(p map[string]int) (action.Info, error) {
+	defer c.Core.Player.SetAirborne(player.Grounded)
+	switch c.Core.Player.Airborne() {
+	case player.AirborneXianyun:
```
jwalk-511, line 71: Leaving note for plunge reference. If Mav has AirborneXianyun, Mav will perform lowPlungeXY whether or not she has bike state

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+}
+
+func (*char) GetHittableEntityList() []HittableEntity {
+	return bikeHittableEntityList
+}
+
+// Iterate through CA frames, starting at hitmark
+func (c *char) CalculateValidCollisionFrames(durationCA int, collisionFrames [2]int, lastHitFrame int) []int {
+	validFrames := []int{}
+	currentFrame := 35 // Spin starts on frame 35 of CA anim (full windup)
```
jwalk-511, line 610: I don't think this function works when called from the ca hook cb. If a dendro core is created on some arbitrary frame in the middle of a function, Time Since Start should be something like c.Core.F - c.caState.srcF (but accounting for hitlag).

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+	isForceFinalHit := false // Used when exceeding CA duration, forces CAF
+
+	if final == 1 {
+		return c.BikeChargeAttackFinal(0, skippedWindupFrames), nil
+	}
+
+	// Do not allow starting with a partial CA hold
+	if durationCA > 0 && c.caState.cAtkFrames > 0 {
+		// Cap duration to lowest of 1 spin, remaining NS, or max CA time
+		durationCA = min(durationCA, bikeChargeAttackSpinFrames, nightSoulDuration, bikeChargeAttackMaximumDuration-c.caState.cAtkFrames)
```
jwalk-511, line 156: Recommend adding to documentation that the "hold" parameter represents time to remain in CA state not counting windup time, as well as the fact that the minimum is 1f and maximum is 1 complete rotation (bikeChargeAttackSpinFrames). Also note that by default, the logic is not to hold for 0f but to spin until the primary target is hit.

Also recommend leaving a note for the skill itself (not for the param), that the duration will automatically be cut short if max CA time is reached or NS expires- but this note is less important since it's a reasonable assumption to make anyway.

jwalk-511 (commented):
Review is not done, just leaving comments so far as a starting point (Notably, I have reviewed all functions called by BikeCharge from line 121 to 147, and functions called by those functions, recursively. But I have not reviewed functions only called by BikeCharge after line 147, including `HoldBikeChargeAttack` and `CountBikeChargeAttack`)

internal/characters/mavuika/charge.go
Aftermathrar: True, I changed this recently from the calc you had to cAtkFrames (when trying to separate out windup) and didn't realize the implication. I will change it back.

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+}
+
+func (c *char) buildValidGadgetList() []HittableEntity {
+	gadgets := c.Core.Combat.GadgetsWithinArea(combat.NewCircleHitOnTarget(c.Core.Combat.Player(), nil, 8), nil)
+	var hittableGadgets []HittableEntity
+	for _, g := range gadgets {
+		if g == nil {
+			continue
+		}
+		switch g.GadgetTyp() {
```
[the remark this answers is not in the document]
Aftermathrar: I might be interpreting this wrong, but it looks like all gadgets are targettable, so a default case would be handling pretty much every other gadget type and may not know if something is "new" unless the current case select has every gadget type explicitly handled. I think either way would require a bug report in the future, but I might be missing an easier solution here.

internal/characters/mavuika/plunge.go
Aftermathrar: Done, added bike/unbiked versions for XY plunge and unbiked plunge from a walked bike jump.

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+	return newMinSpinDuration
+}
+
+func (c *char) BuildBikeChargeAttackHittableTargetList() []HittableEntity {
+	return append(c.buildValidTargetList(), c.buildValidGadgetList()...)
+}
+
+func (c *char) buildValidTargetList() []HittableEntity {
+	enemies := c.Core.Combat.EnemiesWithinArea(combat.NewCircleHitOnTarget(c.Core.Combat.Player(), nil, 8), nil)
+	hittableEnemies := []HittableEntity{}
```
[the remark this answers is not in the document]
Aftermathrar: Done.

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+// Calculate start and end frames for each spin during which target is within Mav hitbox
+// Return false if target is not circle or has no overlap
+func (c *char) BikeHitboxIntersectionAngles(v combat.Target, f []int, offsetAngle float64) bool {
+	enemyShape := v.Shape()
+	var enemyRadius float64
+	switch v := enemyShape.(type) {
+	case *geometry.Circle:
+		enemyRadius = v.Radius() // Rt
+	default:
+		return false
```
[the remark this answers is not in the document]
Aftermathrar: Added error saying Mav needs circular hitboxes for CA calcs.

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+		c.caState = ChargeState{}
+		c.caState.StartFrame = c.Core.F
+		c.caState.LastHit = make(map[int]int)
+		for _, t := range bikeHittableEntities {
+			targetIndex := int(t.Entity.Key())
+			c.caState.LastHit[targetIndex] = 0
+		}
+		c.Core.Tasks.Add(func() {
+			c.bikeChargeAttackHook()
+		}, bikeChargeAttackStartupHitmark-1)
```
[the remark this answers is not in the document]
Aftermathrar: Went with the first option, since cores could be missed in the scenario you described.

internal/characters/mavuika/charge.go
```diff
@@ -0,0 +1,763 @@
...
+			break
+		}
+	}
+	if !isTargetForCountsHittable {
+		return false, action.Info{}, errors.New("primary target is not within flamestrider area")
+	}
+	return true, action.Info{}, nil
+}
+
+// Important Events: OnTargetDied, OnTargetMoved (also emits on player move?), OnDendroCore
```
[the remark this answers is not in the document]
Aftermathrar: Comment updated.

internal/characters/mavuika/charge.go
Aftermathrar: Updated documentation to add this.

internal/characters/mavuika/charge.go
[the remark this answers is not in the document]
jwalk-511: I think you're right; I was assuming that most gadgets would have some flag indicating they shouldn't be targeted by player abilities but if that isn't present then it isn't a super huge issue.

It could only possibly matter if some future character can create a gadget that can be hit by any other character, which is a rare case anyway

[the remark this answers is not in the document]
Charlie-Zheng: Generally if a player cannot hit the gadget, there is no reason to actually code the object as a gadget.

imring (approved):

PR #2313, opened 2025-01-24
Andriod server

Charlie-Zheng:
Added a build for andriod server apk.


Charlie-Zheng:
This is ready for review

pkg/servermode/handler.go
```diff
@@ -273,3 +273,33 @@ func (s *Server) info() http.HandlerFunc {
...
+
+type Progress struct {
+	Curr int
+	Max  int
+	Err  error
+}
+
+func (s *Server) Progress() map[string]Progress {
+	progress := make(map[string]Progress)
+	for id, wk := range s.pool {
```
srliao, line 285: `s.pool` and more specifically, the `worker` held by `s.pool` ideally should not be accessed like this. Although you're not doing any write operations here so it _probably_ won't cause any problem, this is still non-ideal. `s.pool` is not safe for concurrent accesses.

haven't thought about what a better solution would be. possibly some kind of concurrency safe read only method on workers to get the info you want?

Charlie-Zheng: Should I run `s.Lock()` before calling progress?

Issue #2291, opened 2024-12-08
Fix Yelan Burst frames and add N0 frames


just1ce5:
The issue of "N2C Arlecchino combo being almost impossible in-game, but possible in gcsim" is quite popular. I did my investigation and I think that inaccurate Yelan's burst frames may be the reason of that issue.

One may observe that time between Arlecchino N1 animation start and Yelan's burst hitmark is ~23 frames in-game and ~32 frames in gcsim https://gcsim.app/db/WhmPKDGpLq6q

The 7-9 frames difference is Yelan N0 animation start delay, which is "added" to the Yelan burst hitmark frame count.

TODO:

- Subtract Yelan N0 frames from burst hitmarks
- Add Yelan N0
- (optional) Add option `travel` for burst projectiles since in some scenarios N2C Arlecchino may work when the enemy is moving and travel time is long enough.

Evidence:
https://youtu.be/fIgO3E2dQs4

PR #2315, opened 2025-01-25
Modify yelan hitmarks.

just1ce5:
Addresses #2291 .

### DIFF
--- a/internal/characters/yelan/burst.go
+++ b/internal/characters/yelan/burst.go
@@ -11,7 +11,7 @@ import (
 )
 
 var burstFrames []int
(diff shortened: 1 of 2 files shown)


soloxcx:
8 frames for Yelan N0 window sounds too high. Based on your [recording](https://www.youtube.com/watch?v=fIgO3E2dQs4) it is perhaps more like 3-4 frames, which is also most similar to other units. However you need a 2nd character in the team to catch the 'grey-out' frame for when her jump starts. I was planning to tackle that issue so nbd if you don't, but I would re-record Yelan N0 timings according to the frame guide, also add the video to docs

imring:
closed by #2349

PR #2327, opened 2025-01-27
Fix Clorinde eNA AtkSpd Not Applying

Aftermathrar:
Atk Speed did not apply accurately, included adjusting animation length to account for attack speed. 
Hitmarks were moved up slightly (should be more accurate) to assist with BoL logic in configs.


internal/characters/mavuika/asc.go
```diff
@@ -0,0 +1,60 @@
+package mavuika
```
imring, line 1: mavuika?

imring (changes requested):

PR #2332, opened 2025-02-06
Fix xilonen geo sampler being active if all samplers were converted

jwalk-511:
https://discord.com/channels/845087716541595668/1336939788925931561

### DIFF
--- a/internal/characters/xilonen/cons.go
+++ b/internal/characters/xilonen/cons.go
@@ -59,6 +59,10 @@ func (c *char) c2() {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


jwalk-511:
I over-check sampler activity to try to "future-proof" edits. Probably not needed; either the call to activeGeoSampler() or the areas where the method is called could be reverted with no loss in functionality.

I chose to use checks against c.samplersConverted < or >= 3, this could be replaced with checks against cshredElements[attributes.Geo] if desired.

PR #2333, opened 2025-02-09
Add timed Nightsoul Blessing

jwalk-511:
Add functions to Nightsoul module to allow setting up NS with an optional cb for when NS status ends. This CB will only be called if NS expires via the timer, not if NS is manually expired, and is intended to clean up the character state on Nightsoul expiry. As an alternative, an OnNightsoulExpiry event could be used, or an array of cb functions could be allowed.

Kinich refactored to make use of this functionality.

GCSL users can get remaining NS duration through .<character>.status.nightsoul-blessing.

For now, it's assumed that all NS timers will be hitlag extended. If that is not the case, code can be further refactored to accept a hitlag bool, to be passed through to the char.AddStatus method call.

I manually tested a few Kinich configs and got matching results, but I have not run a full DB compare.
If desired, I can split this PR into two; one for NS changes and one for Kinich changes.

### DIFF
--- a/internal/characters/kinich/burst.go
+++ b/internal/characters/kinich/burst.go
@@ -2,6 +2,7 @@ package kinich
 
 import (
 	"github.com/genshinsim/gcsim/internal/frames"
+	"github.com/genshinsim/gcsim/internal/template/nightsoul"
 	"github.com/genshinsim/gcsim/pkg/core/action"
 	"github.com/genshinsim/gcsim/pkg/core/attacks"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
@@ -33,8 +34,10 @@ func (c *char) Burst(p map[string]int) (action.Info, error) {
 	c.AddStatus(burstKey, ajawDuration, false)
 	if c.nightsoulState.HasBlessing() {
 		// extend Nightsoul's Blessing time limit countdown
-		duration := (c.exitStateF - c.Core.F) + 1.7*60
-		c.setNightsoulExitTimer(duration)
+		duration := c.StatusDuration(nightsoul.NightsoulBlessingStatus)
+		if duration > 0 {
+			c.nightsoulState.SetNightsoulExitTimer(duration+1.7*60, c.cancelNightsoul)
+		}
 	}
 
 	ai := combat.AttackInfo{
--- a/internal/characters/kinich/burst.go
+++ b/internal/characters/kinich/burst.go
@@ -2,7 +2,6 @@ package kinich
 
 import (
 	"github.com/genshinsim/gcsim/internal/frames"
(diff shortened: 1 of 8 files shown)


internal/characters/kinich/burst.go
```diff
@@ -33,8 +34,10 @@ func (c *char) Burst(p map[string]int) (action.Info, error) {
 	c.AddStatus(burstKey, ajawDuration, false)
 	if c.nightsoulState.HasBlessing() {
 		// extend Nightsoul's Blessing time limit countdown
-		duration := (c.exitStateF - c.Core.F) + 1.7*60
-		c.setNightsoulExitTimer(duration)
+		duration := c.StatusDuration(nightsoul.NightsoulBlessingStatus)
```
imring, line 37: it's better to create a method `BlessingDuration` in NS.

imring:
what about xilonen?

jwalk-511:
> what about xilonen?

I've added more functionality to Nightsoul to enable duration extension on specified animation states. If a NS would expire naturally during a restricted animation state, and a cb function is given, the Nightsoul state will 
 - grant the character a permanent NS duration
 - subscribe to onEventChanged
 - Remove the NS status and call the on NS Expiry CB when the EventChange is triggered
 
 If a player uses a timed NS duration and has no on expiry CB, and they expect an animation to extend the timed NS state, then that character will have to manually extend NS themselves. Otherwise, the nightsoul module can handle extension.
 
 I have not modified other behavior- if a character wants to remove NS status before the duration is over (e.g. on points expiry), then the character must still manually check animation state.
 
 It would be possible to add additional functionality to the NS module to handle an automatic expiry of the NS state upon points reaching 0, along with extending NS status until an animation is over, but that would be beyond the scope of this PR.
 
 Also created a function within the nightsoul module to return remaining NS duration.

internal/template/nightsoul/nightsoul.go
```diff
@@ -2,40 +2,113 @@ package nightsoul
...
+	}
+
+	src := s.c.F + duration
+	s.ExitStateF = src
+	s.char.QueueCharTask(func() {
+		if s.ExitStateF != src {
+			return
+		}
+
+		if !slices.Contains(s.extendNsStates, s.c.Player.CurrentState()) {
```
imring, line 65: if the character is off-field?

jwalk-511: I think I relied on the fact that any character that has a timed blessing and can extend state (only Kinich and Xilo, that I know of), will lose NS upon char swap.

But I can add an extra check to try to future-proof

jwalk-511: Off-field check was added

PR #2334, opened 2025-02-10
Add Ororon high jump and high_plunge.

jwalk-511:
WIP
 
When Ororon uses a hold jump in-game, he initiates a form of NS transmission, leaping high in the air and gaining the NS Blessing status. This NS Blessing exists independently of the NS Blessing from his A1 passive but can still allow the extended Scroll of the Hero of the Cinder City 4pc effect to trigger. 

While airborne, he can fall, high plunge, or aim. Aim is not implemented here. Ororon cannot cancel his high jump before reaching max height, so low_plunge is also not implemented. To prevent implementation of movement tech, ororon is forcefully transitioned into the fall state on the frame he would open his glider.

When Ororon uses high jump, he drains 75 stamina. Stamina does not resume regeneration until he initiates another action. He does not require all 75 stamina to initiate a high jump and will drain to 0 if less than 75 stamina remains. When Ororon has drained all stamina, ~~he is unable to initiate a normal fall cancel and will only be able to plunge~~ Ororon can't cancel normally but can initiate an equivalent fall cancel by attempting to walk (glide) forward.

Implementation details:
Ororon's high jump code schedules a callback to a Player.exec() call as a proof of concept of forcefully changing the player's animation state. The concepts used should be applicable to any character who automatically executes an action outside of a GCSL statement, such as Mavuika Charge starting in the NA state and transitioning to the CA state, or Chasca transitioning from the charge state to a falling state upon NS expiry. This is not strictly needed for Ororon, and can be changed upon request. 

~~Extended functionality of implementation is still being tested.~~ Behavior for fake frame data has been validated in the added sample config.txt. Stamina behavior has not been validated.

GCSL notes:
jump:
param hold:
 - 0 (default): Initiates a normal jump, calls already-implemented code.
 - < 0: Sets hold to MaxHold. Fall through to below case.
 - \> 0: Subtract 1. Hold=1 implies minimum length hover before fall cancel. Fall will be initiated MinFallCancelFrames + Hold frames after the jump begins.

High_Plunge:
param fall:
- 0 (default) Perform a normal high-plunge.
- 1: Simulate a pressing the jump button after reaching max jump height, initiating free-fall. A plunge cannot be used after a fall has been initiated. 

Frames todo:
- [ ] All jump[hold!=0]->X frames.
- [ ] All high_plunge->X frames.
- [ ] All high_plunge[fall=1]->X frames.
- [ ] NS Transmission start and end relative to action frames. 
- [ ] Stam drain and allowable resume frames.
Other todo:
- [ ] Min required stamina to initiate high jump (Drains 75, but able to initiate with less than 75)
- [ ] High Plunge[fall=0]. Currently, does nothing.
- [ ] Stamina drain adjustment when stamina mods are in effect (e.g. anemo resonance)
- [ ] Update mode_gcsim.js with shortcuts for syntax highlighting (Does this need to be done for new Params?)
- [x] Update documentation

### JUMP
- [x] add hold parameter to JUMP action
- [x] Require <75 stam before using high jump. Only requires 1? 5? TBD
- [x] Automatically transition from JUMP[hold != 0] to FALL action when NS transmission expires, if user does not manually exit (through plunge or plunge[fall=1]) first)
- [x] Grant AirborneOroron 

### STAMINA
- [x] Remove 75 stamina.
- [x] Prevent stamina regen.
- [x] WONTDO: Stam check should actually happen a little after jump is initiated.

### NS BLESSING
- [x] Give NS Transmission blessing status while airborne
- [x] Overload get_status method to identify ororon as having Nightsoul Blessing state when either he has Nightsoul Blessing status OR when he has NS Transmission status
- [x] Remove NS Transmission status when NS transmission ends

### High PLUNGE
- [ ] Implement AirborneOroron plunge (code currently works, no attack sent out, no collision done).
- [x] Implement AirborneOroron plunge[fall!=0] param
- [x] Require AirborneOroron status to use plunge

### STAMINA
- [x] Allow stamina to regen again after plunge
- [x] Allow stamina to regen again after plunge[fall=1]

### NS BLESSING
- [x] Remove NS Transmission on plunge // Assume it happens immediately
- [x] Remove NS Transmission on fall // Assume it happens immediately

Notes: Stamina is probably terribly bugged and also overcomplicated. My ability to accurately test stamina drain and resume timings is hampered by https://library.keqingmains.com/evidence/combat-mechanics/damage/other/client-and-server#high-ping-interactions

### DIFF
--- a/internal/artifacts/obsidiancodex/obsidiancodex.go
+++ b/internal/artifacts/obsidiancodex/obsidiancodex.go
@@ -36,7 +36,8 @@ func NewSet(c *core.Core, char *character.CharWrapper, count int, param map[stri
 		char.AddStatMod(character.StatMod{
 			Base: modifier.NewBase("obsidiancodex-2pc", -1),
 			Amount: func() ([]float64, bool) {
-				if !char.StatusIsActive(nightsoul.NightsoulBlessingStatus) {
+				if !(char.StatusIsActive(nightsoul.NightsoulBlessingStatus) ||
+					char.StatusIsActive(nightsoul.NightsoulTransmissionStatus)) {
 					return nil, false
 				}
 				if c.Player.Active() != char.Index {
--- a/internal/characters/ororon/jump.go
+++ b/internal/characters/ororon/jump.go
@@ -55,6 +55,7 @@ func (c *char) highJump(hold int) (action.Info, error) {
 		}
 
 		fallParam := map[string]int{"fall": 1}
+		// Ideally this would inject the action into the queue of actions to take from the config file, rather than calling exec directly
 		c.Core.Player.Exec(action.ActionHighPlunge, c.Base.Key, fallParam)
 	}
 	// TODO: Is this hitlag extended? Does this skip if the action is canceled?
--- a/internal/characters/ororon/jump.go
+++ b/internal/characters/ororon/jump.go
@@ -1,7 +1,12 @@
 package ororon
 
 import (
+	"fmt"
+
+	"github.com/genshinsim/gcsim/internal/frames"
 	"github.com/genshinsim/gcsim/pkg/core/action"
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
 	"github.com/genshinsim/gcsim/pkg/core/player"
 )
 
@@ -11,17 +16,47 @@ const jumpNsStatusTag = "Ororon-Status-NS-Jump"
 // Set a CB to cancel high jump if max duration exceeded.
 // Grant airborne status.
 // Consume stamina.
+// Hold defines when fall action will automatically be called.
 func (c *char) highJump(hold int) (action.Info, error) {
-	if hold > maxJumpFrames-minCancelFrames {
+	if (hold > maxJumpFrames-minCancelFrames) || (hold < 0) {
 		hold = maxJumpFrames - minCancelFrames
 	}
+
+	// If player runs out of stamina, delay fall. Still allow high_plunge.
+	minFallFramesAdjust := 0
+	if c.Core.Player.Stam <= jumpStamDrainAmt {
+		minFallFramesAdjust = jumpNoStamFallDelayFrames
+		c.Core.Log.NewEvent(fmt.Sprintf("High jump has consumed all stamina. Earliest fall will be delayed by %d frames.", minFallFramesAdjust), glog.LogCooldownEvent, c.Index)
+	}
+	// Earliest user can trigger fall from GCSL
+	c.allowFallFrame = c.Core.F + minCancelFrames + minFallFramesAdjust
+
+	if hold < minFallFramesAdjust {
+		hold = minFallFramesAdjust
+	}
+
 	jumpDur := minCancelFrames + hold
 	c.jmpSrc = c.Core.F
 	src := c.jmpSrc
 	c.Core.Player.SetAirborne(player.AirborneOroron)
 
 	c.QueueCharTask(func() { c.AddStatus(jumpNsStatusTag, jumpNsDuration, true) }, jumpNsDelay)
-	c.QueueCharTask(func() { c.Core.Player.Stam -= jumpStamDrainAmt }, jumpStamDrainDelay)
+
+	jumpStamDrainCb := func() {
+		h := c.Core.Player
+		h.Stam -= jumpStamDrainAmt
+		// While in high jump, ororon cannot start resuming stamina regen until after landing.
+		h.LastStamUse = *h.F + jumpDur + fallFrames - player.StamCDFrames
+		h.Events.Emit(event.OnStamUse, action.ActionJump)
+	}
+	c.QueueCharTask(jumpStamDrainCb, jumpStamDrainDelay)
+
+	act := action.Info{
+		Frames:          frames.NewAbilFunc(jumpHoldFrames[0]),
+		AnimationLength: jumpDur + fallFrames,
+		CanQueueAfter:   minCancelFrames, // earliest cancel
+		State:           action.JumpState,
+	}
 
 	// Trigger a fall after max jump duration
 	fallCb := func() {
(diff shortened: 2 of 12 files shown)


internal/characters/ororon/testconfig.txt
```diff
@@ -0,0 +1,141 @@
+# Team
```
imring, line 1: remove

internal/artifacts/obsidiancodex/obsidiancodex.go
```diff
@@ -36,7 +36,8 @@ func NewSet(c *core.Core, char *character.CharWrapper, count int, param map[stri
 		char.AddStatMod(character.StatMod{
 			Base: modifier.NewBase("obsidiancodex-2pc", -1),
 			Amount: func() ([]float64, bool) {
-				if !char.StatusIsActive(nightsoul.NightsoulBlessingStatus) {
+				if !(char.StatusIsActive(nightsoul.NightsoulBlessingStatus) ||
+					char.StatusIsActive(nightsoul.NightsoulTransmissionStatus)) {
```
imring, line 40: looks bad to me
i mean i'd like to get the NS state from the interface and call `HasBlessing`, but to do that i need to create a method like `GetCustomField(enum)`

internal/characters/ororon/jump.go
```diff
@@ -0,0 +1,80 @@
...
+	src := c.jmpSrc
+	c.Core.Player.SetAirborne(player.AirborneOroron)
+
+	// Don't add NS if jump is cancelled before NS would be added.
+	jumpNsDuration := jumpDur - jumpNsDelay
+	c.QueueCharTask(func() {
+		if src != c.jmpSrc {
+			return
+		}
+		c.AddStatus(jumpNsStatusTag, jumpNsDuration, true)
```
imring, line 34: why is the status added outside of `template/nightsoul.go`?

internal/artifacts/obsidiancodex/obsidiancodex.go
jwalk-511: I'd agree, but artifacts don't get to see the interface at all, only the struct `*character.CharWrapper`.
My initial idea was, if artifacts could see the character interface I'd have nightsoul characters overload the `.StatusIsActive` function to `return Nightsoul.HasBlessing()` when `key==nightsoul.NightsoulBlessingStatus` and `return CharWrapper.StatusIsActive(key)` otherwise.

But I thought such a refactor was beyond the scope of this PR.
If desired, I can put this PR on hold and make that refactor (and also add all methods from CharWrapper to the Character Interface), or add GetCustomField(enum) to the CharWrapper struct, in another PR.

internal/characters/ororon/testconfig.txt
jwalk-511: > remove

Done in latest commit

internal/characters/ororon/jump.go
jwalk-511: > why is the status added outside of `template/nightsoul.go`?

Refactored in latest commit.

jwalk-511: No longer relevant with refactor to remove independent NS transmission status.

internal/characters/ororon/plunge.go
```diff
@@ -0,0 +1,155 @@
...
+	// Plunge -> X
+	plungeFrames = frames.InitAbilSlice(66) // Default is From plunge animation start to swap icon un-gray
+	plungeFrames[action.ActionAttack] = 68
+	plungeFrames[action.ActionAim] = 66
+	plungeFrames[action.ActionSkill] = 65
+	plungeFrames[action.ActionBurst] = 65
+	plungeFrames[action.ActionDash] = 53
+	plungeFrames[action.ActionJump] = 82
+	plungeFrames[action.ActionWalk] = 80
+	plungeFrames[action.ActionSwap] = 66
```
imring, line 37: source?

internal/characters/ororon/ororon.go
```diff
@@ -1,27 +1,66 @@
...
+	// Fall -> X
+	jumpHoldFrames[1] = frames.InitAbilSlice(fallFrames)
+	jumpHoldFrames[1][action.ActionAttack] = 45
+	jumpHoldFrames[1][action.ActionAim] = 46
+	jumpHoldFrames[1][action.ActionSkill] = 45
+	jumpHoldFrames[1][action.ActionBurst] = 46
+	jumpHoldFrames[1][action.ActionDash] = 46
+	jumpHoldFrames[1][action.ActionJump] = 45
+	jumpHoldFrames[1][action.ActionWalk] = 47
+	jumpHoldFrames[1][action.ActionSwap] = 44
```
imring, line 50: move jump frames to jump.go

imring: frames source?

jwalk-511: Frames source for fall->x and for low_plunge->x: https://docs.google.com/spreadsheets/d/1ru_C_1vh_0k3u-NdJFcq079N_vstn055PVsL66eCoN8/edit?gid=2018562987#gid=2018562987

jumpHoldFrames[1] corresponds to fall frames, which is used by both low_plunge[fall=1] and by high_jump, when the jump is allowed to expire naturally. I left it in ororon.go to make it more obvious it wasn't only for jump.go

I'll move frames to jump.go anyway

internal/characters/ororon/jump.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		c.nightsoulState.EnterBlessing(c.nightsoulState.Points())
+		c.inTransmissionBlessing = true
+	}, jumpNsDelay)
+
+	// Consume stamina.
+	c.QueueCharTask(func() {
+		h := c.Core.Player
+		// Apply stamina reduction mods.
+		stamDrain := h.AbilStamCost(c.Index, action.ActionJump, map[string]int{"hold": 1})
```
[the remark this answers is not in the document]
jwalk-511: That's true; I was trying to make the function just a little bit more lightweight by only passing the hold integer as a parameter, rather than the entire p map. But the fact that I have to re-construct the map and pass it to h.AbilStamCost probably cancels that out anyway

internal/characters/ororon/ororon.go
jwalk-511: I don't think anyone's verified the frames I counted yet.
Once they are verified, I think they should be added to the main oro frames sheet, and then I can link that within the comments in the code

internal/characters/ororon/plunge.go
jwalk-511: Source shared in ororon.go review comment

PR #2344, opened 2025-02-19
Fix Nilou A4

jwalk-511:
Added a custom additional tag to detect Nilou bloom, rather than try to fix gadgets.
Added 2 new public functions in reactables. Only 1 needs to be public, but wasn't sure which was preferred. If desired, I can remove the nilou-specific func entirely and only leave the generic tagging function.


shizukayuki:
fixed in #2345

PR #2372, opened 2025-03-28
add long nights oath

imring:
- [x] New artifact package
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation


internal/artifacts/longnightsoath/longnightsoath.go
```diff
@@ -0,0 +1,110 @@
...
+			Base: modifier.NewBase("longnightsoath-2pc", -1),
+			Amount: func(atk *combat.AttackEvent, t combat.Target) ([]float64, bool) {
+				if atk.Info.AttackTag != attacks.AttackTagPlunge {
+					return nil, false
+				}
+				return m, true
+			},
+		})
+	}
+	if count >= 4 {
```
Charlie-Zheng, line 65: you could do 
```go
if count < 4 {
  return &s, nil
}
[...]
```

PR #2374, opened 2025-03-28
Add Iansan

imring:
- [x] New character package
- [x] Run pipeline with added config (generates character curve, talent stats, `.dm.go` files)
- [x] Character key
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation


ErrorRaffyline0:
I think I speak for everyone when I say it is expected that any maintainer working on this PR would give up early on, when being hit with the realization that 90% of Iansan's kit relies upon implementing movement values for hundreds of character animations. But thanks for trying at least. ❤

internal/characters/iansan/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		c.burstBuff[attributes.ATK] = min(c.TotalAtk()*rate, maxATK[c.TalentLvlBurst()])
+
+		c.QueueCharTask(c.updateATKBuff(src), 0.3*60)
+	}
+}
+
+func (c *char) burstMovementRestore(args ...interface{}) {
+	if !c.StatusIsActive(burstStatus) {
+		c.Core.Events.Unsubscribe(event.OnActionExec, burstBuffStatus)
```
shizukayuki: shouldn't be doing this really, just have it always subscribed and return early.

internal/characters/iansan/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		if !c.StatusIsActive(burstStatus) {
+			c.burstBuff[attributes.ATK] = 0
+			return
+		}
+
+		rate := highATK
+		if c.nightsoulState.Points() < 42 {
+			rate = lowATK * c.nightsoulState.Points()
+		}
+		c.burstBuff[attributes.ATK] = min(c.TotalAtk()*rate, maxATK[c.TalentLvlBurst()])
```
shizukayuki: this needs to update on swap too, independent of the 0.3s ticker

internal/characters/iansan/asc.go
```diff
@@ -0,0 +1,111 @@
...
+		c.a1Increase = true
+	}
+	c.Core.Events.Subscribe(event.OnNightsoulGenerate, cb, "iansan-a1-generate")
+	c.Core.Events.Subscribe(event.OnNightsoulConsume, cb, "iansan-a1-consume")
+}
+
+func (c *char) a1ATK() {
+	m := make([]float64, attributes.EndStatType)
+	m[attributes.ATKP] = 0.2
+	c.AddStatMod(character.StatMod{
```
shizukayuki: this will allocate on every call, so move the slice to char struct.

internal/characters/iansan/attack.go
```diff
@@ -0,0 +1,90 @@
...
+	attackFrames[1][action.ActionCharge] = 25
+
+	attackFrames[2] = frames.InitNormalCancelSlice(attackHitmarks[2], 55) // N3 -> Walk
+	attackFrames[2][action.ActionAttack] = 53
+}
+
+func (c *char) Attack(p map[string]int) (action.Info, error) {
+	if c.StatusIsActive(fastSkill) {
+		// TODO: or c.Core.Player.Exec(action.ActionCharge, c.Base.Key, nil)
+		return c.ChargeAttack(p)
```
shizukayuki, line 44: can we throw an error here?

pkg/core/attacks/attack.go
```diff
@@ -20,6 +20,7 @@ const (
 	AttackTagNightsoul
 	AttackTagKinichCannon
 	AttackTagVarkaSpecial
+	AttackTagIansanBisonsaurus
```
shizukayuki: remove if nothing checks for it.

internal/characters/iansan/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		} else {
+			c.nightsoulState.GeneratePoints(15)
+		}
+	}, 40)
+
+	c.burstSrc = c.Core.F
+	c.burstRestoreNS = 0
+	c.updateATKBuff(c.burstSrc)()
+	c.applyBuffTask(c.burstSrc)
+	c.Core.Events.Subscribe(event.OnActionExec, c.burstMovementRestore, burstBuffStatus)
```
shizukayuki: do it once at init.

internal/characters/iansan/attack.go
imring: by using c.Core.Player.Exec? yes, but we need to make a custom action.Info, not from c.ChargeAttack

internal/characters/iansan/burst.go
```diff
@@ -0,0 +1,159 @@
...
+				c.c2ATKBuff(active)
+				return c.burstBuff
+			},
+		})
+
+		c.QueueCharTask(c.updateATKBuff(src), 0.3*60)
+	}
+}
+
+func (c *char) burstMovementRestore(args ...interface{}) {
```
jwalk-511: Iansan needs to filter out self movement

jwalk-511 (changes requested):

internal/characters/iansan/burst.go
jwalk-511: Needs checking: If Iansan would otherwise restore 0 NS points, does she still try to restore points through A1/C4?
https://github.com/genshinsim/gcsim/pull/2374/changes#diff-14ee7e82c32ef9d049dcd07c4fd43476fde74705cd06cf7f8d4257608564539eR103

pkg/gcs/validation/validate.go
```diff
@@ -1,15 +1,31 @@
...
+	"slices"
+
 	"github.com/genshinsim/gcsim/pkg/core/action"
 	"github.com/genshinsim/gcsim/pkg/core/keys"
 )
 
+// generic params that can be used for any character
+var ignoreParams = []string{
+	// iansan burst
+	"movement",
```
jwalk-511: I think there should be some documentation on this param. It could be added as a documentation for parameters for Iansan's burst, with a note that it should actually be used as a param for other character actions.

internal/characters/iansan/attack.go
shizukayuki: current is fine, maybe just remove the comment.

quite few other chars do a similar thing of an action calling into another.

internal/characters/iansan/burst.go
```diff
@@ -90,6 +85,23 @@ func (c *char) Burst(p map[string]int) (action.Info, error) {
...
+
+	for _, char := range c.Core.Player.Chars() {
+		char.AddStatMod(character.StatMod{
+			Base:  modifier.NewBaseWithHitlag(burstBuffStatus, -1),
+			Extra: true,
+			Amount: func() []float64 {
+				if !c.StatusIsActive(burstStatus) {
+					return nil
+				}
+				return c.burstBuff
```
shizukayuki, line 100: missing active char check

internal/characters/iansan/cons.go
```diff
@@ -9,44 +9,78 @@ import (
...
+		}
+
+		if prev != c.Index() {
+			c.Core.Player.Chars()[prev].DeleteStatMod(c2Key)
+		}
+
+		if next != c.Index() {
+			// we don't need to worry about granting this buff when iansan gains her A1, because she
+			// will always be the active character (using skill or using burst)
+			c.Core.Player.Chars()[next].AddStatMod(character.StatMod{
```
imring, line 61: use c.Core.Player.ByIndex(next)

internal/characters/iansan/burst.go
Charlie-Zheng: Based on https://discord.com/channels/845087716541595668/983391844631212112/1550170161137451110
https://www.youtube.com/watch?v=lcTahzPrb5E
It shows that A1 still applies if Iansan is on field

pkg/gcs/validation/validate.go
Charlie-Zheng: It's included in the burst docs

PR #2404, opened 2025-05-16
Add Finale of the Deep Galleries

yihuajack:
- [x] New artifact package
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation

What we know now is:
- Mavuika's energy is 0 so is able to trigger 4pc
- NA can consistently refresh the stop-playing cd of burst


internal/artifacts/finaleofthedeepgalleries/finaleofthedeepgalleries.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type Set struct {
+	Index int
+	Count int
+
+	c                 *core.Core
+	char              *character.CharWrapper
+	procNormalExpireF int
+	procBurstExpireF  int
```
imring, line 29: i'd prefer to use statuses

yihuajack: Updated

imring (approved):

PR #2406, opened 2025-05-19
add .element.t0.burning field

soloxcx:
example config
```
bennett char lvl=90/90 cons=6 talent=9,9,9;
bennett add weapon="umbrella" refine=5 lvl=90/90;

nahida char lvl=90/90 cons=0 talent=9,9,9; 
nahida add weapon="evenstar" refine=1 lvl=90/90;

xingqiu char lvl=90/90 cons=6 talent=9,9,9;
xingqiu add weapon="favsword" refine=3 lvl=90/90;

options iteration=1000;
energy every interval=480,720 amount=1;

target lvl=100 resist=0.1 radius=2 pos=0,2.4 hp=10000000;

active nahida;

nahida attack;
bennett skill;

while .element.t0.burning {
  bennett attack;
}

xingqiu skill;
wait(120);
```

### DIFF
--- a/pkg/reactable/reactable.go
+++ b/pkg/reactable/reactable.go
@@ -46,6 +46,7 @@ var modifierElement = []attributes.Element{
 	attributes.Frozen,
 	attributes.Anemo,
 	attributes.Geo,
+	attributes.Pyro,
 	attributes.UnknownElement,
 }
 
@@ -62,6 +63,7 @@ var ModifierString = []string{
 	"frozen",
 	"anemo",
 	"geo",
+	"burning",
 	"",
 }
 


pkg/core/attributes/element.go
```diff
@@ -22,6 +22,7 @@ const (
 	Geo
 	NoElement
 	Physical
+	Burning
```
imring, line 25: instead of adding `Burning` element make a condition `ele == "burning"` in `evalElement` (https://github.com/genshinsim/gcsim/blob/main/pkg/conditional/enemy.go) and use `IsBurning` (example: https://github.com/kengzzzz/gcsim/blob/main/internal/artifacts/unfinishedreverie/unfinishedreverie.go)

imring (changes requested):

imring (approved):

PR #2407, opened 2025-05-19
start to check hutao c6 after hp set

imring:


Charlie-Zheng:
Maybe we should start the check at a random delay after the sim starts?

Charlie-Zheng:
NVM I took a closer look at the code and I see why it starts at frame 1

PR #2450, opened 2025-08-29
Fixed Ororon c6 buff application

Charlie-Zheng:
https://discord.com/channels/845087716541595668/1410785446237175818/1410785446237175818

> The buff is only for the current active unit. The sim has the buff persist for the unit that swapped off and needs Ororon’s passive to proc for it to be applied to the new onfielder.
 https://gcsim.app/db/MjnNK8gfdPnL


Charlie-Zheng:
Hmmm is the buffer hitlag affected per character? Or is it hitlag affected on Ororon only?

imring (approved):

PR #2497, opened 2025-10-09
Add aura ownership tracking to reactable

Charlie-Zheng:
This PR refactors Reactable to prepare for Lunar Charged implementation, which needs to know which characters currently have aura on the enemy.

### DIFF
--- a/pkg/reactable/reactable.go
+++ b/pkg/reactable/reactable.go
@@ -264,6 +264,10 @@ func (r *Reactable) IsBurning() bool {
 	return false
 }
 
+func (r *Reactable) reduceMod(mod info.ReactionModKey, amt info.Durability) {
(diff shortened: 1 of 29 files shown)


pkg/reactable/reactable.go
```diff
@@ -10,8 +11,10 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/info"
 )
 
+const maxChars = 4
+
 type Reactable struct {
-	Durability [info.ReactionModKeyEnd]info.Durability
+	Durability [info.ReactionModKeyEnd][maxChars]info.Durability
```
imring: move `maxChars` to `pkg/core/info` (maybe character.go?) package with 4 in `SetupCharactersInCore` or change the type for Durability to `map[int][info.ReactionModKeyEnd]info.Durability`

PR #2501, opened 2025-10-11
Add Dahlia

JohnnyToumieh:
- [x] New character package
- [x] Run pipeline with added config (generates character curve, talent stats, .generated.json files)
- [x] Character key
- [x] Update mode_gcsim.js with shortcuts for syntax highlighting
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation

### DIFF
--- a/internal/characters/dahlia/burst.go
+++ b/internal/characters/dahlia/burst.go
@@ -23,8 +23,7 @@ const (
 	burstEnergyDrain             = 8
 	burstCDStart                 = 0
 
-	benisonMaxGenerate  = 4
-	maxFavonianFavorExt = 5 * 60 // Checked using AS buffs and on-field Dahlia full NA string
+	benisonMaxGenerate = 4
 
 	burstFavonianFavor    = "dahlia-favonian-favor"
 	burstBenisonStacksKey = "dahlia-benison-stacks"
@@ -73,7 +72,6 @@ func (c *char) Burst(p map[string]int) (action.Info, error) {
 		// Increasing Favonisn Favor duration if C4+
 		// It lasts for exactly 15s (pre-hitlag), so shield + Attack Speed remain after Q CD is over
 		favonianFavorDuration := 12*60 + c.c4FavonianFavorBonusDur()
-		c.favonianFavorMaxExpiry = c.Core.F + favonianFavorDuration + maxFavonianFavorExt
 
 		// Add "Favonian Favor" status to Dahlia
 		// (team members should technically get it but only he can extend it with hitlag)
@@ -152,6 +150,6 @@ func (c *char) addBenisonStack(stacks, charIndex int) {
 
 	// If shield is already gone but new stacks got generated, create shield (after some delay)
 	if !c.hasShield() {
-		c.tryRegenShield()
+		c.tryRegenShield(burstShieldAfterBenisonStack)
 	}
 }
--- a/internal/characters/dahlia/dahlia.go
+++ b/internal/characters/dahlia/dahlia.go
@@ -21,8 +21,6 @@ type char struct {
 	normalAttackCount    int
 
 	attackSpeedBuff []float64
-
-	favonianFavorMaxExpiry int // TO-DO: used for shield since they don't support hitlag, but the value is wrong
 }
 
 func NewChar(s *core.Core, w *character.CharWrapper, _ info.CharacterProfile) error {
--- a/internal/characters/dahlia/shield.go
+++ b/internal/characters/dahlia/shield.go
@@ -42,7 +42,7 @@ func (s *shd) OnDamage(dmg float64, ele attributes.Element, bonus float64) (floa
 	taken, ok := s.Tmpl.OnDamage(dmg, ele, bonus)
 
 	if !ok {
-		s.c.tryRegenShield()
+		s.c.tryRegenShield(burstShieldRegenerated)
 	}
 	return taken, ok
 }
@@ -55,16 +55,16 @@ func (c *char) shieldHP() float64 {
 	return burstShieldPP[c.TalentLvlBurst()]*c.MaxHP() + burstShieldFlat[c.TalentLvlBurst()]
 }
 
-func (c *char) tryRegenShield() {
+func (c *char) tryRegenShield(delay int) {
 	if c.currentBenisonStacks <= 0 {
 		return
 	}
-	if c.StatusDuration(burstFavonianFavor) < burstShieldRegenerated {
+	if c.StatusDuration(burstFavonianFavor) < delay {
 		return
 	}
 	c.Core.Tasks.Add(func() {
 		c.currentBenisonStacks--
 		c.genShield()
 		c.c2()
-	}, burstShieldRegenerated)
+	}, delay)
 }


internal/characters/dahlia/attack.go
```diff
@@ -0,0 +1,78 @@
...
+	attackFrames[1][action.ActionAttack] = 21                                // N2 -> N3
+	attackFrames[1][action.ActionCharge] = 21                                // N2 -> CA
+
+	attackFrames[2] = frames.InitNormalCancelSlice(attackHitmarks[2][1], 47) // N3 -> W
+	attackFrames[2][action.ActionAttack] = 39                                // N3 -> N4
+	attackFrames[2][action.ActionCharge] = 39                                // N3 -> CA
+
+	attackFrames[3] = frames.InitNormalCancelSlice(attackHitmarks[3][0], 62) // N4 -> W
+	attackFrames[3][action.ActionAttack] = 71                                // N4 -> N1
+	attackFrames[3][action.ActionCharge] = 500                               // TODO: this action is illegal; need better way to handle it
```
Charlie-Zheng, line 41: This doesn't actually work. The animation length is capped at 62. I'd recommend adding:

```
func (c *char) NextQueueItemIsValid(_ keys.Char, a action.Action, p map[string]int) error {
	if a == action.ActionCharge {
		if c.NormalCounter == 0 {
				return player.ErrInvalidChargeAction
		}
	}

	return nil
}
```

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Hydro,
+		Durability: 50,
+		Mult:       burst[c.TalentLvlBurst()],
+	}
+
+	c.Core.QueueAttack(
+		ai,
+		combat.NewCircleHitOnTarget(c.Core.Combat.Player(), info.Point{Y: 0.8}, 5.5),
+		0, // TODO: snapshot?
```
Charlie-Zheng: If we don't know the snapshot we usually assume it snapshots on the frame it hits, so it should snapshot at `burstHitmark` until further testing determines a more specific timing

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+	c.SetCDWithDelay(action.ActionBurst, 15*60, burstCDStart) // Delay is actually 0 but left for clarity
+	c.ConsumeEnergy(burstEnergyDrain)
+
+	// After a short delay, add the "Favonian Favor" status and all its effects
+	c.Core.Tasks.Add(func() {
+		// Increasing shield duration if C4+ (it lasts for exactly 15s, so shield remains after Q CD is over)
+		favonianFavorDuration := 12 * 60
+		if c.Base.Cons >= 4 {
+			favonianFavorDuration = 15 * 60
+		}
```
Charlie-Zheng: Nitpick: I perfer it if it was a function in cons.go
```
func (c *char) c4ShieldBonusDur() int {
  if c.Base.Cons < 4 {
    return 0
  }
  return 3 * 60
}
```

and then here you would do
```
favonianFavorDuration := 12 * 60 + c.c4ShieldBonusDur()
```

I personally like being able to see all the cons effects in cons.go

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		c.favonianFavorExpiry = c.Core.F + favonianFavorDuration
+
+		// Add "Favonian Favor" status for all party members
+		for _, char := range c.Core.Player.Chars() {
+			char.AddStatus(burstFavonianFavor, favonianFavorDuration, true)
+		}
```
Charlie-Zheng: Is this actually hitlag affect per character? Or does it only get hitlag extended by hitlag on Dahlia? 
If it's unknown, just add a `// TODO: Is this hitlag extended per character or only by hitlag on Dahlia`

internal/characters/dahlia/asc.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return false
+}
+
+func (c *char) addAttackSpeedbuff(char *character.CharWrapper) {
+	char.AddStatMod(character.StatMod{
+		Base:         modifier.NewBaseWithHitlag(attackSpeedKey, c.favonianFavorExpiry-c.Core.F),
+		AffectedStat: attributes.AtkSpd,
+		Amount: func() ([]float64, bool) {
+			return c.attackSpeedBuff, true
```
Charlie-Zheng, line 94: You can do
```
if c.Core.Player.Active() != char.Index() {
  return nil, false
}
return c.attackSpeedBuff, true
```
instead of deleting and adding the buff on swap

internal/characters/dahlia/asc.go
```diff
@@ -0,0 +1,115 @@
...
+		if burstAttackSpeed < 0 {
+			burstAttackSpeed = 0
+		} else if burstAttackSpeed > 0.2 {
+			burstAttackSpeed = 0.2
+		}
+
+		// If C6, add another 10% Attack Speed
+		if c.Base.Cons >= 6 {
+			burstAttackSpeed += 0.1
+		}
```
Charlie-Zheng: Nitpick: I perfer it if it was a function in cons.go
```
func (c *char) c6AtkSpd() int {
  if c.Base.Cons < 6 {
    return 0.0
  }
  return 0.1
}
```
and then here you would do

```
burstAttackSpeed += c.c6AtkSpd()
```

I personally like being able to see all the cons effects in cons.go

internal/characters/dahlia/asc.go
```diff
@@ -0,0 +1,115 @@
...
+		if c.a4Src != src {
+			return
+		}
+
+		burstAttackSpeed := c.MaxHP() * 0.001 * 0.005
+		if burstAttackSpeed < 0 {
+			burstAttackSpeed = 0
+		} else if burstAttackSpeed > 0.2 {
+			burstAttackSpeed = 0.2
+		}
```
Charlie-Zheng: Nitpick: We can assume MaxHP is positive, so this can be simplified down to:

```
burstAttackSpeed := min(c.MaxHP() * 0.001 * 0.005, 0.2)
```

internal/characters/dahlia/asc.go
Charlie-Zheng, line 94: This current give AtkSpd to CAs as well I believe. There needs to be an animation state check I think
```
if c.Core.Player.CurrentState() != action.NormalAttackState {
	return nil, false
}
```
However there are some intricacies with when animation state transitions happen and when AtkSpd is checked so be sure to double check if this change is good with someone

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+	stacks = min(stacks, c.benisonGenStackLimit)
+	c.benisonGenStackLimit -= stacks
+	c.currentBenisonStacks += stacks
+
+	// If C1, Dahlia restores 2.5 Energy per stack gained
+	if c.Base.Cons >= 1 {
+		c.AddEnergy(c1Key, 2.5*float64(stacks))
+	}
+
+	c.Core.Log.NewEvent(strconv.Itoa(stacks)+" "+burstBenisonStacksKey+" added", glog.LogShieldEvent, charIndex).
```
Charlie-Zheng: Use `fmt.Sprintf("%v %v added, stacks, burstBenisonStacksKey)` instead of `strconv.Itoa`

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		Write("benison_stacks_remaining", c.currentBenisonStacks).
+		Write("benison_stacks_generated", benisonMaxGenerate-c.benisonGenStackLimit)
+
+	// If shield is already gone but new stacks got generated, create shield (after some delay)
+	if !c.hasShield() && c.favonianFavorExpiry > c.Core.F+burstShieldAfterBenisonStack {
+		c.QueueCharTask(func() {
+			c.genShield()
```
Charlie-Zheng: This should consume a benison stack `c.currentBenisonStacks--`

internal/characters/dahlia/burst.go
```diff
@@ -0,0 +1,161 @@
...
+		return
+	}
+	stacks = min(stacks, c.benisonGenStackLimit)
+	c.benisonGenStackLimit -= stacks
+	c.currentBenisonStacks += stacks
+
+	// If C1, Dahlia restores 2.5 Energy per stack gained
+	if c.Base.Cons >= 1 {
+		c.AddEnergy(c1Key, 2.5*float64(stacks))
+	}
```
Charlie-Zheng: Nitpick: I perfer it if it was a function in cons.go
```
func (c *char) c1OnBenisonEnergy(stacks int) {
  if c.Base.Cons < 1 {
    return 
  }
  c.AddEnergy(c1Key, 2.5*float64(stacks))
}
```
and then here you would do
```
c.c1OnBenisonEnergy(stacks)
```
I personally like being able to see all the cons effects in cons.go

internal/characters/dahlia/skill.go
```diff
@@ -0,0 +1,90 @@
...
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Hydro,
+		Durability: 25,
+		Mult:       skill[c.TalentLvlSkill()],
+	}
+
+	c.Core.QueueAttack(
+		ai,
+		combat.NewCircleHitOnTarget(c.Core.Combat.Player(), nil, 3),
+		0, // TO-DO: Should this be skillHitmarks[hold] instead?
```
Charlie-Zheng: Generally we assume things snapshot at hitmark until otherwise indicated

ui/packages/docs/src/components/Actions/character_data.json
```diff
@@ -727,6 +727,48 @@
...
+      "legal": true
+    },
+    {
+      "ability": "low_plunge",
+      "notes": "Previous action must be a jump buffed via Xianyun's burst for example."
+    },
+    {
+      "ability": "high_plunge",
+      "notes": "Previous action must be a jump buffed via Xianyun's burst for example."
+    },
```
Charlie-Zheng: These aren't implemented

ui/packages/docs/src/components/Params/character_data.json
```diff
@@ -337,6 +337,23 @@
...
+    {
+      "ability": "low_plunge",
+      "param": "collision",
+      "desc": "0 for no collision dmg (default), 1 for collision dmg."
+    },
+    {
+      "ability": "high_plunge",
+      "param": "collision",
+      "desc": "0 for no collision dmg (default), 1 for collision dmg."
+    },
```
Charlie-Zheng: These aren't implemented

internal/characters/dahlia/config.yml
```diff
@@ -0,0 +1,39 @@
+package_name: dahlia
+genshin_id: 10000115
+key: dahlia
+action_param_keys:
+  low_plunge:
+    - param: "collision"
+  high_plunge:
+    - param: "collision"
```
Charlie-Zheng: His plunge isn't implemented. This is missing the hold param for his skill
```
  skill:
    - param: "short_hold"
 ```

File: internal/characters/dahlia/skill.go [hunk 15]
```diff
@@ -0,0 +1,90 @@
...
+	skillFrames[1][action.ActionBurst] = 35   // hE -> Q
+	skillFrames[1][action.ActionSkill] = 37   // hE -> tE / hE (TO-DO: check if this includes both)
+	skillFrames[1][action.ActionDash] = 33    // hE -> D
+	skillFrames[1][action.ActionJump] = 32    // hE -> J
+	skillFrames[1][action.ActionSwap] = 33    // hE -> Swap
+}
+
+func (c *char) Skill(p map[string]int) (action.Info, error) {
+	hold := 0
+	if p["short_hold"] != 0 {
```
Charlie-Zheng, line 46: Why "short_hold" instead of the usual "hold" param?

Charlie-Zheng: On second thought this is fine, since the short_hold param only changes the frames and not the mechanics of the E

PR #2519, opened 2025-10-20
Added the ability to miss for Zhongli's pillar

JohnnyToumieh:
Note: mostly relevant for ERC, guide making, etc...

Current issue to discuss:
- tE refreshes the pillar location
- hE does not

So let's say someone does:
```
neuvillette things
zhongli[hold=1]
neuvillette things
furina things
zhongli[hold=1, miss=100]
neuvillette things
```
Currently since the 2nd hold doesn't create a new pillar, that miss rate does not matter. You'd need to wait 30s for the pillar to expire and whichever zhongli[hold=1, miss=???] is next (with 0 or 100 missrate), would apply to the next pillar created.

Is that fine? Or would the expected behavior for the miss rate to update.
When i think about Zhongli's C1, it can be implemented without messing with it but idt it would be pretty.

---

Slightly unrelated, but I also had 2 more new field ideas:
- An increasing miss rate. So over said 30s duration, the miss rate goes up with each tick so later ticks are more likely to miss due to enemies moving away from said pillar. It could also semi-simulate that pillar getting destroyed later.
- Frame where the stele gets destroyed. So not instant destruction like hold_nostele but delayed.

If I don't get any No(s) for the 2nd idea, i can easily implement that but i'm separating the PRs.

### DIFF
--- a/internal/characters/zhongli/stele.go
+++ b/internal/characters/zhongli/stele.go
@@ -11,7 +11,7 @@ import (
 
 const particleICDKey = "zhongli-particle-icd"
 
-func (c *char) newStele(dur int, missRate int) {
+func (c *char) newStele(dur, missRate int) {
 	flat := c.a4Skill()
 	// deal damage when created
 	ai := info.AttackInfo{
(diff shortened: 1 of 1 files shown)


JohnnyToumieh:
https://discord.com/channels/845087716541595668/1379686559577669713

PR #2533, opened 2025-11-15
Add Sacrificer's Staff

Watchfuls:
Weapon PR checklist:
- [x] New weapon package
- [x] Run pipeline with added config (generates weapon curve, .generated.json files)
- [x] Weapon key
- [x] Update documentation

I was told to open a PR first and then go through the checklist so I'll do that!


internal/weapons/spear/sacrificersstaff/sacrificersstaff.go
```diff
@@ -0,0 +1,67 @@
...
+			return false
+		}
+		if atk.Info.AttackTag != attacks.AttackTagElementalArt && atk.Info.AttackTag != attacks.AttackTagElementalArtHold {
+			return false
+		}
+
+		if !char.StatModIsActive(buffKey) {
+			w.stacks = 0
+		}
+		w.stacks = max(w.stacks+1, 3)
```
aurceive: There's a bug here. It should be like this:
```go
w.stacks = min(w.stacks+1, 3)
```

PR #2535, opened 2025-12-10
allow 100 lvl for characters & updated lvl curves

imring:


imring:
> I think you still need to add the level scaling for reactions

added

Charlie-Zheng:
Do we want to restrict "impossible" levels? E.G. 97/100, or invalid max levels (95/100, 90/100, 110/110)

PR #2536, opened 2025-12-11
Implement lunar-bloom

Matoba004:


internal/artifacts/scrolloftheheroofcindercity/scrolloftheheroofcindercity.go
Charlie-Zheng, line 1: I think this is missing the mapping from ReactionTypeLunarBloom to Elements in the `reactToElements` map

internal/characters/baizhu/asc.go
```diff
@@ -71,6 +71,8 @@ func (c *char) a4() {
 			case attacks.AttackTagHyperbloom:
 			case attacks.AttackTagBurgeon:
 			case attacks.AttackTagBurningDamage:
+			case attacks.AttackTagDirectLunarBloom:
+				return limitHP * 0.007, false
 			default:
 				return 0, false
 			}
```
Charlie-Zheng, line 79: I think change to the follow for better flow
```golang  
mult := 0.02
switch ai.AttackTag {
case attacks.AttackTagBloom:
case attacks.AttackTagHyperbloom:
case attacks.AttackTagBurgeon:
case attacks.AttackTagBurningDamage:
case attacks.AttackTagDirectLunarBloom:
  mult := 0.07
default:
return 0, false
}
return limitHP * mult, false
```

internal/characters/nahida/cons.go
```diff
@@ -33,22 +33,27 @@ func (c *char) c2() {
...
 		}
 
-		if !t.StatusIsActive(skillMarkKey) {
-			return false
-		}
-
 		// TODO: should this really be +=??
 		ae.Snapshot.Stats[attributes.CR] += 0.2
-		ae.Snapshot.Stats[attributes.CD] += 1
+		ae.Snapshot.Stats[attributes.CD] = 1
```
Charlie-Zheng: Change to the follow for better flow. Also fixes a correctness issue. (Your code would buff lunarbloom by 0.3 CR but set the CD to 1.0)
```
		switch ae.Info.AttackTag {
		case attacks.AttackTagBurningDamage | attacks.AttackTagBloom | attacks.AttackTagHyperbloom | attacks.AttackTagBurgeon:
		    // TODO: should this really be +=??
		    ae.Snapshot.Stats[attributes.CR] += 0.2
		    ae.Snapshot.Stats[attributes.CD] = 1
        case attacks.AttackTagDirectLunarBloom:
			ae.Snapshot.Stats[attributes.CR] += 0.1
			ae.Snapshot.Stats[attributes.CD] += 0.2
		default:
			return false
		}

		c.Core.Log.NewEvent("nahida c2 buff", glog.LogCharacterEvent, ae.Info.ActorIndex).
				Write("final_crit", ae.Snapshot.Stats[attributes.CR])
        return fasle
```

pkg/core/attacks/icd.go
```diff
@@ -73,6 +74,7 @@ func init() {
 	ICDGroupResetTimer[ICDGroupBurning] = 120
 	ICDGroupResetTimer[ICDGroupTravelerDewdrop] = 90
 	ICDGroupResetTimer[ICDGroupTravelerBurst] = 480
+	ICDGroupResetTimer[ICDGroupDirectLunarBloom] = 90
```
Charlie-Zheng: Delete this, it isn't needed and Lauma/Nefer's attacks don't have this ICDGroup on them

pkg/core/attributes/stats.go
```diff
@@ -38,6 +38,7 @@ const (
 	BaseHP
 	BaseATK
 	BaseDEF
+	Elevation
```
Charlie-Zheng: Move this to AttackInfo struct, like how the BaseDmgBonus is set up

pkg/core/info/combat.go
```diff
@@ -42,6 +42,7 @@ type AttackInfo struct {
 	StrikeType       attacks.StrikeType
 	UseDef           bool    // we use this instead of flatdmg to make sure stat snapshotting works properly
 	UseHP            bool    // we use this instead of flatdmg to make sure stat snapshotting works properly
+	UseEM            bool    // we use this instead of flatdmg to make sure stat snapshotting works properly
```
Charlie-Zheng: I think you should update the comment to signify that Lunar damages need to differentiate between EM MVs and flat damage

pkg/enemy/damage.go
```diff
@@ -105,9 +107,11 @@ func (e *Enemy) calc(atk *info.AttackEvent, evt glog.Event) (float64, bool) {
 	}
 
 	if attacks.DirectLunarReactionStartDelim < atk.Info.AttackTag && atk.Info.AttackTag < attacks.DirectLunarReactionEndDelim {
+		elevation := atk.Snapshot.Stats[attributes.Elevation]
```
Charlie-Zheng: Put elevation on the outside. There isn't any reason why it can't be applied to normal talent damage in the future

pkg/enemy/damage.go
Charlie-Zheng, line 1: You need to adjust the location where flat damage is added for reaction and direct lunar damage

pkg/reactable/bloom.go
```diff
@@ -42,9 +44,17 @@ func (r *Reactable) TryBloom(a *info.AttackEvent) bool {
...
 	a.Reacted = true
-
+	if _, ok := r.core.Flags.Custom[LunarBloomEnableKey]; ok {
+		r.lunarBloom()
+	}
 	r.addBloomGadget(a)
-	r.core.Events.Emit(event.OnBloom, r.self, a)
+
+	bloomEvent := event.OnBloom
+	if _, ok := r.core.Flags.Custom[LunarBloomEnableKey]; ok {
```
Charlie-Zheng: I think you can just do 
```
if r.core.Flags.Custom[LunarBloomEnableKey] != 0 {
    bloomEvent = event.OnLunarBloom
}
```
Values that aren't in the map are considered 0.

Either way you should check if the flag is set to 0 or not.

pkg/core/attacks/icd.go
Matoba004: by this you mean the whole thing? while the direct lb attacks should be handled by having no durability in ai?

pkg/core/attributes/stats.go
Matoba004: i assume that in attack struct it's purpose wont be to act as key for "elevation" stat but value of it, right?

pkg/core/attacks/icd.go
Charlie-Zheng: Yeah, the attacks just don't have durability (that's the default)

pkg/core/attributes/stats.go
Charlie-Zheng: Yeah, and the elevation buffs can be provided using an OnEnemyHit subscription

PR #2537, opened 2025-12-14
Implement Lauma

Matoba004:
Blocked by #2536

- [x]  New character package
- [x]  Config in character package
- [x]  Run pipeline with added config (generates character curve, talent stats, .generated.json files)
- [x]  Character key
- [x]  Shortcuts for character key
- [x]  Update mode_gcsim.js with shortcuts for syntax highlighting
- [x]  Add Character package to imports
- [x]  Normal Attack
- [x]  Charge Attack / Aimed Shot
- [x]  Skill
- [x]  Burst
- [x]  A1
- [x]  A4
- [x]  C1
- [x]  C2
- [x]  C3
- [x]  C4
- [x]  C5
- [x]  C6
- [ ]  Other necessary talents (custom dash/jump, low/high plunge, ...)
- [x]  Hitlag
- [x]  ICD
- [x]  StrikeType
- [x]  PoiseDMG (blunt attacks only for now)
- [x]  Hitboxes
- [x]  Attack durability
- [x]  Particles
- [x]  Frames
- [x]  Update documentation
- [x]  Xingqiu/Yelan N0 (optional)
- [ ]  Xianyun Plunge (optional)


internal/characters/kuki/charge.go
```diff
@@ -53,7 +53,7 @@ func (c *char) ChargeAttack(p map[string]int) (action.Info, error) {
 	return action.Info{
 		Frames:          frames.NewAbilFunc(chargeFrames),
 		AnimationLength: chargeFrames[action.InvalidAction],
-		CanQueueAfter:   chargeHitmarks[len(chargeHitmarks)-1],
```
Charlie-Zheng, line 56: Why is this change required?

internal/characters/lauma/attack.go
```diff
@@ -0,0 +1,83 @@
...
+		Mult:       attack[c.NormalCounter][c.TalentLvlAttack()],
+	}
+
+	if c.Base.Cons >= 6 && len(c.paleHymnStacks) != 0 {
+		ai.Abil = "Normal C6 Pale Hymn"
+		ai.AttackTag = attacks.AttackTagDirectLunarBloom
+		ai.Durability = 0
+		ai.Mult = 1.5
+		ai.IgnoreDefPercent = 1
+
```
Charlie-Zheng, line 59: Add `ai.UseEM = True`
Add `ai.ICDTag = attacks.ICDTagDefault`

internal/characters/lauma/attack.go
```diff
@@ -0,0 +1,83 @@
...
+		c.paleHymnStacks = c.paleHymnStacks[:1]
```
Charlie-Zheng: This truncates to the first stack left. You should use a queue. Golang doesn't have a queue in the standard library so probably use some existing queue implementation from online. Alternatively, since this queue is single ended, you can use a channel as a queue.

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+			return burstFrames[next]
+		},
+		AnimationLength: burstFrames[action.InvalidAction],
+		CanQueueAfter:   burstFrames[action.ActionSwap], // earliest cancel
+		State:           action.BurstState,
+	}, nil
+}
+
+func (c *char) setupPaleHymnBuff() {
```
Charlie-Zheng: Generally I just call this `initBurst()`.

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+			return false
+		}
+
+		c.paleHymnStacks = c.paleHymnStacks[:1]
+
+		return false
+	}, "lauma-pale-hymn-buff")
+}
+
+func (c *char) addC6PaleHymn(a info.AttackCB) {
```
Charlie-Zheng: Move this to cons.go, rename to `addC6PaleHymnCB`

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		default:
+			return false
+		}
+
+		if ae.Info.Abil == "Frostgrove Sanctuary C6" {
+			return false
+		}
+
+		c.paleHymnStacks = c.paleHymnStacks[:1]
```
Charlie-Zheng: This just truncates to only 1 pale hymn stack. You should reduce the number of stacks by 1 instead. Make a function that does the stack removal logic and call that here

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+func init() {
+	core.RegisterCharFunc(keys.Lauma, NewChar)
+}
+
+type char struct {
+	*tmpl.Character
+	ascendantGleam       bool
+	deerStateReady       bool
+	c6SkillPaleHymnCount int
+	paleHymnStacks       []paleHymnStack
```
Charlie-Zheng: I recommend having two separate queues for the normal stacks and the c6 stacks. When the expiry for the c6 stacks is reached then clear all the c6 stacks. Otherwise take a stack from whichever queue has the older stack.
Update the palehymnStack struct to
```
type paleHymnStack struct {
    src int
	expiry int
}
```
Golang doesn't have a stdlib queue so you will need to grab one from online or add a library for it.

Then update paleHymnStacks to a queue
```
paleHymnStacks   Queue[paleHymnStack]
c6PaleHymnStacks Queue[int]
c6PaleHymnExpiry int
```
The c6 ones share expiry so they don't need to use the struct

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+		switch ae.Info.AttackTag {
+		case attacks.AttackTagBurningDamage | attacks.AttackTagBloom | attacks.AttackTagHyperbloom | attacks.AttackTagBurgeon:
+			ae.Info.FlatDmg += em * bloomDmgIncrease[c.TalentLvlBurst()]
+
+			if c.Base.Cons >= 2 {
+				ae.Info.FlatDmg += em * 5
+			}
+		case attacks.AttackTagDirectLunarBloom:
+
+			// check if lauma c6 skill gets buffed despite not consuming hymn
```
Charlie-Zheng: Also check if lauma c6 NA gets buffed consuming two stacks at least per NA?

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		if len(c.paleHymnStacks) == 0 {
+			return false
+		}
+
+		ae := args[1].(*info.AttackEvent)
+		em := c.Stat(attributes.EM)
+
+		switch ae.Info.AttackTag {
+		case attacks.AttackTagBurningDamage | attacks.AttackTagBloom | attacks.AttackTagHyperbloom | attacks.AttackTagBurgeon:
```
Charlie-Zheng: Remove burning and add AttackTagBountifulCore

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+			ae.Info.FlatDmg += em * bloomDmgIncrease[c.TalentLvlBurst()]
+
+			if c.Base.Cons >= 2 {
```
Charlie-Zheng: Move the c2 bonus damage to cons.go

```
func (c *char) c2PaleHymnScaling(isLunar bool) float64 {
    if c.Base.Cons < 2 {
         return 0
    }
    if isLunar {
        return 4
    }
    return 5
}
```

Then here you can just call `ae.Info.FlatDmg += em * c.c2PaleHymnScaling(false)`

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+				ae.Info.FlatDmg += em * 5
+			}
+		case attacks.AttackTagDirectLunarBloom:
+
+			// check if lauma c6 skill gets buffed despite not consuming hymn
+
+			ae.Info.FlatDmg += em * lunarBloomDmgIncrease[c.TalentLvlBurst()]
+
+			if c.Base.Cons >= 2 {
```
Charlie-Zheng: Same as above but with `ae.Info.FlatDmg += em * c.c2PaleHymnScaling(true)`

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+				ae.Info.FlatDmg += em * 4
+			}
+		default:
+			return false
+		}
+
+		if ae.Info.Abil == "Frostgrove Sanctuary C6" {
```
Charlie-Zheng: Make the c6 abil name a variable so that it's not fragile to naming changes

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (c *char) addC6PaleHymn(a info.AttackCB) {
+	if c.Base.Cons < 6 {
+		return
+	}
+
+	c.addPaleHymn(2, true)
+}
+
+func (c *char) addPaleHymn(amount int, isC6 bool) {
```
Charlie-Zheng: You can just split the addPaleHymn function into two functions, instead of having a flag

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+
+func init() {
+	// deer state frames
+	deerFrames = frames.InitAbilSlice(64) // skill
+	deerFrames[action.ActionAttack] = 61
+	deerFrames[action.ActionCharge] = 63
+	// deerFrames[action.ActionSkillHoldFramesOnly] = 63
+	deerFrames[action.ActionBurst] = 63
+	deerFrames[action.ActionSkill] = 63
+	deerFrames[action.ActionSwap] = 0
```
Charlie-Zheng: Cancel frame shouldn't be before the transformation frame. Use `chargeDeerFrame`

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// charge attack frames
+	chargeFrames = frames.InitAbilSlice(68) // CA -> swap
+	chargeFrames[action.ActionAttack] = 67
```
Charlie-Zheng: Set Skill/Burst/Dash/Jump to use the CA release frame

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/info"
+)
+
+var (
+	deerFrames   []int
+	chargeFrames []int
+)
+
+const (
+	chargeHitmark   = 73
```
Charlie-Zheng: Add CA release frame at 67

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+		CanQueueAfter:   chargeFrames[action.ActionAttack],
+		State:           action.ChargeAttackState,
+	}, nil
+}
+
+func (c *char) enterDeerState() (action.Info, error) {
+	c.deerStateStaminaBleed()
+	c.AddStatus(deerStatusKey, 10*60, true)
+
+	c.endDeerStateCondition()
```
Charlie-Zheng: move this to the Init() function.

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return action.Info{
+		Frames:          frames.NewAbilFunc(deerFrames),
+		AnimationLength: deerFrames[action.InvalidAction],
+		CanQueueAfter:   deerFrames[action.ActionSwap],
+		State:           action.ChargeAttackState,
+	}, nil
+}
+
+func (c *char) endDeerStateCondition() {
```
Charlie-Zheng: rename this to `chargeInit()`. Add a subscription to cancel deer state when swapping. You can reference/copy another character's `onExitField` function for it

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (c *char) deerStateStaminaBleed() {
+	if !c.StatusIsActive(deerStatusKey) {
+		return
+	}
+	staminaCost := 25.0 / 60.0
+	if c.Core.Player.Stam < staminaCost {
+		c.endDeerState()
+	}
+	c.Core.Player.RestoreStam(-staminaCost)
```
Charlie-Zheng: Use `c.Core.Player.UseStam()` instead of restoreStam with negative amount

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+func (c *char) AnimationStartDelay(k info.AnimationDelayKey) int {
+	switch k {
+	case info.AnimationXingqiuN0StartDelay:
+		return 10
+	case info.AnimationYelanN0StartDelay:
+		return 10
+	default:
+		return c.Character.AnimationStartDelay(k)
+	}
+}
```
Charlie-Zheng, line 88: Add `ActionStam` function and adjust stamina costs for CA based on whether in Deer state or not. Should be 50 stamina in deer state, and 0 stamina outside of deer state.

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+		c.c6Elevation()
+	}
+
+	c.setupPaleHymnBuff()
+
+	return nil
+}
+
+func (c *char) ActionReady(a action.Action, p map[string]int) (bool, action.Failure) {
+	return c.Character.ActionReady(a, p)
```
Charlie-Zheng, line 76: Add CD for Lauma deer state transform here. Return a Cooldown failure if the deer state isn't ready

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+	c.Core.Tasks.Add(c.deerStateStaminaBleed, 1)
+}
+
+func (c *char) endDeerState() {
+	c.DeleteStatus(deerStatusKey)
+	c.Core.Tasks.Add(func() {
+		c.deerStateReady = true
+		c.Core.Log.NewEvent("deer state refreshed", glog.LogCharacterEvent, c.Index())
```
Charlie-Zheng: Use a cooldown event `c.Core.Log.NewEventBuildMsg(glog.LogCooldownEvent, c.Index(), "charge cooldown ready")`

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+	deerFrames[action.ActionSwap] = 0
+
+	// charge attack frames
+	chargeFrames = frames.InitAbilSlice(68) // CA -> swap
+	chargeFrames[action.ActionAttack] = 67
+}
+
+func (c *char) ChargeAttack(p map[string]int) (action.Info, error) {
+	// by default enters different state where she doesnt hit enemies and consumes 25 stamina per second
+	if c.deerStateReady {
```
Charlie-Zheng: check if actually in the deer state. If not in deer state, enter deer state.

internal/characters/lauma/dash.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"errors"
+
+	"github.com/genshinsim/gcsim/pkg/core/action"
+)
+
+func (c *char) Dash(p map[string]int) (action.Info, error) {
+	if c.StatusIsActive(deerStatusKey) {
+		return action.Info{}, errors.New("dash called in deer state")
```
Charlie-Zheng: `return action.Info{}, errors.New("cannot use dash in lauma spirit envoy state")`

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+
+var (
+	deerFrames   []int
+	chargeFrames []int
+)
+
+const (
+	chargeHitmark   = 73
+	chargeDeerFrame = 67
+	deerStatusKey   = "lauma-deer-state"
```
Charlie-Zheng: use `lauma-spirit-envoy`

internal/characters/lauma/jump.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"errors"
+
+	"github.com/genshinsim/gcsim/pkg/core/action"
+)
+
+func (c *char) Jump(p map[string]int) (action.Info, error) {
+	if c.StatusIsActive(deerStatusKey) {
+		return action.Info{}, errors.New("jump in deer state isn't implemented")
```
Charlie-Zheng: `return action.Info{}, errors.New("jump in lauma spirit envoy state isn't implemented")`

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+func NewChar(s *core.Core, w *character.CharWrapper, _ info.CharacterProfile) error {
+	c := char{}
+	c.Character = tmpl.NewWithWrapper(s, w)
+
+	c.EnergyMax = 60
+	c.NormalHitNum = normalHitNum
+	c.SkillCon = 5
+	c.BurstCon = 3
+	c.deerStateReady = true
+	c.c6SkillPaleHymnCount = 8
```
Charlie-Zheng: make this start from 0 and count up instead of starting from 8 and counting down

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/genshinsim/gcsim/pkg/modifier"
+)
+
+const (
+	c1Key     = "lauma-threads-of-life"
+	c1HitMark = 5
+)
+
+func (c *char) c1() {
+	// on lb proc heal
```
Charlie-Zheng, line 29: Add
```
if c.Base.Cons < 1 {
    return
}
```

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+		c.Core.Tasks.Add(func() {
+			c.c1()
+		}, 19*6)
+
+		return true
+	}, "lauma-c1")
+}
+
+func (c *char) c2() {
+	// if !c.ascendantGleam {
```
Charlie-Zheng: Add
```
if c.Base.Cons < 2 {
    return
}
```

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+	// 	return
+	// }
+
+	bonus := 0.4
+
+	for _, x := range c.Core.Player.Chars() {
+		this := x
```
Charlie-Zheng: delete this

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+		// special case if applying 4 Noblesse to holder to fix this mess:
```
Charlie-Zheng: remove extraneous comments

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+	if count >= 2 {
+		c.ascendantGleam = true
+	} else {
+		c.ascendantGleam = false
+	}
+
+	c.a1()
+	c.a4()
+
+	if c.Base.Cons >= 1 {
```
Charlie-Zheng: remove the conditional here, and do the checking in c.c1() function

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+		c.c1()
+	}
+
+	if c.Base.Cons >= 2 {
```
Charlie-Zheng: remove the conditional here, and do the checking in c.c2() function

internal/characters/lauma/lauma.go
```diff
@@ -0,0 +1,88 @@
...
+		c.c2()
+	}
+
+	if c.Base.Cons >= 6 {
```
Charlie-Zheng: remove the conditional here, and do the checking in c.c6Elevation() function

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+				if atk.AttackTag != attacks.AttackTagDirectLunarBloom {
+					return 0, false
+				}
+				return bonus, false
+			},
+		})
+	}
+}
+
+func (c *char) c4Refund(a info.AttackCB) {
```
Charlie-Zheng: Rename to `c4RefundCB`

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if c.StatusIsActive(laumaC4RefundKey) {
+		return
+	}
+	c.AddEnergy("lauma-c4-refund", 5)
+	c.AddStatus(laumaC4RefundKey, 5*60, true)
+}
+
+func (c *char) c6Elevation() {
+	if !c.ascendantGleam {
```
Charlie-Zheng, line 144: Add
```
if c.Base.Cons < 6 {
    return
}
```

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+
+		bonus := 0.25
+
+		if c.Core.Flags.LogDebug {
+			c.Core.Log.NewEvent("lauma-c6-elevation", glog.LogCharacterEvent, c.Index()).Write("bonus", bonus)
+		}
+
+		atk.Info.Elevation += bonus
+		return false
+	}, lunarbloomBonusKey)
```
Charlie-Zheng: use `lunarbloomBonusKey+"-c6"`

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (c *char) c6Elevation() {
+	if !c.ascendantGleam {
+		return
+	}
+
+	c.Core.Events.Subscribe(event.OnEnemyHit, func(args ...any) bool {
+		atk := args[1].(*info.AttackEvent)
+
+		switch atk.Info.AttackTag {
```
Charlie-Zheng: Don't use a switch for only one case. Change to if statement

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+		case attacks.AttackTagDirectLunarBloom:
+		default:
+			return false
+		}
+
+		bonus := 0.25
```
Charlie-Zheng: bonus is constant, just use 0.25 directly, or make it a const outside of the function ` const c6bonus = 0.25`

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+		// heal active character
+		c.Core.Tasks.Add(func() {
+			c.Core.Player.Heal(info.HealInfo{
+				Type:    info.HealTypeAbsolute,
+				Message: "Lauma C1 (Heal)",
+				Src:     healAmt,
+			})
+		}, c1HitMark)
+
+		c.Core.Tasks.Add(func() {
```
Charlie-Zheng: use c.QueueCharTask()

internal/characters/lauma/cons.go
```diff
@@ -0,0 +1,108 @@
...
+			c.c1()
+		}, 19*6)
```
Charlie-Zheng: Just use 1.9*60

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+	skillFrames[1][action.ActionSwap] = 56
+}
+
+func (c *char) Skill(p map[string]int) (action.Info, error) {
+	c.AddStatus(c1Key, 20*60, true)
+	c.AddStatus(frostgroveSanctuaryKey, 15*60, true)
+	c.AddStatus(a1Key, 20*60, true)
+
+	if c.Base.Cons >= 6 {
+		c.c6SkillPaleHymnCount = 8
```
Charlie-Zheng: make it down up from 0 instead of down from 8.
I recommend moving this to a function in cons.go called `c6OnSkill()`

pkg/reactable/lunarbloom.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/glog"
+)
+
+const (
+	maxVerdantDew        = 3
+	verdantDewStartFrame = 13
+	verdantDewEndFrame   = 140
+	nextDewFrameRefresh  = 150
+	LbKey                = "lunar-bloom"
+	verdantDewKey        = "verdant-dew"
```
Charlie-Zheng: Should be public `VerdantDewKey`

pkg/reactable/lunarbloom.go
```diff
@@ -0,0 +1,51 @@
...
+	nextDewFrameKey      = "verdant-dew-next"
+)
+
```
Charlie-Zheng: Add a function to query the amount of verdant dew

pkg/reactable/bloom.go
```diff
@@ -42,9 +44,17 @@ func (r *Reactable) TryBloom(a *info.AttackEvent) bool {
 	a.Info.Durability -= consumed
 	a.Info.Durability = max(a.Info.Durability, 0)
 	a.Reacted = true
-
+	if _, ok := r.core.Flags.Custom[LunarBloomEnableKey]; ok {
```
Charlie-Zheng: save this into a var so you don't need to access the map twice

pkg/reactable/bloom.go
```diff
@@ -63,8 +73,17 @@ func (r *Reactable) tryQuickenBloom(a *info.AttackEvent) {
 	consumed := r.reduce(attributes.Hydro, avail, 2)
 	r.reduceMod(info.ReactionModKeyQuicken, consumed)
 
+	if _, ok := r.core.Flags.Custom[LunarBloomEnableKey]; ok {
```
Charlie-Zheng: save this into a var so you don't need to access the map twice

pkg/enemy/damage.go
```diff
@@ -44,6 +44,8 @@ func (e *Enemy) calc(atk *info.AttackEvent, evt glog.Event) (float64, bool) {
 		a = atk.Snapshot.Stats.MaxHP()
 	case atk.Info.UseDef:
 		a = atk.Snapshot.Stats.TotalDEF()
+	case atk.Info.UseEM:
+		a = atk.Snapshot.Stats[attributes.EM]
 	default:
 		a = atk.Snapshot.Stats.TotalATK()
 	}
```
Charlie-Zheng, line 56: Missing the adjustments for flat dmg for reactions/direct lunar damage adding in a different location than normally

pkg/enemy/damage.go
```diff
@@ -123,6 +125,9 @@ func (e *Enemy) calc(atk *info.AttackEvent, evt glog.Event) (float64, bool) {
 		x *= 3.0
 	}
 
+	elevation := atk.Info.Elevation
+	damage *= 1 + elevation
+
```
Charlie-Zheng, line 119: Add base dmg mult and elevation into the log as well. probably adjust the log to only log the relevant stat for the multiplier. (Only log def info if UseDef is true, only log hp info if UseHP is true, and only log EM info if UseEM is true)

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+		case attacks.AttackTagDirectLunarBloom:
+		default:
+			return false
+		}
+
+		ae.Snapshot.Stats[attributes.CR] += 0.1
+		ae.Snapshot.Stats[attributes.CD] += 0.2
+
+		c.Core.Log.NewEvent("lauma a1 buff", glog.LogCharacterEvent, ae.Info.ActorIndex).
+			Write("final_crit", ae.Snapshot.Stats[attributes.CR])
```
Charlie-Zheng: also log the CD

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+		switch ae.Info.AttackTag {
+		case attacks.AttackTagDirectLunarBloom:
+		default:
+			return false
+		}
+
+		ae.Snapshot.Stats[attributes.CR] += 0.1
+		ae.Snapshot.Stats[attributes.CD] += 0.2
+
+		c.Core.Log.NewEvent("lauma a1 buff", glog.LogCharacterEvent, ae.Info.ActorIndex).
```
Charlie-Zheng: Check if Debug flag is enabled

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+			Write("final_crit", ae.Snapshot.Stats[attributes.CR])
+
+		return false
+	}, "lauma-a1-reaction-dmg-buff")
+}
+
+func (c *char) a4() {
```
Charlie-Zheng: rename to a4Init

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/genshinsim/gcsim/pkg/reactable"
+)
+
+const (
+	lunarbloomBonusKey = "lauma-lunarbloom-bonus"
+	a1Key              = "light-for-the-frosty-night"
+	nextDewFrameKey    = "verdant-dew-next"
+)
+
+func (c *char) a1() {
```
Charlie-Zheng: Rename to a1Init()

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+			return m, true
+		},
+	})
+}
+
+func (c *char) lunarbloomInit() {
+	c.Core.Flags.Custom[reactable.LunarBloomEnableKey] = 1
+
+	c.Core.Flags.Custom[nextDewFrameKey] = 150
```
Charlie-Zheng: Why is this needed?

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// TODO: moonsign?
```
Charlie-Zheng: Remove this

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// TODO: every 1 EM that Lauma has increasing Lunar-Bloom's Base DMG by 0.0175%, up to a maximum of 14%.
```
Charlie-Zheng: Remove the todo

internal/characters/lauma/charge.go
```diff
@@ -0,0 +1,132 @@
...
+	c.Core.Player.RestoreStam(-staminaCost)
+	c.Core.Tasks.Add(c.deerStateStaminaBleed, 1)
+}
+
+func (c *char) endDeerState() {
+	c.DeleteStatus(deerStatusKey)
+	c.Core.Tasks.Add(func() {
+		c.deerStateReady = true
+		c.Core.Log.NewEvent("deer state refreshed", glog.LogCharacterEvent, c.Index())
+	}, 4*60)
```
Charlie-Zheng: Need to add the a4 CDR effect. Change to 4*60*c.a4CaCDR()

internal/characters/lauma/asc.go
```diff
@@ -0,0 +1,139 @@
...
+		bonus := min(em*0.000175, 0.14)
+
+		if c.Core.Flags.LogDebug {
+			c.Core.Log.NewEvent("lauma adding lunarbloom base damage", glog.LogCharacterEvent, c.Index()).Write("bonus", bonus)
+		}
+
+		atk.Info.BaseDmgBonus += bonus
+		return false
+	}, lunarbloomBonusKey)
+}
```
Charlie-Zheng, line 136: Add `c.a4CaCDR()` for adjusting her charge CD. Return 1.0 if not A4, otherwise `return 1.0 - cdr`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		AnimationLength: skillFrames[0][action.InvalidAction],
+		CanQueueAfter:   skillFrames[0][action.ActionSwap], // earliest cancel
+		State:           action.SkillState,
+	}, nil
+}
+
+func (c *char) skillHold() (action.Info, error) {
+	ai := info.AttackInfo{
+		ActorIndex: c.Index(),
+		Abil:       "Runo: Dawnless Rest of Karsikko (Hold Hit 1)",
```
Charlie-Zheng: use `Runo: Dawnless Rest of Karsikko`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		ICDGroup:   attacks.ICDGroupDefault,
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Dendro,
+		Durability: 25,
+		Mult:       skillHold1[c.TalentLvlSkill()],
+	}
+
+	aiDirectLB := info.AttackInfo{
+		ActorIndex:       c.Index(),
+		Abil:             "Runo: Dawnless Rest of Karsikko (Hold Hit 2)",
```
Charlie-Zheng: Use `Runo: Dawnless Rest of Karsikko (Lunar-Bloom)`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+	frostgroveSanctuaryFirstHitPress  = 31
+	frostgroveSanctuaryFirstHitHold   = 24
+	frostgroveSanctuaryInterval       = 117
+	skillHoldHitmark                  = 45
+	skillConsumeDew                   = 29
+	skillOffset                       = 0
+	frostgroveSanctuaryKey            = "lauma-frostgrove-sanctuary"
+	frostgroveSanctuaryParticleICDKey = "lauma-frostgrove-sanctuary-particle-icd"
+	laumaC4RefundKey                  = "lauma-c4-refund"
+	verdantDewKey                     = "verdant-dew"
```
Charlie-Zheng: Delete. Use the one in reaction.go after you make it public

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (c *char) moonSong(dewCount int) {
+	if !c.StatusIsActive(burstKey) {
+		return
+	}
+
+	c.addPaleHymn(dewCount*6, false)
+}
+
+func (c *char) applySkillShred(a info.AttackCB) {
```
Charlie-Zheng: rename to `applySkillShredCB`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+	e, ok := a.Target.(*enemy.Enemy)
+	if !ok {
+		return
+	}
+	shredAmount := skillResShred[c.TalentLvlSkill()]
+	e.AddResistMod(info.ResistMod{
+		Base:  modifier.NewBaseWithHitlag("lauma-skill-shred", 10*60),
```
Charlie-Zheng: `"lauma-skill-shred-dendro"` Otherwise mods with the same name will override each other

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		Ele:   attributes.Dendro,
+		Value: -shredAmount,
+	})
+	e.AddResistMod(info.ResistMod{
+		Base:  modifier.NewBaseWithHitlag("lauma-skill-shred", 10*60),
```
Charlie-Zheng: `"lauma-skill-shred-hydro"` Otherwise mods with the same name will override each other

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		Ele:   attributes.Hydro,
+		Value: -shredAmount,
+	})
+}
+
+func (c *char) frostgroveSantuary(a info.AttackCB) {
```
Charlie-Zheng: Does the sanctuary only spawn if you hit an enemy with the skill? I don't think this should be a callback. This function should be `frostgroveSantuaryTick`. Just call the first tick with the correct delay

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		ICDTag:     attacks.ICDTagNone,
+		ICDGroup:   attacks.ICDGroupDefault,
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Dendro,
+		Durability: 25,
+		Mult:       frostgroveSanctuaryAtk[c.TalentLvlSkill()],
+	}
+
+	frostgroveSanctuaryFirstHit := frostgroveSanctuaryFirstHitPress
+	if a.AttackEvent.Info.Abil == "Runo: Dawnless Rest of Karsikko (Hold Hit 1)" {
```
Charlie-Zheng: Delete this since you are calling the first tick with the correct delay

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		frostgroveSanctuaryFirstHit = frostgroveSanctuaryFirstHitHold
+	}
+
+	c.Core.Tasks.Add(func() {
```
Charlie-Zheng: Add a src check so that using skill twice doesn't make two frostgroves

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		c.applySkillShred,
+		c.c4Refund,
+	)
+
+	if c.Base.Cons >= 6 {
+		if c.c6SkillPaleHymnCount <= 0 {
+			return
+		}
+
+		c.c6SkillPaleHymnCount--
```
Charlie-Zheng: Count up from 0 instead of down from 8

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+			6,
+		),
+		frostgroveSanctuaryFirstHit,
+		frostgroveSanctuaryFirstHit,
+		c.particleCB,
+		c.applySkillShred,
+		c.c4Refund,
+	)
+
+	if c.Base.Cons >= 6 {
```
Charlie-Zheng: Move this to cons.go in a new function `c6OnFrostgroveTick`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+
+		aiC6 := info.AttackInfo{
+			ActorIndex:       c.Index(),
+			Abil:             "Frostgrove Sanctuary C6",
+			AttackTag:        attacks.AttackTagDirectLunarBloom,
+			ICDTag:           attacks.ICDTagNone,
+			ICDGroup:         attacks.ICDGroupDefault,
+			StrikeType:       attacks.StrikeTypeDefault,
+			Element:          attributes.Dendro,
+			Durability:       0,
```
Charlie-Zheng: Delete

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+var skillFrames [][]int
+
+const (
+	skillPressHitmark                 = 16
+	frostgroveSanctuaryFirstHitPress  = 31
+	frostgroveSanctuaryFirstHitHold   = 24
+	frostgroveSanctuaryInterval       = 117
+	skillHoldHitmark                  = 45
+	skillConsumeDew                   = 29
+	skillOffset                       = 0
```
Charlie-Zheng: Is this actually 0? If it's actually 0 then remove all the offsets and use just `nil` for the offset.

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+			info.Point{Y: skillOffset},
+			6,
+		),
+		skillHoldHitmark,
+		skillHoldHitmark,
+		c.frostgroveSantuary,
+		c.applySkillShred,
+	)
+
+	c.Core.Tasks.Add(func() {
```
Charlie-Zheng: Use c.QueueCharTask

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+		dewCount := c.Core.Flags.Custom[verdantDewKey]
+
+		aiDirectLB.Mult = skillHold2[c.TalentLvlSkill()] * dewCount
+
+		c.Core.Flags.Custom[verdantDewKey] -= dewCount
```
Charlie-Zheng: Just do `c.Core.Flags.Custom[verdantDewKey] = 0.0`

internal/characters/lauma/skill.go
```diff
@@ -0,0 +1,296 @@
...
+	return action.Info{
+		Frames:          frames.NewAbilFunc(skillFrames[1]),
+		AnimationLength: skillFrames[1][action.InvalidAction],
+		CanQueueAfter:   skillFrames[1][action.ActionSwap], // earliest cancel
+		State:           action.SkillState,
+	}, nil
+}
+
+func (c *char) moonSong(dewCount int) {
+	if !c.StatusIsActive(burstKey) {
```
Charlie-Zheng, line 212: Add Moon Song stacks when burst isn't active

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+
+const (
+	burstKey          = "lauma-burst"
+	burstDeerFrame    = 115
+	paleHymnGainFrame = 96
+)
+
+func (c *char) Burst(p map[string]int) (action.Info, error) {
+	c.AddStatus(c1Key, 20*60, true)
+
```
Charlie-Zheng: Add consuming moon song stacks

internal/characters/lauma/burst.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+const (
+	burstKey          = "lauma-burst"
+	burstDeerFrame    = 115
+	paleHymnGainFrame = 96
+)
+
+func (c *char) Burst(p map[string]int) (action.Info, error) {
+	c.AddStatus(c1Key, 20*60, true)
```
Charlie-Zheng: Move this to a function in cons.go called `c1OnBurst()`. Only add the status if c1 or higher.

pkg/core/attacks/icd.go
```diff
@@ -88,6 +88,7 @@ func init() {
 	ICDGroupDamageSequence[ICDGroupPoleExtraAttack] = []float64{1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
 	ICDGroupDamageSequence[ICDGroupReactionA] = []float64{1, 1, 0, 0, 0, 0, 0, 0, 0, 0}
 	ICDGroupDamageSequence[ICDGroupReactionB] = []float64{1, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+
```
Charlie-Zheng: remove the extra new line

Charlie-Zheng (changes requested):

internal/characters/lauma/asc.go
Matoba004: Because in lunarbloom i am counting down instead of counting up, it can be removed with slight change to lunarbloom.

internal/characters/lauma/attack.go
Matoba004: ai.ICDTag = attacks.ICDTagNone?

Charlie-Zheng: Yeah

internal/characters/lauma/asc.go
Matoba004: like this?

Write("final_critrate", ae.Snapshot.Stats[attributes.CR]).
Write("final_critdmg", ae.Snapshot.Stats[attributes.CD])

internal/characters/lauma/skill.go
Matoba004: not "Runo: Dawnless Rest of Karsikko (Hold)" at least?

internal/characters/lauma/lauma.go
Matoba004: why would there be cooldown failure for action that should just result in different action?

pkg/reactable/lunarbloom.go
Matoba004: func (r *Reactable) GetVerdantDew() int {
	return int(r.core.Flags.Custom[VerdantDewKey])
}

you mean something like this?

should i put it in reactable interface and use it in characters to get the amount of verdant dew?

internal/characters/lauma/burst.go
```diff
@@ -88,92 +83,69 @@ func (c *char) initBurst() {
...
-	if c.c6PaleHymnStacks.Len() == 0 {
+	if c.c6PaleHymnCount <= 0 {
 		c.paleHymnStacks.PopFront()
 		return
 	}
+
+	// pop whichever one is closer to expiry
 	currentPaleHymn := c.paleHymnStacks.Front()
-	currentC6PaleHymn := c.c6PaleHymnStacks.Front()
+	currentC6PaleHymn := c.c6PaleHymnExpiry
```
Matoba004: oldest first not closest to expiry first

Matoba004: You changed the priority from oldest to closest to expiry, my tests of the c6 stuff say that this isn't correct, oldest first is correct behavior.

internal/characters/lauma/cons.go
```diff
@@ -140,11 +146,11 @@ func (c *char) c6OnFrostgroveTick() {
 	if c.Base.Cons < 6 {
 		return
 	}
-	if c.c6SkillPaleHymnCount > 7 {
+	if c.c6PaleHymnCount > 7 {
```
Matoba004: What if one stack gets consumed? Wouldn't that allow potentially nine stacks to be generated? (skill doesnt actually hit 9 times but if it did it would be an issue)

internal/characters/lauma/charge.go
```diff
@@ -140,14 +144,25 @@ func (c *char) deerStateStaminaBleed() {
 	if c.Core.Player.Stam < staminaCost {
 		c.endDeerState()
 	}
-	c.Core.Player.UseStam(staminaCost, action.ActionWait)
+	c.Core.Player.UseStam(staminaCost, action.ActionCharge)
```
Matoba004: Bloodstained Chivalry shouldn't affect this ability. If action.ActionCharge means it does other abitity type should be used.

Charlie-Zheng: Did you check what exactly affects the ability then? Also, you did test that right?

Matoba004: I didn't check what exactly affects it but I did this one test.

Charlie-Zheng: Can you test if it is affected by running stamina mods or special sprint stamina mods? If neither of these affect it, can you also test if standing without moving consumes stamina?

Matoba004: It's not affected by either sprint stamina mods or alternate sprint stamina mods. You cannot stand still in spirit envoy state.

Passives used in tests:
sprint stamina - kazuha passive
alternate sprint stamina - wine and song passive

Charlie-Zheng (dismissed):

PR #2579, opened 2026-02-26
Hexerei implementation

Matoba004:


Charlie-Zheng:
Remove hexerei from characters that don't have it implemented.

Charlie-Zheng:
@Matoba004 because it's incorrect to say the character is Hexerei (and thus enable the Secret Rite) without their Hexerei talents implemented.

PR #2590, opened 2026-03-25
Update Diona skill.go

jwalk-511:
Assuming https://discord.com/channels/845087716541595668/1486415594424242338/1486419483630243901 is true, needs to be tested

### DIFF
--- a/internal/characters/diona/skill.go
+++ b/internal/characters/diona/skill.go
@@ -121,13 +121,8 @@ func (c *char) pawsPewPew(f, travel, pawCount int) {
 					shd.HP = shdHp
 					shd.Src = skillSrc
 					shd.Expires += dur
-					maxDur := c.Core.F + dur*5
-					if shd.Expires > maxDur {
-						shd.Expires = maxDur
-					}
-				} else {
-					shd.Expires += dur
 				}
+				shd.Expires += dur
 			} else {
 				shd = &shield.Tmpl{
 					ActorIndex: c.Index(),
@@ -139,6 +134,11 @@ func (c *char) pawsPewPew(f, travel, pawCount int) {
 					Ele:        attributes.Cryo,
 					Expires:    c.Core.F + dur, // 15 sec
 				}
+			}
+
+			maxDur := c.Core.F + dur*5
+			if shd.Expires > maxDur {
+				shd.Expires = maxDur
 			}
 			// TODO: check that this is actually properly extending duration
 			c.Core.Player.Shields.Add(shd)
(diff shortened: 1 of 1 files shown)


jwalk-511:
Now following logic of https://discord.com/channels/845087716541595668/1486415594424242338/1486428157992112320

jwalk-511:
Test config here: https://gcsim.app/sh/MGHnGG76NqJR

internal/characters/diona/skill.go
```diff
@@ -109,25 +110,36 @@ func (c *char) pawsPewPew(f, travel, pawCount int) {
...
 			dur := int(pawDur[c.TalentLvlSkill()] * 60)
 			exist := c.Core.Player.Shields.Get(shield.DionaSkill)
-			var shd *shield.Tmpl
+
 			if exist != nil {
 				// update
 				shd, _ = exist.(*shield.Tmpl)
+				if shd.Src != skillSrc {
+					shd.HP = shdHp
+					shd.Src = skillSrc
```
Charlie-Zheng, line 122: Since you update the shield Src, does this affect how effective Shield HP is logged?

jwalk-511: I don't know, and I didn't bother to find out since https://github.com/genshinsim/gcsim/blob/e560c41a4480ac41f115ce36c922c10eaa48171b/internal/characters/diona/skill.go#L102 is still an open question, and I don't know how Shield HP should be logged either way

Charlie-Zheng:
Seems like the effective shield HP is lower than it used to be when Sac Bow is used? It's the same as before without sac bow.
DB sim: https://gcsim.app/db/cQnQrkWtkdHJ
image

From this PR: https://gcsim.app/sh/n79GJPrWQWpJ
image

jwalk-511:
> Seems like the effective shield HP is lower than it used to be when Sac Bow is used? It's the same as before without sac bow. DB sim: https://gcsim.app/db/cQnQrkWtkdHJ image
> 
> From this PR: https://gcsim.app/sh/n79GJPrWQWpJ image

As far as I can tell, this is purely because Diona has less shield uptime with the bugfix, not due to any way I changed SRC.
When using Diona Sac Bow in the current version of SIM, I get the higher effective HP when I allow it to have full uptime, and lower effective HP if I artificially stretch the duration of the sim, regardless of whether I use 1 or 2 skills.

2 skills, full uptime. 12,185 eHP (12s duration)
https://gcsim.app/sh/RzL9C7ChMrHg
1 Skill, full uptime. 12,151 eHP (12s)
https://gcsim.app/sh/RzL9C7ChMrHg

2 Skills, half uptime. 6200 eHP (48s because bug)
https://gcsim.app/sh/nnFj7mb7qhT6
1 Skill, half uptime. 6238 eHP (24s duration)
https://gcsim.app/sh/CNMb7D9rBLkr

Charlie-Zheng:
That looks reasonable then.

Charlie-Zheng (approved):

PR #2591, opened 2026-03-29
Add timeout to damage mode sims to prevent common infinite loops

jwalk-511:
- Simulation config, duration field used as a timeout setting for damage mode, instead of being ignored.
- Sim currently sets timeout to 10 minutes with damage mode, no matter what user specifies.
- This can be changed to use whatever timer the user puts in their config, if desired.
- Default can be changed, if desired.
- A warning can be added to results when timeout occurs, if desired.

The purpose of this PR is to stop simulations early when infinite actions are provided to the gcsl parser. This PR does not attempt to stop other kinds of infinite loops, or act as a generic memory watchdog.

### DIFF
--- a/pkg/core/info/actionlist.go
+++ b/pkg/core/info/actionlist.go
@@ -42,6 +42,7 @@ type HurtSettings struct {
 
 type SimulatorSettings struct {
 	Duration          float64 `json:"-"`
+	TimeoutFrames     int     `json:"timeout_frames"`
 	DamageMode        bool    `json:"damage_mode"`
 	EnableHitlag      bool    `json:"enable_hitlag"`
 	DefHalt           bool    `json:"def_halt"` // for hitlag
--- a/pkg/core/info/actionlist.go
+++ b/pkg/core/info/actionlist.go
@@ -41,12 +41,12 @@ type HurtSettings struct {
 }
 
 type SimulatorSettings struct {
-	Duration          float64 `json:"-"`
-	TimeoutFrames     int     `json:"timeout_frames"`
-	DamageMode        bool    `json:"damage_mode"`
-	EnableHitlag      bool    `json:"enable_hitlag"`
-	DefHalt           bool    `json:"def_halt"` // for hitlag
-	IgnoreBurstEnergy bool    `json:"ignore_burst_energy"`
+	Duration           float64 `json:"-"`
+	DamageModeDuration float64 `json:"timeout_frames"`
+	DamageMode         bool    `json:"damage_mode"`
+	EnableHitlag       bool    `json:"enable_hitlag"`
+	DefHalt            bool    `json:"def_halt"` // for hitlag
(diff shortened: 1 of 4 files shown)


jwalk-511:
Current behavior with 3 commits
- If damage mode is not used, `timeout_frames` is ignored. `duration` is used to specify runtime, defaults to 90s.
- If damage mode is used, `duration` is ignored. `timeout_frames` is used to specify the maximum number of frames to be simulated, defaults to 10 minutes.

pkg/core/info/actionlist.go
```diff
@@ -41,11 +41,12 @@ type HurtSettings struct {
 }
 
 type SimulatorSettings struct {
-	Duration          float64 `json:"-"`
-	DamageMode        bool    `json:"damage_mode"`
-	EnableHitlag      bool    `json:"enable_hitlag"`
-	DefHalt           bool    `json:"def_halt"` // for hitlag
-	IgnoreBurstEnergy bool    `json:"ignore_burst_energy"`
+	Duration           float64 `json:"-"`
+	DamageModeDuration float64 `json:"timeout_frames"`
```
Matoba004: Figure out what we want to call this. In any case if we want it to be DamageModeDuration the json should match parseOptions. (but linter already gave this warning)

pkg/simulation/run.go
```diff
@@ -109,12 +109,19 @@ func (s *Simulation) popQueue() int {
 
 func initialize(s *Simulation) (stateFn, error) {
 	go s.eval.Start()
+
+	s.C.Flags.DamageMode = s.cfg.Settings.DamageMode
```
Matoba004, line 113: Is there a reason why this had to move?

pkg/simulation/run.go
```diff
@@ -109,12 +109,19 @@ func (s *Simulation) popQueue() int {
...
 	// run sim for 90s if no duration set
 	if s.cfg.Settings.Duration == 0 {
 		// fmt.Println("no duration set, running for 90s")
 		s.cfg.Settings.Duration = 90
 	}
-	s.C.Flags.DamageMode = s.cfg.Settings.DamageMode
+
+	// Timeout frames are currently only used in damage mode
+	if s.cfg.Settings.DamageModeDuration == 0 {
+		s.cfg.Settings.DamageModeDuration = 10 * 60
```
Matoba004, line 124: Retype as 600, this seems like it's 10s instead of 600s.

pkg/simulation/run.go
jwalk-511: For perception reasons; if damage mode is present it takes precedence over a duration being set. Moving this earlier reflects this logic, though programmatically it makes no difference

pkg/simulation/run.go
jwalk-511: Changed comments to reflect it's measured in seconds instead of frames

Charlie-Zheng (dismissed):

Charlie-Zheng (approved):

PR #2595, opened 2026-04-19
Implement lunar-crystallize

TheSast:
An implementation of lunar-crystallize.
Some of these changes are in preparation of #2596

### DIFF
--- a/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
+++ b/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/genshinsim/gcsim/pkg/core"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
+	"github.com/genshinsim/gcsim/pkg/core/construct"
 	"github.com/genshinsim/gcsim/pkg/core/event"
 	"github.com/genshinsim/gcsim/pkg/core/info"
 	"github.com/genshinsim/gcsim/pkg/core/keys"
@@ -51,6 +52,7 @@ func NewSet(c *core.Core, char *character.CharWrapper, count int, param map[stri
 
 	if count >= 4 {
 		c.Events.Subscribe(event.OnShielded, s.OnShielded(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
+		c.Events.Subscribe(event.OnConstructSpawned, s.OnConstructSpawned(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnShieldBreak, s.OnShieldBreak(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnCharacterSwap, s.OnCharacterSwap(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnSkill, s.OnSkill(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
@@ -67,6 +69,18 @@ func (s *Set) OnShielded() func(args ...any) {
 		}
 		if shd.Type() == shield.Crystallize {
 			s.lastF = shd.Expiry()
+		}
+	}
+}
+
+func (s *Set) OnConstructSpawned() func(args ...any) {
+	return func(args ...any) {
+		c := args[0].(construct.Construct)
+		if s.core.Player.Active() != s.char.Index() {
+			return
+		}
+		if c.Type() == construct.GeoConstructLunarCrystallize {
+			s.lastF = c.Expiry()
 		}
 	}
 }
--- a/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
+++ b/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/genshinsim/gcsim/pkg/core"
-	"github.com/genshinsim/gcsim/pkg/core/attacks"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 	"github.com/genshinsim/gcsim/pkg/core/construct"
 	"github.com/genshinsim/gcsim/pkg/core/event"
@@ -13,7 +12,6 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/player/character"
 	"github.com/genshinsim/gcsim/pkg/core/player/shield"
 	"github.com/genshinsim/gcsim/pkg/modifier"
-	"github.com/genshinsim/gcsim/pkg/reactable"
 )
 
 func init() {
@@ -21,11 +19,12 @@ func init() {
 }
 
 type Set struct {
-	Index int
-	core  *core.Core
-	char  *character.CharWrapper
-	lastF int
-	Count int
+	Index      int
+	core       *core.Core
+	char       *character.CharWrapper
+	tickerSrcF int
+	lastF      int
+	Count      int
 }
 
 func (s *Set) SetIndex(idx int) { s.Index = idx }
@@ -54,8 +53,6 @@ func NewSet(c *core.Core, char *character.CharWrapper, count int, param map[stri
 
 	if count >= 4 {
 		c.Events.Subscribe(event.OnShielded, s.OnShielded(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
-		c.Events.Subscribe(event.OnConstructSpawned, s.OnConstructSpawned(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
-		c.Events.Subscribe(event.OnLunarReactionAttack, s.OnLunarReactionAttack(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnShieldBreak, s.OnShieldBreak(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnCharacterSwap, s.OnCharacterSwap(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnSkill, s.OnSkill(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
@@ -72,30 +69,6 @@ func (s *Set) OnShielded() func(args ...any) {
 		}
 		if shd.Type() == shield.Crystallize {
 			s.lastF = shd.Expiry()
-		}
-	}
-}
-
-func (s *Set) OnConstructSpawned() func(args ...any) {
-	return func(args ...any) {
-		c := args[0].(construct.Construct)
-		if c.Type() == construct.GeoConstructLunarCrystallize {
-			s.core.Status.Add("nighttimewhispers-4pc-moondrift", reactable.lcrDur)
-			if s.core.Player.Active() != s.char.Index() {
-				s.lastF = s.core.F + s.core.Status.Duration("nighttimewhispers-4pc-moondrift")
-			}
-		}
-	}
-}
-
-func (s *Set) OnLunarReactionAttack() func(args ...any) {
-	return func(args ...any) {
-		ae := args[1].(info.AttackEvent)
-		if ae.Info.AttackTag == attacks.AttackTagReactionLunarCrystallize && s.core.Status.Duration("nighttimewhispers-4pc-moondrift") > 0 { // should be impossible for a moondrift to be absent during a moondrift attack?
-			s.core.Status.Add("nighttimewhispers-4pc-moondrift", reactable.lcrDur)
-			if s.core.Player.Active() != s.char.Index() {
-				s.lastF = s.core.F + s.core.Status.Duration("nighttimewhispers-4pc-moondrift")
-			}
 		}
 	}
 }
@@ -124,6 +97,8 @@ func (s *Set) OnCharacterSwap() func(args ...any) {
 		}
 		switch s.char.Index() {
 		case active:
+			s.tickerSrcF = s.core.F
+			s.moondrifTicker(s.tickerSrcF)
 			s.lastF = max(shd.Expiry(), moondriftExpiry)
 		case prev:
 			s.lastF = s.core.F + 60
@@ -137,6 +112,8 @@ func (s *Set) OnSkill() func(args ...any) {
 		if s.core.Player.Active() != s.char.Index() {
 			return
 		}
+		s.tickerSrcF = s.core.F
+		s.moondrifTicker(s.tickerSrcF)
 		s.char.AddStatMod(character.StatMod{
 			Base:         modifier.NewBaseWithHitlag("nighttimewhispers-4pc", 10*60),
 			AffectedStat: attributes.GeoP,
@@ -150,4 +127,33 @@ func (s *Set) OnSkill() func(args ...any) {
 			},
 		})
 	}
+}
+
+func (s *Set) moondrifTicker(src int) {
+	if s.tickerSrcF != src {
+		return
+	}
+	if s.core.Player.Active() != s.char.Index() {
+		return
+	}
+	if !s.char.StatModIsActive("nighttimewhispers-4pc") {
+		return
+	}
+	moondriftNearby := false
+	moondrifts, _ := s.core.Constructs.ConstructsByType(construct.GeoConstructLunarCrystallize)
+	playerPos := s.core.Combat.Player().Pos()
+	for _, moondrift := range moondrifts {
+		if playerPos.Distance(moondrift.Pos()) < 20 {
+			moondriftNearby = true
+			break
+		}
+	}
+	if moondriftNearby {
+		s.core.Status.Add("nighttimewhispers-4pc-moondrift", 60*2)
+		s.lastF = s.core.F + 60*2
+	} else if s.core.Status.Duration("nighttimewhispers-4pc-moondrift") > 0 {
+		s.core.Status.Add("nighttimewhispers-4pc-moondrift", 60)
+		s.lastF = s.core.F + 60
+	}
+	s.core.Tasks.Add(func() { s.moondrifTicker(src) }, 60*2)
 }
--- a/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
+++ b/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/genshinsim/gcsim/pkg/core"
+	"github.com/genshinsim/gcsim/pkg/core/attacks"
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 	"github.com/genshinsim/gcsim/pkg/core/construct"
 	"github.com/genshinsim/gcsim/pkg/core/event"
@@ -12,6 +13,7 @@ import (
 	"github.com/genshinsim/gcsim/pkg/core/player/character"
 	"github.com/genshinsim/gcsim/pkg/core/player/shield"
 	"github.com/genshinsim/gcsim/pkg/modifier"
+	"github.com/genshinsim/gcsim/pkg/reactable"
 )
 
 func init() {
@@ -53,6 +55,7 @@ func NewSet(c *core.Core, char *character.CharWrapper, count int, param map[stri
 	if count >= 4 {
 		c.Events.Subscribe(event.OnShielded, s.OnShielded(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnConstructSpawned, s.OnConstructSpawned(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
+		c.Events.Subscribe(event.OnLunarReactionAttack, s.OnLunarReactionAttack(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnShieldBreak, s.OnShieldBreak(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnCharacterSwap, s.OnCharacterSwap(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
 		c.Events.Subscribe(event.OnSkill, s.OnSkill(), fmt.Sprintf("nighttimewhispers-4pc-%v", char.Base.Key.String()))
@@ -76,11 +79,23 @@ func (s *Set) OnShielded() func(args ...any) {
 func (s *Set) OnConstructSpawned() func(args ...any) {
 	return func(args ...any) {
 		c := args[0].(construct.Construct)
-		if s.core.Player.Active() != s.char.Index() {
-			return
-		}
 		if c.Type() == construct.GeoConstructLunarCrystallize {
-			s.lastF = c.Expiry()
+			s.core.Status.Add("nighttimewhispers-4pc-moondrift", reactable.lcrDur)
+			if s.core.Player.Active() != s.char.Index() {
+				s.lastF = s.core.F + s.core.Status.Duration("nighttimewhispers-4pc-moondrift")
+			}
+		}
+	}
+}
+
+func (s *Set) OnLunarReactionAttack() func(args ...any) {
+	return func(args ...any) {
+		ae := args[1].(info.AttackEvent)
+		if ae.Info.AttackTag == attacks.AttackTagReactionLunarCrystallize && s.core.Status.Duration("nighttimewhispers-4pc-moondrift") > 0 { // should be impossible for a moondrift to be absent during a moondrift attack?
+			s.core.Status.Add("nighttimewhispers-4pc-moondrift", reactable.lcrDur)
+			if s.core.Player.Active() != s.char.Index() {
+				s.lastF = s.core.F + s.core.Status.Duration("nighttimewhispers-4pc-moondrift")
+			}
 		}
 	}
 }
@@ -103,12 +118,13 @@ func (s *Set) OnCharacterSwap() func(args ...any) {
 		prev := args[0].(int)
 		active := args[1].(int)
 		shd := s.core.Player.Shields.Get(shield.Crystallize)
+		moondriftExpiry := s.core.F + s.core.Status.Duration("nighttimewhispers-4pc-moondrift")
 		if shd == nil {
 			return
 		}
 		switch s.char.Index() {
 		case active:
-			s.lastF = shd.Expiry()
+			s.lastF = max(shd.Expiry(), moondriftExpiry)
 		case prev:
 			s.lastF = s.core.F + 60
 		}
--- a/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
+++ b/internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
@@ -98,7 +98,7 @@ func (s *Set) OnCharacterSwap() func(args ...any) {
 		switch s.char.Index() {
 		case active:
 			s.tickerSrcF = s.core.F
-			s.moondrifTicker(s.tickerSrcF)
+			s.moondriftTicker(s.tickerSrcF)
 			s.lastF = max(shd.Expiry(), moondriftExpiry)
 		case prev:
 			s.lastF = s.core.F + 60
@@ -113,7 +113,7 @@ func (s *Set) OnSkill() func(args ...any) {
 			return
 		}
 		s.tickerSrcF = s.core.F
-		s.moondrifTicker(s.tickerSrcF)
+		s.moondriftTicker(s.tickerSrcF)
 		s.char.AddStatMod(character.StatMod{
 			Base:         modifier.NewBaseWithHitlag("nighttimewhispers-4pc", 10*60),
 			AffectedStat: attributes.GeoP,
@@ -129,7 +129,7 @@ func (s *Set) OnSkill() func(args ...any) {
 	}
 }
 
-func (s *Set) moondrifTicker(src int) {
+func (s *Set) moondriftTicker(src int) {
 	if s.tickerSrcF != src {
 		return
 	}
@@ -155,5 +155,5 @@ func (s *Set) moondrifTicker(src int) {
 		s.core.Status.Add("nighttimewhispers-4pc-moondrift", 60)
 		s.lastF = s.core.F + 60
 	}
-	s.core.Tasks.Add(func() { s.moondrifTicker(src) }, 60*2)
+	s.core.Tasks.Add(func() { s.moondriftTicker(src) }, 60*2)
 }
--- a/internal/characters/gorou/cons.go
+++ b/internal/characters/gorou/cons.go
@@ -61,9 +61,23 @@ func (c *char) c1() {
 
 // C2:
 // While General's Glory is in effect, its duration is extended by 1s when a nearby
-// active character obtains an Elemental Shard from a Crystallize reaction.
+// active character obtains an Elemental Shard from a Crystallize reaction, or triggers a Lunar-Crystallize reaction.
 // This effect can occur once every 0.1s. Max extension is 3s.
 func (c *char) c2() {
+	c.Core.Events.Subscribe(event.OnLunarCrystallize, func(args ...any) {
+		if c.Core.Status.Duration(generalGloryKey) <= 0 {
+			return
+		}
+		ae := args[1].(info.AttackEvent)
+		if ae.Info.ActorIndex != c.Core.Player.Active() {
+			return
+		}
+		if c.c2Extension >= 3 {
+			return
+		}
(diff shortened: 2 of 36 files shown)


Charlie-Zheng:
We don't use stuff from wfpsim

internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
```diff
@@ -71,6 +73,18 @@ func (s *Set) OnShielded() func(args ...any) {
 	}
 }
 
+func (s *Set) OnConstructSpawned() func(args ...any) {
```
Charlie-Zheng, line 76: I think this is incorrect? The construct expiry can get extended but this wouldn't get updated

internal/characters/gorou/cons.go
```diff
@@ -61,9 +61,24 @@ func (c *char) c1() {
 
 // C2:
 // While General's Glory is in effect, its duration is extended by 1s when a nearby
-// active character obtains an Elemental Shard from a Crystallize reaction.
+// active character obtains an Elemental Shard from a Crystallize reaction, or triggers a Lunar-Crystallize reaction.
 // This effect can occur once every 0.1s. Max extension is 3s.
 func (c *char) c2() {
+	c.Core.Events.Subscribe(event.OnLunarCrystallize, func(args ...any) {
```
Charlie-Zheng, line 67: Add a check for enemy
```
		if _, ok := args[0].(*enemy.Enemy); !ok {
			return
		}
```

internal/weapons/claymore/verdict/verdict.go
```diff
@@ -72,6 +72,19 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p info.WeaponProfile)
...
+	c.Events.Subscribe(event.OnLunarCrystallize, func(args ...any) {
+		if !char.StatModIsActive(buffKey) {
+			w.stacks = 0
+		}
+		if w.stacks < 2 {
+			w.stacks++
+		}
+		c.Log.NewEvent("verdict adding stack", glog.LogWeaponEvent, char.Index()).
+			Write("stacks", w.stacks)
+		char.AddStatus(buffKey, buffDuration, true)
```
Charlie-Zheng, line 85: Refactor lines 77-85 and 64-72 into a function or lambda

pkg/core/event/event.go
```diff
@@ -74,7 +75,10 @@ const (
 	OnPlunge       // nil
 	OnAimShoot     // nil
 	OnDash
-	OnLunarChargedReactionAttack // target, AttackEvent; event so predamagemods can be applied to the individual LC contributions. Emitted once per contributor
+	// OnLunarReactionAttack            // target, AttackEvent; event so predamagemods can be applied to the individual Lunar contributions. Emitted once per contributor // TODO: move to a lunar.go ?
+	OnLunarChargedReactionAttack     // target, AttackEvent; event so predamagemods can be applied to the individual LC contributions. Emitted once per contributor
+	OnLunarCrystallizeReactionAttack // target, AttackEvent; event so predamagemods can be applied to the individual LCr contributions. Emitted once per contributor
```
Charlie-Zheng: Replace `OnLunarChargedReactionAttack` and `OnLunarCrystallizeReactionAttack` for just `OnLunarReactionAttack`. Subscribers can filter on the AttackTag if they need only a specific reaction. This means you have to go through Ineffa/Flins and maybe their sigs to add filtering based on AttackTag

Charlie-Zheng (changes requested):

Charlie-Zheng:
I didn't look at the actual implementation of LCr, since I'm not too familiar with it. I see that in Linnea's cons she can trigger Moondrift outside of the reactable. You could consider making the Moondrift trigger a free/floating function (so not tied to the reactable) and you pass in the contributor and/or their stats.

Basically take change the function signature, and apply any needed changes to pass in any required information
```diff
- func (r *Reactable) DoLCrAttackWithContrib(contribMap [info.MaxChars]bool) {
+ func DoLCrAttackWithContrib(contrib [info.MaxChars]*character.CharacterWrapper) {
```

TheSast:
> You could consider making the Moondrift trigger a free/floating function (so not tied to the reactable) and you pass in the contributor and/or their stats.
> 
> Basically take change the function signature, and apply any needed changes to pass in any required information
> 
> ```diff
> - func (r *Reactable) DoLCrAttackWithContrib(contribMap [info.MaxChars]bool) {
> + func DoLCrAttackWithContrib(contrib [info.MaxChars]*character.CharacterWrapper) {
> ```

The attack is targeted at an enemy, for which I use the Reactable, thus I need it still so that I have target for the attack pattern of the LCr attack.
I'll ask someone with C2 Linnea to test whether or not a Moondrift Harmony is actually triggered without an enemy.

Charlie-Zheng:
Also merge in the most recent update on main. You haven't given maintainers access to your branch for this PR, so you'll have to keep on top of merging in any updates from main. Please add support for Columbina's A4 causing each moondrift to have a 33% chance to trigger an extra time using a `LcrExtraHitOverride` flag.

Please use the formatter when making PRs: `go fmt`

TheSast:
> Please add support for Columbina's A4 causing each moondrift to have a 33% chance to trigger an extra time using a `LcrExtraHitOverride` flag.

Support for it is already present. Is it not functioning? Does it need to be altered?
```go
// Perform a Lunar Crystallize reaction 3-hit attack with the given contributors
func (r *Reactable) DoLCrAttackWithContrib(contribMap [info.MaxChars]bool) {
	for _, delay := range []int{1, 4, 7} {
		r.core.Tasks.Add(func() { r.doSingleLCrAttack(contribMap) }, delay)
		if chance, ok := r.core.Flags.Custom[LcrExtraHitOverride]; ok && r.core.Rand.Float64() < chance {
			r.core.Tasks.Add(func() { r.doSingleLCrAttack(contribMap) }, delay)
		}
	}
}
```
> Please use the formatter when making PRs: `go fmt`

I thought I had run it before pushing, I'll push the formatted version right now.

Charlie-Zheng:
Oh I see the flag `LcrExtraHitOverride`, that should be sufficient. Sorry I was looking in the wrong spot for the implementation earlier.

Charlie-Zheng:
> > You could consider making the Moondrift trigger a free/floating function (so not tied to the reactable) and you pass in the contributor and/or their stats.
> > Basically take change the function signature, and apply any needed changes to pass in any required information
> > ```diff
> > - func (r *Reactable) DoLCrAttackWithContrib(contribMap [info.MaxChars]bool) {
> > + func DoLCrAttackWithContrib(contrib [info.MaxChars]*character.CharacterWrapper) {
> > ```
> 
> The attack is targeted at an enemy, for which I use the Reactable, thus I need it still so that I have target for the attack pattern of the LCr attack. I'll ask someone with C2 Linnea to test whether or not a Moondrift Harmony is actually triggered without an enemy.

You can also pass in a target in the signature of the function
```
func DoLCrAttackWithContrib(contrib [info.MaxChars]*character.CharacterWrapper, target info.Target)
```

Charlie-Zheng:
Are you going to be fixing the issues for `nighttimewhispersintheechoingwoods.go` etc? If not I will take ownership of the this PR to do those fixes, since I believe you wanted to focus on the changes required for linnea

internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
TheSast: I also wasn't setting the correct expiry for the effect once the character swapped back on field.

internal/characters/gorou/cons.go
TheSast: What is this check meant to guard against? I don't think the ability requires an enemy to be present.

Charlie-Zheng: Does not trigger on reactions that occurred on gadgets

internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
```diff
@@ -71,6 +76,30 @@ func (s *Set) OnShielded() func(args ...any) {
...
```
Charlie-Zheng: Based on my reading of the DM, how this set works is that on every 2s (starts when buff is added, doesn't check when off field, restarts timer on switch in, stop checking when the base buff is removed), it checks for a moondrift. If a moondrift is found it sets the Moondrift nearby flag to 1 (when the base buff expires it sets the flag back to 0). If a moondrift is not found it sets the Moondrift nearby flag to 0 after a 1s delay.

internal/characters/mona/cons.go
```diff
@@ -161,7 +164,15 @@ func (c *char) c4() {
 		}
 
 		if c.Core.Flags.LogDebug {
-			c.Core.Log.NewEvent("Mona C4 CR added to Lunarcharged", glog.LogPreDamageMod, ae.Info.ActorIndex).
+			var reaction string
+			switch ae.Info.AttackTag {
+			case attacks.AttackTagDirectLunarCharged:
```
Charlie-Zheng: The attack tag of the LunarReactionAttacks are AttackTag**Reaction**LunarCharged etc, not Direct.

pkg/reactable/reactable.go
```diff
@@ -18,10 +18,12 @@ type Reactable struct {
...
 	// ec specific
-	ecAtk      info.AttackInfo // index of owner of next ec ticks
-	ecSnapshot info.Snapshot
-	ecTickSrc  int
-	lcAtkOwner int // index of owner of next LC tick
+	ecAtk           info.AttackInfo // index of owner of next ec ticks
+	ecSnapshot      info.Snapshot
+	ecTickSrc       int
+	lcAtkOwner      int                 // index of owner of next LC tick
+	lcrAtkOwner     int                 // index of owner of next LCr attack of 3 moondrifts
```
Charlie-Zheng: This isn't needed here. The owner is the trigger of the LcrAttack, which you should have access to in the AttackEvent when the reaction is triggered.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if a.Info.Durability < info.ZeroDur {
+		return false
+	}
+
+	if r.core.Status.Duration(LcrKey) > 0 {
+		r.extendLunarCrystallizeConstructDur()
+	} else {
+		// TODO: Check if constructs expiring will reset the counter
+		r.core.Constructs.NewNoLimitCons(r.newLunarCrystallizeConstruct(r.self.Direction(), r.self.Pos().Add(info.Point{Y: 1, X: 0})), false)
```
Charlie-Zheng: Add a TODO about the exact location of spawned moondrifts

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+func (r *Reactable) TryAddLCr(a *info.AttackEvent) bool {
+	if r.GetAuraDurability(info.ReactionModKeyHydro) <= info.ZeroDur {
+		return false
+	}
+
+	if a.Info.Durability < info.ZeroDur {
+		return false
+	}
+
+	if r.core.Status.Duration(LcrKey) > 0 {
```
Charlie-Zheng: This should actually check for Moondrifts in a 20 unit radius and spawn more if there are less than 3 moondrifts in that 20 unit radius.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+
+	r.core.Flags.Custom[lcrCountKey] += 1
+	r.core.Status.Add(LcrKey, LcrDur)
+	r.addLCrContributor(a)
+
+	if r.core.Flags.Custom[lcrCountKey] >= 3 {
+		// trigger three attacks
+		r.core.Flags.Custom[lcrCountKey] = 0
+		r.core.Events.Emit(event.OnMoondriftHarmony, r.self, &a)
+		r.core.Log.NewEvent("Lunar Crystallize attack triggered", glog.LogElementEvent, a.Info.ActorIndex)
```
Charlie-Zheng: Use "Moondrift Harmony triggered" instead

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// reduce
+	consumed := r.reduce(attributes.Hydro, a.Info.Durability, 0.5)
+	a.Info.Durability -= consumed
+	a.Info.Durability = max(a.Info.Durability, 0)
+	a.Reacted = true
+	r.lcrAtkOwner = a.Info.ActorIndex
+
+	// event
+	r.core.Events.Emit(event.OnLunarCrystallize, r.self, a)
```
Charlie-Zheng: I think this should be emitted before Moondrift Harmony event? Unless there is a specific reason it is being emitted after.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+
+	return true
+}
+
+// TODO this needs to be global?
```
Charlie-Zheng: Yeah the list of contributors needs to be made global. This is relevant for multi enemy simulations

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+	if r.GetAuraDurability(info.ReactionModKeyHydro) <= info.ZeroDur {
+		return false
+	}
+
+	if a.Info.Durability < info.ZeroDur {
+		return false
+	}
+
+	if r.core.Status.Duration(LcrKey) > 0 {
+		r.extendLunarCrystallizeConstructDur()
```
Charlie-Zheng: The extension should only occur when triggering moondrift harmony

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+	for charInd, dur := range r.Durability[info.ReactionModKeyHydro] {
+		if dur <= info.ZeroDur {
+			continue
+		}
+		r.lcrContributors[charInd] = true
+	}
+}
+
+func (r *Reactable) extendLunarCrystallizeConstructDur() {
+	matched, _ := r.core.Constructs.ConstructsByType(construct.GeoConstructLunarCrystallize)
```
Charlie-Zheng, line 116: The extension should only apply to Moondrifts in a 20 unit radius

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/genshinsim/gcsim/pkg/core/event"
+	"github.com/genshinsim/gcsim/pkg/core/glog"
+	"github.com/genshinsim/gcsim/pkg/core/info"
+)
+
+const (
+	LcrKey              = "lunarcrystallize"
+	LcrExtraHitOverride = "lunarcrystallize-bonus-hit-chance"
+	lcrCountKey         = "lunarcrystallize-count"
+	LcrDur              = 9 * 60
```
Charlie-Zheng: Why does LcrDur need to be public?

pkg/reactable/reactable.go
```diff
@@ -18,10 +18,12 @@ type Reactable struct {
...
+	lcrContributors [info.MaxChars]bool // patecipation map of the contributors of next LCr attack of 3 moondrifts
```
Charlie-Zheng: Possibly fix the spelling of `patecipation` when you move this to be global, if the comment is still needed. 

For making this global, I'd suggest using something like `r.core.Flags.Custom[fmt.Sprintf("%v-%i", lcrContribKey, ae.Info.ActorIndex]` or maybe encoding the contributors as bitwise flags by doing `r.core.Flags.Custom[lcrContribKey]`, casting to int and doing bitwise flag checks.

Charlie-Zheng (changes requested):
This is also missing changes for Archaic Petra 4pc. In the future, Cryo Resonance and Blizzard Strayer will need some changes to rename LunarChargedReactionAttack to LunarReactionAttack once I fix it (they are bugged and don't buff the Moondrifts Crit Rate). Technically they are correct for LC since LC cannot trigger when CryoRes/4BS are active.

pkg/reactable/lunarcrystallize.go
TheSast: This was specifically referring to turning `addLCrContributor` into `AddLCrContributor`. I need to further test to see how contributors are handled in multi enemy scenarios.

pkg/reactable/lunarcrystallize.go
TheSast: Of the target or the active character? I'll assume the latter

pkg/reactable/lunarcrystallize.go
TheSast: `nighttimewhispersintheechoingwoods.go` references it. If preferred I could inline it there instead, though, I think using a properly named constant is best.

internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
```diff
@@ -121,3 +129,32 @@ func (s *Set) OnSkill() func(args ...any) {
...
+	moondrifts, _ := s.core.Constructs.ConstructsByType(construct.GeoConstructLunarCrystallize)
+	playerPos := s.core.Combat.Player().Pos()
+	for _, moondrift := range moondrifts {
+		if playerPos.Distance(moondrift.Pos()) < 20 {
+			moondriftNearby = true
+			break
+		}
+	}
+	if moondriftNearby {
+		s.core.Status.Add("nighttimewhispers-4pc-moondrift", reactable.LcrDur)
```
Charlie-Zheng: The status duration is supposed to be infinite when there is moondrifts nearby. The LcrDur is not needed to be referenced here.

TheSast: Oh right, I forgot to change that when I altered the logic for the check.

pkg/reactable/lunarcrystallize.go
TheSast: Said file no longer references it.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	// reduce
+	consumed := r.reduce(attributes.Hydro, a.Info.Durability, 0.5)
+	a.Info.Durability -= consumed
+	a.Info.Durability = max(a.Info.Durability, 0)
+	a.Reacted = true
+
+	return true
+}
+
+func AddLCrContributor(charInd int, core *core.Core) {
```
Charlie-Zheng: How do you anticipate outside code needing to call `AddLCrContributor`?

Also they should be under the reactable struct instead of being free functions

```go
func (r *Reactable) AddLCrContributor(charInd int) {
    ...
}
```

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	core.Flags.Custom[fmt.Sprintf("%v-%v", lcrContributionKey, charInd)] = 1
+}
+
+func RemoveLCrContributor(charInd int, core *core.Core) {
```
Charlie-Zheng: How do you anticipate outside code needing to call `RemoveLCrContributor`?

Also they should be under the reactable struct instead of being free functions

```go
func (r *Reactable) RemoveLCrContributor(charInd int) {
    ...
}
```

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	core.Flags.Custom[fmt.Sprintf("%v-%v", lcrContributionKey, charInd)] = 0
+}
+
+func LCrContributors(core *core.Core) [info.MaxChars]bool {
```
Charlie-Zheng: How do you anticipate outside code needing to call `LCrContributors`?

Also they should be under the reactable struct instead of being free functions

```go
func (r *Reactable) LCrContributors() {
    ...
}
```

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (r *Reactable) extendNearbyLunarCrystallizeConstructDur() {
+	matched, _ := r.core.Constructs.ConstructsByType(construct.GeoConstructLunarCrystallize)
+	playerPos := r.core.Combat.Player().Pos()
+	for _, construct := range matched {
+		c, ok := (construct).(*skillConstruct)
+		if !ok {
+			continue
+		}
+		if playerPos.Distance(construct.Pos()) >= 20 {
```
Charlie-Zheng: Use `playerPos.Distance(construct.Pos()) > 20`

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+		snap := char.Snapshot(&ai)
+
+		ae := info.AttackEvent{
+			Info:        ai,
+			Pattern:     ap,
+			SourceFrame: core.F,
+			Snapshot:    snap,
+		}
+
+		// Emit even so PreDamageMods can be applied to the individual LCr contributions
```
Charlie-Zheng: Emit even**t**

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+			continue
+		}
+		c.expiry = r.core.F + lcrDur
+	}
+}
+
+type skillConstruct struct {
+	src    int
+	expiry int
+	react  *Reactable
```
Charlie-Zheng: the `react` field is not needed

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+			continue
+		}
+		if playerPos.Distance(construct.Pos()) >= 20 {
+			continue
+		}
+		c.expiry = r.core.F + lcrDur
+	}
+}
+
+type skillConstruct struct {
```
Charlie-Zheng: Rename the construct to `MoondriftConstruct` or `LunarCrystallizeConstruct` construct. It doesn't originate from a skill so this name isn't quite appropriate. Also move the construct code into a separate file will help with organization

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	playerPos := r.core.Combat.Player().Pos()
+	for _, moondrift := range moondrifts {
+		if playerPos.Distance(moondrift.Pos()) < 20 {
+			moondriftsNearby += 1
+			break
+		}
+	}
+	if moondriftsNearby < 3 {
+		for _, moondrift := range moondrifts {
+			r.core.Constructs.Destroy(moondrift.Key())
```
Charlie-Zheng, line 54: Why are these being destroyed? I believe the mechanic is that it just spawns some extras if there aren't enough in range, it doesn't destroy the moondrifts that are out of range

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+		return false
+	}
+
+	if a.Info.Durability < info.ZeroDur {
+		return false
+	}
+
+	r.core.Flags.Custom[lcrCountKey] += 1
+	r.core.Status.Add(LcrKey, lcrDur)
+	r.contributeToLCrWithAttack(a)
```
Charlie-Zheng: Based on my testing, the LCr that summons the moondrift doesn't count towards the contributors

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	if r.GetAuraDurability(info.ReactionModKeyHydro) <= info.ZeroDur {
+		return false
+	}
+
+	if a.Info.Durability < info.ZeroDur {
+		return false
+	}
+
+	r.core.Flags.Custom[lcrCountKey] += 1
+	r.core.Status.Add(LcrKey, lcrDur)
```
Charlie-Zheng: This status appears to be purely cosmetic. What functionality depends on this status? For users, they can see constructs being spawned in the `construct` section of the Sample.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+		for _, moondrift := range moondrifts {
+			r.core.Constructs.Destroy(moondrift.Key())
+		}
+		// TODO: Check if constructs expiring will reset the counter
+		// TODO: Set accurate location for spawned moondrifts
+		r.core.Constructs.NewNoLimitCons(r.newLunarCrystallizeConstruct(r.self.Direction(), r.self.Pos().Add(info.Point{Y: 1, X: 0})), false)
+		r.core.Constructs.NewNoLimitCons(r.newLunarCrystallizeConstruct(r.self.Direction(), r.self.Pos().Add(info.Point{Y: -0.5, X: 0.866})), false)
+		r.core.Constructs.NewNoLimitCons(r.newLunarCrystallizeConstruct(r.self.Direction(), r.self.Pos().Add(info.Point{Y: -0.5, X: -0.866})), false)
+	} else if r.core.Status.Duration(LcrKey) > 0 && r.core.Flags.Custom[lcrCountKey] >= 3 {
+		r.extendNearbyLunarCrystallizeConstructDur()
```
Charlie-Zheng: Why is this gated by the status? The duration always be extended when `r.core.Flags.Custom[lcrCountKey] >= 3`. You could move `r.extendNearbyLunarCrystallizeConstructDur()` into the code block that triggers the actual attacks

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+
+	return true
+}
+
+func AddLCrContributor(charInd int, core *core.Core) {
+	core.Flags.Custom[fmt.Sprintf("%v-%v", lcrContributionKey, charInd)] = 1
+}
+
+func RemoveLCrContributor(charInd int, core *core.Core) {
+	core.Flags.Custom[fmt.Sprintf("%v-%v", lcrContributionKey, charInd)] = 0
```
Charlie-Zheng: Move `fmt.Sprintf("%v-%v", lcrContributionKey, charInd)` into a function so it's not duplicated across `AddLCrContributor`, `RemoveLCrContributor` and `LCrContributors`

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+	r.core.Tasks.Add(func() {
+		for _, char := range r.core.Player.Chars() {
+			RemoveLCrContributor(char.Index(), r.core)
+		}
+	}, 7)
+}
+
+// Perform a Lunar Crystallize reaction 3-hit attack with the given contributors
+func DoLCrAttackWithContrib(contribMap [info.MaxChars]bool, target info.Target, core *core.Core, owner int) {
+	for _, delay := range []int{1, 4, 7} {
```
Charlie-Zheng: Make the hitmarks a global var. Also are these frame counted?

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func (r *Reactable) DoLCrAttack() {
+	DoLCrAttackWithContrib(LCrContributors(r.core), r.self, r.core, r.lcAtkOwner)
+	// clear contributors after last attack
+	r.core.Tasks.Add(func() {
+		for _, char := range r.core.Player.Chars() {
+			RemoveLCrContributor(char.Index(), r.core)
+		}
+	}, 7)
```
Charlie-Zheng: Access the last element of the hitmarks array instead. If you don't do hitmarks + 1, add a comment that this task needs to be after the DoLCrAttackWithContribto ensure the LCrContributors are removed after the last attack, instead of before.

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+}
+
+func doSingleLCrAttack(contribMap [info.MaxChars]bool, target info.Target, core *core.Core, owner int) {
+	contributions := []lcrContribution{}
+
+	ap := combat.NewSingleTargetHit(target.Key())
+
+	// Do we need to make a new one for each character?
```
Charlie-Zheng: I don't think this comment is needed?

pkg/reactable/lunarcrystallize.go
```diff
@@ -0,0 +1,276 @@
...
+
+const (
+	LcrKey              = "lunarcrystallize"
+	LcrExtraHitOverride = "lunarcrystallize-bonus-hit-chance"
+	lcrContributionKey  = "lunarcrystallize-contribution"
+	lcrCountKey         = "lunarcrystallize-count"
+	lcrDur              = 9 * 60
+)
+
+var lcrContributorMult = []float64{0.6, 0.3, 0.05, 0.05} // TODO: move to a lunar.go ?
```
Charlie-Zheng: it's defined separately in the DM so I think it's alright to be defined separately here. You can remove the comment

pkg/reactable/lunarcharged.go
```diff
@@ -19,7 +19,7 @@ const (
 	lcIcdKey         = "lunarcharged-cloud-icd"
 )
 
-var lcContributorMult = []float64{1.0, 1.0 / 2.0, 1.0 / 12.0, 1.0 / 12.0}
+var lcContributorMult = []float64{0.6, 0.3, 0.05, 0.05} // TODO: move to a lunar.go ?
```
Charlie-Zheng: it's defined separately in the DM so I think it's alright to be defined separately here. You can remove the comment

pkg/core/event/event.go
```diff
@@ -74,7 +75,8 @@ const (
 	OnPlunge       // nil
 	OnAimShoot     // nil
 	OnDash
-	OnLunarChargedReactionAttack // target, AttackEvent; event so predamagemods can be applied to the individual LC contributions. Emitted once per contributor
+	OnLunarReactionAttack            // target, AttackEvent; event so predamagemods can be applied to the individual Lunar contributions. Emitted once per contributor // TODO: move to a lunar.go ?
```
Charlie-Zheng: Remove the `// TODO: move to a lunar.go ?`, the event cannot be defined in a seperate file

pkg/core/event/event.go
```diff
@@ -41,14 +41,15 @@ const (
...
+	OnLunarCrystallize   // target, AttackEvent
 	OnShatter            // target, AttackEvent; at the end to simplify all reaction event subs since it's normally not considered as an elemental reaction
 	ReactionEventEndDelim
 	OnDendroCore // Gadget
 	// other stuff
 	OnStamUse           // abil
 	OnShielded          // shield
 	OnShieldBreak       // shield break
-	OnConstructSpawned  // nil
+	OnConstructSpawned  // construct
```
Charlie-Zheng: This doesn't need to be changed

internal/weapons/claymore/verdict/verdict.go
Charlie-Zheng: This one seems to be a duplicate?

internal/weapons/claymore/verdict/verdict.go
```diff
@@ -53,14 +53,35 @@ func NewWeapon(c *core.Core, char *character.CharWrapper, p info.WeaponProfile)
...
 		shd := args[0].(shield.Shield)
 		if shd.Type() != shield.Crystallize {
 			return
 		}
+		gainSeal()
+	}, fmt.Sprintf("verdict-seal-%v", char.Base.Key.String()))
+
+	// seal gain on lunar-crystallize trigger
+	c.Events.Subscribe(event.OnLunarCrystallize, func(args ...any) {
+		gainSeal()
```
Charlie-Zheng: This is missing the 1s ICD on gaining a seal through LCr

internal/characters/navia/skill.go
```diff
@@ -254,10 +254,16 @@ func (c *char) addShrapnelBuffs(snap *info.Snapshot, count int) {
 
 // shrapnelGain adds Shrapnel Stacks when a Crystallise Shield is picked up.
 // Stacks should last 300s but this is way too long to bother
-// When a character in the party obtains an Elemental Shard created from the Crystallize reaction,
+// When a character in the party obtains an Elemental Shard created from the Crystallize reaction, or when they trigger the Lunar-Crystallize reaction,
 // Navia will gain 1 Crystal Shrapnel charge. Navia can hold up to 6 charges of Crystal Shrapnel at once.
 // Each time Crystal Shrapnel gain is triggered, the duration of the Shards you have already will be reset.
 func (c *char) shrapnelGain() {
+	c.Core.Events.Subscribe(event.OnLunarCrystallize, func(args ...any) {
+		if c.shrapnel < 6 {
```
Charlie-Zheng: Should have an enemy filter
```go
if _, ok := args[0].(*enemy.Enemy); !ok {
    return
}
````

File: internal/characters/gorou/cons.go [hunk 39]
```diff
@@ -61,9 +62,26 @@ func (c *char) c1() {
...
```
Charlie-Zheng, line 68: I think you can reduce some code duplication here by moving the similar code into a hook

```go
hook := func() {
		if c.Core.Status.Duration(generalGloryKey) <= 0 {
			return
		}
		if c.c2Extension >= 3 {
			return
		}
		c.c2Extension++
		c.Core.Status.Extend(generalGloryKey, 60)
}
```

And the OnLunarCrystallize does the Enemy/ActiveChar check,
the OnShielded does the Crystallize shield check

internal/artifacts/nighttimewhispersintheechoingwoods/nighttimewhispersintheechoingwoods.go
```diff
@@ -121,3 +128,32 @@ func (s *Set) OnSkill() func(args ...any) {
 		})
 	}
 }
+
+func (s *Set) moondrifTicker(src int) {
```
Charlie-Zheng: should be moondrif**t**Ticker

Charlie-Zheng (changes requested):

Charlie-Zheng (commented):
Oh, also need to remove Columbina from the incomplete characters list in `pkg\result\check.go`

Charlie-Zheng:
Apply this patch

[add-lcr.patch](https://github.com/user-attachments/files/29168948/add-lcr.patch)

Also rebase onto the newest main after applying the patch

pkg/core/event/event.go
TheSast: I think this was left from an early implementation of the whispers set.

Charlie-Zheng (approved):

PR #2602, opened 2026-05-02
Update Nahida C2 Lunarbloom to attackMod

Charlie-Zheng:
Adjusted Nahida c2 logging to reflect CDMG buff for Lunarbloom


Charlie-Zheng:
Could also use this PR to fix Nahida C2 to use an attackMod. Not sure why it was original done using event subscription instead of attackMod?

Issue #2615, opened 2026-05-14
Make Attack Tag field in the sample a string instead of a number


Charlie-Zheng:
Currently the attack-tag field in the sample is a number, representing the internal AttackTag value, which is subject to change as more attack tags are added.

This should be changed to a string so it is useful for end consumers, and not subject to change when a new tag is added

<img width="621" height="656" alt="Image" src="https://github.com/user-attachments/assets/7034c671-9d3e-4f3b-ba36-4aacea205309" />

PR #2616, opened 2026-05-17
Add string functions for attack tags

jwalk-512:
func (i AttackTag) String() string - Returns string value for the attack tag

func AttackTagString(s string) (AttackTag, error) - returns attack tag given the named string

To generate
`go get github.com/dmarkham/enumer`
Go to directory containing file
Add comment `//go:generate enumer -type=AttackTag` if necessary (needed to auto-gen on other files, not needed ever again for attack tags) run `go generate` within the directory containing the file to generate on.

This should probably be added to Taskfile.yml but I'm not sure where it's appropriate to add it
The first half of supporting https://github.com/genshinsim/gcsim/issues/2615
If this is accepted, sample messages can me modified to call AttackTag.String() instead of just passing AttackTag


Charlie-Zheng:
Is it possible to configure it so that instead of it being named "attacktagskill" it can be just "skill"? If it needs the "attacktagskill", can it be snake case?

jwalk-512:
> Is it possible to configure it so that instead of it being named "attacktagskill" it can be just "skill"? If it needs the "attacktagskill", can it be snake case?

Done with `-trimprefix=AttackTag -transform=snake`

PR #2620, opened 2026-05-26
Added Disenchantment in Deep Shadow

Matoba004:
Artifact PR checklist:

- [x] New artifact package
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [ ] Update documentation


internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
```diff
@@ -0,0 +1,81 @@
...
+func (s *Set) SetIndex(idx int) { s.Index = idx }
+func (s *Set) GetCount() int    { return s.Count }
+func (s *Set) Init() error      { return nil }
+
+// 2pc - ATK +18%.
+// 4pc - When Normal Attacks hit opponents, there is a 36% chance that it will trigger Valley Rite, which will increase Normal Attack DMG by 70% of ATK.
+//
+//	This effect will be dispelled 0.05s after a Normal Attack deals DMG.
+//	If a Normal Attack fails to trigger Valley Rite, the odds of it triggering the next time will increase by 20%.
+//	This trigger can occur once every 0.2s.
```
Charlie-Zheng: Either delete this or correct it

internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
```diff
@@ -0,0 +1,81 @@
...
+func NewSet(core *core.Core, char *character.CharWrapper, count int, param map[string]int) (info.Set, error) {
+	s := Set{Count: count}
+
+	if count >= 2 {
```
Charlie-Zheng: Instead of
```
if count >= 2 {
    ...
}

if count >= 4 {
    ...
}
return
```
you can do 
```
if count < 2 {
    return
}
...
if count < 4 {
    return
}
....
return
}
```
This reduces the total amount of nesting

internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
```diff
@@ -0,0 +1,81 @@
...
+		m := make([]float64, attributes.EndStatType)
+		m[attributes.ATKP] = 0.18
+		char.AddStatMod(character.StatMod{
+			Base:         modifier.NewBase("echoes-2pc", -1),
```
Charlie-Zheng: it's not echos

internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
```diff
@@ -0,0 +1,81 @@
...
+		// TODO: check if it works while character is off-field
+
+		char.AddAttackMod(character.AttackMod{
+			Base: modifier.NewBase("disenchantment-on-superconduct-4pc", -1),
+			Amount: func(_ *info.AttackEvent, t info.Target) []float64 {
+				e, ok := t.(*enemy.Enemy)
+				if !ok {
+					return nil
+				}
+				if e.StatusIsActive("superconduct-phys-shred") {
```
Charlie-Zheng: Make the status a constant in superconduct.go and use it here, instead of using the string

pkg/shortcut/artifacts.go
```diff
@@ -31,6 +31,8 @@ var SetNameToKey = map[string]keys.Set{
 	"desertpavilionchronicle":            keys.DesertPavilionChronicle,
 	"desertpavilion":                     keys.DesertPavilionChronicle,
 	"dpc":                                keys.DesertPavilionChronicle,
+	"disenchantmentinthedeepshadow":      keys.DisenchantmentInTheDeepShadow,
+	"disenchantment":                     keys.DisenchantmentInTheDeepShadow,
```
Charlie-Zheng, line 35: Could also look at using `deepshadow` or something for the set. `disenchantment` is rather long. Not sure what shorthand people are using for the set right now

Charlie-Zheng (changes requested):

internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
Matoba004: I think I will have to leave this one to you.

Charlie-Zheng (approved):
Looks good. I'm pretty sure the crit rate works when the character is off field, but I haven't tested it

internal/artifacts/disenchantmentinthedeepshadow/disenchantmentinthedeepshadow.go
```diff
@@ -0,0 +1,81 @@
...
+				return 0
+			}
+
+			return 0.8
+		},
+	})
+
+	// TODO: check if it works while character is off-field
+
+	char.AddAttackMod(character.AttackMod{
```
Charlie-Zheng, line 67: I think you need to add the OnLunarReactionAttack or whatever here to make sure reaction LC gets the buff.

Charlie-Zheng (changes requested):

Charlie-Zheng (approved):
Will cause a build error when #2595 is merged due to OnLunarChargedReactionAttack being renamed to OnLunarReactionAttack. Otherwise looks good.

PR #2622, opened 2026-05-26
Add Celestial Gift

TheSast:
Artifact PR checklist:
- [x] New artifact package
- [x] Run pipeline with added config (generates .generated.json files)
- [x] Artifact key
- [x] 2pc
- [x] 4pc
- [x] Update documentation


internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+						}
+						buffStrength := 0.4
+						s.buff[attributes.EleToDmgP(s.core.Player.ActiveChar().Base.Element)] = buffStrength
+						return s.buff
+					},
+				})
+			}
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBaseWithHitlag("celestialgift-4pc-holder-"+attributes.ElementString[char.Base.Element], duration),
+				AffectedStat: attributes.NoStat,
```
Charlie-Zheng: This should be `AffectedStat: attributes.EleToDmgP(char.Base.Element)`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+		holder := char.Index()
+		if c.Player.Active() != holder {
+			return
+		}
+		duration := 60 * 20
+		for _, x := range s.core.Player.Chars() {
+			if s.core.Player.GetHexereiCount() >= 2 {
+				x.AddStatMod(character.StatMod{
+					Base:         modifier.NewBaseWithHitlag("celestialgift-4pc-activechar", duration),
+					AffectedStat: attributes.NoStat,
```
Charlie-Zheng: This should be `AffectedStat: s.core.Player.ActiveChar().Base.Element)`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+	c.Events.Subscribe(event.OnSkill, func(args ...any) {
+		holder := char.Index()
+		if c.Player.Active() != holder {
+			return
+		}
+		duration := 60 * 20
+		for _, x := range s.core.Player.Chars() {
+			if s.core.Player.GetHexereiCount() >= 2 {
+				x.AddStatMod(character.StatMod{
+					Base:         modifier.NewBaseWithHitlag("celestialgift-4pc-activechar", duration),
```
Charlie-Zheng: You should name the active char buff the same as the holder buff to prevent the same element buffs from stacking. You current code allows a pyro holder + a pyro on fielder to apply both buffs

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+func init() {
+	core.RegisterSetFunc(keys.CelestialGift, NewSet)
+}
+
+type Set struct {
+	char  *character.CharWrapper
+	core  *core.Core
+	Index int
+	Count int
+	buff  []float64
```
Charlie-Zheng: You don't need to store the char, core, Index, or buff fields, they don't get used anywhere

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+		return &s, nil
+	}
+
+	s.buff = make([]float64, attributes.EndStatType)
+	c.Events.Subscribe(event.OnSkill, func(args ...any) {
+		holder := char.Index()
+		if c.Player.Active() != holder {
+			return
+		}
+		duration := 60 * 20
```
Charlie-Zheng: The duration should be a const at the start of the file

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+		for _, x := range s.core.Player.Chars() {
+			if s.core.Player.GetHexereiCount() >= 2 {
```
Charlie-Zheng: use `isHexSecretRite := s.core.Player.GetHexereiCount() >= 2` and then use this bool in the various places you check if secret rite is active

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,119 @@
...
+						attributes.DendroP,
+						attributes.CryoP,
+						attributes.GeoP,
+					} {
+						s.buff[elementP] = 0
+					}
+					buffStrength := 0.2
+					if s.core.Player.GetHexereiCount() >= 2 {
+						buffStrength = 0.4
+					}
```
Charlie-Zheng: Do this check before doing AddStatMod so both the active char and the holder buff use the same `buffStrength` variable

Charlie-Zheng (changes requested):

internal/artifacts/celestialgift/celestialgift.go
TheSast: I'll assume you meant `AffectedStat: attributes.EleToDmgP(s.core.Player.ActiveChar().Base.Element)`

internal/artifacts/celestialgift/celestialgift.go
TheSast: `Index` and `Count` are required for this `Set` to be a valid `info.Set`

internal/artifacts/celestialgift/celestialgift.go
TheSast: The first `AddStatMod` is the buff based on the active character's element, which can only be active if Hexerei Secret Rite is active, meaning that it will always be 40%

Charlie-Zheng: Add

```
buffStrength := 0.2
if isHexereiSecretRite {
    buffStrength = 0.4
}
```
in between line 65-66.

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,121 @@
...
+		holder := char.Index()
+		if c.Player.Active() != holder {
+			return
+		}
+		isHexereiSecretRite := c.Player.GetHexereiCount() >= 2
+		for _, x := range c.Player.Chars() {
+			if isHexereiSecretRite {
+				activeCharDmgP := attributes.EleToDmgP(c.Player.ActiveChar().Base.Element)
+				x.AddStatMod(character.StatMod{
+					Base:         modifier.NewBaseWithHitlag("celestialgift-4pc-"+attributes.ElementString[x.Base.Element], duration),
```
Charlie-Zheng: Use `"celestialgift-4pc-"+x.Base.Element.String()`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,121 @@
...
+	s.buff = make([]float64, attributes.EndStatType)
+	c.Events.Subscribe(event.OnSkill, func(args ...any) {
+		holder := char.Index()
+		if c.Player.Active() != holder {
+			return
+		}
+		isHexereiSecretRite := c.Player.GetHexereiCount() >= 2
+		for _, x := range c.Player.Chars() {
+			if isHexereiSecretRite {
+				activeCharDmgP := attributes.EleToDmgP(c.Player.ActiveChar().Base.Element)
```
Charlie-Zheng: The active char buff needs to reapply on swap. Make a seperate subscription to OnSwap and apply the active buff there. Make sure you don't delete the holder's buff or any other holder's buffs.

Charlie-Zheng (changes requested):

internal/artifacts/celestialgift/celestialgift.go
TheSast: The current implementation doesn't require the `StatMod` to work correctly.

internal/artifacts/celestialgift/celestialgift.go
TheSast: Sure?

internal/artifacts/celestialgift/celestialgift.go
Charlie-Zheng: Raiden doesn't have Hydro% buff even though Aino is on field:
image
https://gcsim.app/sh/TDJCQhJbz6wQ
[sample.gz](https://github.com/user-attachments/files/28369790/sample.gz)

TheSast: With the latest implementation, it should now work correctly:
image
[sample.gz](https://github.com/user-attachments/files/28371955/sample.gz)

Charlie-Zheng: This works correctly in terms of giving the buff, but not in terms of hitlag extension. Currently you could have the buffs last different amounts of times on different chars based on their hitlag experienced

TheSast: Moved the buff duration tracking with hitlag to a status applied to the holder of the set.

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,110 @@
...
+			if c.StatusIsActive("celestialgift-4pc") {
+				return true
+			}
+		}
+		return false
+	}
+
+	s.buff = make([]float64, attributes.EndStatType)
+	c.Events.Subscribe(event.OnSkill, func(args ...any) {
+		if c.Player.Active() != char.Index() {
```
Charlie-Zheng: Add a message:
```
// TODO: Description says "The equipping character can trigger this effect while off-field."
// but currently no characters can use skills off field, so this is ignored
```

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+			holderElement := char.Base.Element
+			holderDmgP := attributes.EleToDmgP(holderElement)
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBase("celestialgift-4pc-"+holderElement.String(), -1),
+				AffectedStat: holderDmgP,
+				Amount: func() []float64 {
+					clear(s.buff)
+					s.buff[holderDmgP] = buffStrength
+					if !buffActive() {
```
Charlie-Zheng: This only checks that the buff is active on some character, not that the buff is active on the holder

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,110 @@
...
+		},
+	})
+
+	if count < 4 || !char.IsHexerei {
+		return &s, nil
+	}
+
+	buffActive := func() bool {
+		for _, c := range c.Player.Chars() {
+			if c.StatusIsActive("celestialgift-4pc") {
```
Charlie-Zheng: Make `"celestialgift-4pc"` a const

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,110 @@
...
+						}
+						s.buff[attributes.EleToDmgP(activeCharElement)] = buffStrength
+						return s.buff
+					},
+				})
+			}
+			holderElement := char.Base.Element
+			holderDmgP := attributes.EleToDmgP(holderElement)
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBase("celestialgift-4pc-"+holderElement.String(), -1),
```
Charlie-Zheng: Since this stat mod lasts forever, add it once on init instead of once every time the hold uses skill.

Also add this mod for every element on the team.

So use the below on init. Then on skill you just need to add the `celestial4Key` status to the holder
```
for _, c0 := range c.Player.Chars() {
  elem := c0.Base.Element
  dmgp := attributes.EleToDmgP(elem)
  for _, c1 := range c.Player.Chars() {
    c1.AddStatMod(character.StatMod {
       Base: modifier.NewBase("celestialgift-4pc-"+elem.String(), -1),
       AffectedStat: dmgp 
       Amount: func() [] {
         clear(s.buff)
         s.buff[dmgp] = buffStrength
         foundBuff := false
         for _, c := range c.Player.Chars() {
			if c.StatusIsActive(celestial4Key) {
			  if c.Base.Elem == elem {
				return s.buff
			  }
		      foundBuff = true
			}
          }
          if !foundBuff  {
             return nil
          } 
          if c.Player.HexereiCount < 2 {
             return nil
          }
          if c.Player.ActiveChar().Element != elem {
             return nil  
		  }
		return s.buff
      }
    }
  }
}
```

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,129 @@
...
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBase(set4Key+"-"+elem.String(), -1),
+				AffectedStat: buffedDmgP,
+				Amount: func() []float64 {
+					clear(s.buff)
+					s.buff[buffedDmgP] = buffStrength
+					if buffActiveOfElement(elem) {
+						return s.buff
+					}
+					if buffActive() && isHexereiSecretRite && c.Player.ActiveChar().Base.Element == elem {
```
Charlie-Zheng: do 

```
if isHexereiSecretRite && c.Player.ActiveChar().Base.Element == elem && buffActive()
```
since the buffActive check is more expensive.

Charlie-Zheng (changes requested):

Charlie-Zheng (approved):
LGTM

Charlie-Zheng (changes requested):
Due to more testing in game, it looks like my understanding of the artifact's mechanics was incorrect. I'm changing my review to request changes, but pending more TC I can't really give guidance on how exactly to change the code right now

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,165 @@
...
+							s.buff[buffedDmgP] = buffStrength
+							if holderBuffActiveOfElement(elem) {
+								return s.buff
+							}
+							return nil
+						},
+					})
+				}
+			}
+		}
```
jwalk-511: Since you are subscribing to skill and applying the active char buff on swap, you don't need to loop over elements here. You can just apply the buff for the setholder's element instead.

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,165 @@
...
+			return
+		}
+		char.AddStatus(set4Key, duration, true)
+		for _, x := range c.Player.Chars() {
+			x.DeleteStatus(set4KeyActive)
+		}
+		for elem := range elems {
+			buffedDmgP := attributes.EleToDmgP(elem)
+			for _, x := range c.Player.Chars() {
+				if !holderBuffActiveOfElement(elem) {
```
jwalk-511: A setholder should be able to refresh the duration of the buff if they were the one to initially apply the element buff. If following Shizuka's advice of not doing this, since it requires a hack to implement, you should allow any character to refresh the buff instead of no character. Additionally, a comment should be left documenting the discrepancy from in-game behavior.

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+		}
+	}, fmt.Sprintf(set4Key+"-%v", char.Base.Key.String()))
+
+	if !isHexereiSecretRite {
+		return &s, nil
+	}
+
+	c.Events.Subscribe(event.OnCharacterSwap, func(args ...any) {
+		if c.Player.Active() != char.Index() {
```
jwalk-511: buff should be reapplied on any char swap, not just the holder char swap

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,165 @@
...
+				}
+			}
+		}
+	}, fmt.Sprintf(set4Key+"-%v", char.Base.Key.String()))
+
+	if !isHexereiSecretRite {
+		return &s, nil
+	}
+
+	c.Events.Subscribe(event.OnCharacterSwap, func(args ...any) {
```
jwalk-511: OnSwap should immediately delete all previous set4Key-elem-active buffs, and then only add a new buff if allowed/not prevented

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,165 @@
...
+			return
+		}
+		if !holderBuffActive() {
+			return
+		}
+		char.AddStatus(set4KeyActive, duration, true)
+		elem := c.Player.ActiveChar().Base.Element
+		buffedDmgP := attributes.EleToDmgP(elem)
+		for _, x := range c.Player.Chars() {
+			if !holderBuffActiveOfElement(elem) {
```
jwalk-511: this can be bumped earlier; `activeBuffActive(elem)` will give the same result regardless of which char you're iterating through (no need to check 4 times).

jwalk-511 (changes requested):

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	elems := map[attributes.Element]bool{}
+	for _, x := range c.Player.Chars() {
+		elems[x.Base.Element] = true
+	}
+
+	buffStrength := 0.2
+	isHexereiSecretRite := c.Player.GetHexereiCount() >= 2
```
jwalk-511: this depends on other chars, so needs to be in `init` instead of `new`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,187 @@
...
+		clearActiveBuff()
+		elem := char.Base.Element
+		if holderBuffActiveOfElementNotFrom(elem, char.Index()) {
+			return
+		}
+		char.AddStatus(set4KeyHolder, duration, true)
+		buffedDmgP := attributes.EleToDmgP(elem)
+		for _, x := range c.Player.Chars() {
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBase(set4KeyHolder+"-"+elem.String(), -1),
```
jwalk-511: since this is a -1 duration buff Vespertine may want you to move it back to being applied once at init as it was before, though this isn't wrong.

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,187 @@
...
+		// but currently no characters can use skills while off field, so this is ignored
+		if c.Player.Active() != char.Index() {
+			return
+		}
+		clearActiveBuff()
+		elem := char.Base.Element
+		if holderBuffActiveOfElementNotFrom(elem, char.Index()) {
+			return
+		}
+		char.AddStatus(set4KeyHolder, duration, true)
```
jwalk-511: Referencing https://github.com/genshinsim/gcsim/pull/2622/changes#diff-2e679e37c9e96e9c3782a9635d0f745cd6fef9b4b6849470c11224b073dbbc5eR160

Whenever a setholder uses elemental skill, the timer for the active char buff should be refreshed to full (or overwritten as new).
This timer is the same as the equipping character's buff duration would be if it wasn't blocked. However, even if the equipping character's buff is blocked by another setholder of the same element, the active character buff is still allowed to be set to 20s

Charlie-Zheng (commented):
For documentation, you need to update `ui\packages\docs\src\components\Names\artifact_data.json` to contain the nicknames, and then, depending on how you decide to deal with the multiple users problem, add the issue to `ui\packages\docs\src\components\Issues\artifact_data.json`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+	count := s.Count
+	c := s.core
+	char := s.char
+
+	if count < 4 || !char.IsHexerei {
+		return nil
+	}
+
+	holderBuffActive := func() bool {
+		for _, c := range c.Player.Chars() {
```
Charlie-Zheng: Using `c` here is unclear because the original `c` variable referred to `*core.Core`, but it refers to a `*character.CharWrapper` here. Use like `other_char` or something

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+			if c.StatusIsActive(set4KeyHolder) {
+				return true
+			}
+		}
+		return false
+	}
+
+	holderBuffDuration := func() int {
+		x := 0
+		for _, c := range c.Player.Chars() {
```
Charlie-Zheng: Using `c` here is unclear because the original `c` variable referred to `*core.Core`, but it refers to a `*character.CharWrapper` here. Use like `other_char` or something

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+			x = max(x, c.StatusDuration(set4KeyHolder))
+		}
+		return x
+	}
+
+	holderBuffIsActiveOfElement := func(e attributes.Element) bool {
+		for _, c := range c.Player.Chars() {
```
Charlie-Zheng: Using `c` here is unclear because the original `c` variable referred to `*core.Core`, but it refers to a `*character.CharWrapper` here. Use like `other_char` or something

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+			if c.StatusIsActive(set4KeyHolder) && c.Base.Element == e {
+				return true
+			}
+		}
+		return false
+	}
+
+	holderBuffIsActiveOfElementNotFrom := func(e attributes.Element, i int) bool {
+		for _, x := range c.Player.Chars() {
```
Charlie-Zheng: Change `x` to match the `other_char` or whatever you change the earlier `c`s into

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+			if x.StatusIsActive(set4KeyHolder) && x.Base.Element == e && x.Index() != i {
+				return true
+			}
+		}
+		return false
+	}
+
+	activeBuffIsActive := func() bool {
+		for _, x := range c.Player.Chars() {
```
Charlie-Zheng: Change `x` to match the `other_char` or whatever you change the earlier `c`s into

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+			if x.StatusIsActive(set4KeyActive) {
+				return true
+			}
+		}
+		return false
+	}
+
+	clearActiveBuff := func() {
+		for _, x := range c.Player.Chars() {
```
Charlie-Zheng: Change `x` to match the `other_char` or whatever you change the earlier `c`s into

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+	holderBuffIsActiveOfElement := func(e attributes.Element) bool {
+		for _, c := range c.Player.Chars() {
+			if c.StatusIsActive(set4KeyHolder) && c.Base.Element == e {
+				return true
+			}
+		}
+		return false
+	}
+
+	holderBuffIsActiveOfElementNotFrom := func(e attributes.Element, i int) bool {
```
Charlie-Zheng: Add a comment above this function explaining what exactly it's trying to do
```
// returns true if there is another party member of the same element with the 4pc buff active,
// otherwise returns false
```

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+		for _, x := range c.Player.Chars() {
+			if x.StatusIsActive(set4KeyHolder) && x.Base.Element == e && x.Index() != i {
+				return true
+			}
+		}
+		return false
+	}
+
+	activeBuffIsActive := func() bool {
```
Charlie-Zheng: Add a comment
```
// active char buff is active if Hexerei Secret Rite is active and there is a party member with the 4pc buff
// don't need to check Secret Rite because the active char stat mod isn't added when not Secret Rite
```

internal/artifacts/celestialgift/celestialgift.go
TheSast: Vespertine stated that they prefer it to be absent over ineligible if possible

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+	clearActiveBuff := func() {
+		for _, x := range c.Player.Chars() {
+			x.DeleteStatus(set4KeyActive)
+		}
+	}
+
+	refreshActiveBuff := func() {
+		for _, x := range c.Player.Chars() {
+			x.AddStatus(set4KeyActive, duration, true)
+		}
```
Charlie-Zheng: Should only add the status to the holder

internal/artifacts/celestialgift/celestialgift.go
Charlie-Zheng: This function doesn't need to take in the element, and the index, since these are always the element and index of the `s.char`

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+	}, fmt.Sprintf(set4KeyHolder+"-%v", char.Base.Key.String()))
+
+	if !isHexereiSecretRite {
+		return nil
+	}
+
+	c.Events.Subscribe(event.OnCharacterSwap, func(args ...any) {
+		clearActiveBuff()
+		remainingDuration := holderBuffDuration()
```
Charlie-Zheng: This should be getting the activeBuff duration instead of the holder Buff duration

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		elem := c.Player.ActiveChar().Base.Element
+		if holderBuffIsActiveOfElement(elem) {
+			return
+		}
+		char.AddStatus(set4KeyActive, remainingDuration, true)
+		buffedDmgP := attributes.EleToDmgP(elem)
+		for _, x := range c.Player.Chars() {
+			x.AddStatMod(character.StatMod{
+				Base:         modifier.NewBase(set4KeyHolder+"-"+elem.String()+"-active", -1),
```
Charlie-Zheng: Buff duration should be based on the remaning duration

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+		clearActiveBuff()
+		remainingDuration := holderBuffDuration()
+		if remainingDuration == 0 {
+			return
+		}
+		elem := c.Player.ActiveChar().Base.Element
+		if holderBuffIsActiveOfElement(elem) {
+			return
+		}
+		char.AddStatus(set4KeyActive, remainingDuration, true)
```
Charlie-Zheng: This isn't needed

Charlie-Zheng:
[celestial-gift.patch](https://github.com/user-attachments/files/29184449/celestial-gift.patch)

After you rebase onto main to get access to Nicole and Durin, you can check the following test configs to see how the behaviour has changed:
https://gcsim.app/sh/7pFkj6ghqqPM
https://gcsim.app/sh/mDPRwwQPczdP
https://gcsim.app/sh/wcBh87fPRcQ9

internal/artifacts/celestialgift/celestialgift.go
```diff
@@ -0,0 +1,198 @@
...
+	activeBuffIsActive := func() bool {
+		for _, x := range c.Player.Chars() {
+			if x.StatusIsActive(set4KeyActive) {
+				return true
+			}
+		}
+		return false
+	}
+
+	clearActiveBuff := func() {
```
Charlie-Zheng: This function isn't actually removing the buff from characters

Charlie-Zheng (approved):

PR #2627, opened 2026-05-31
Implement Durin

Charlie-Zheng:
- [x] New character package
- [x] Character key
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation


internal/characters/durin/charge.go
```diff
@@ -0,0 +1,52 @@
...
+	chargeFrames = frames.InitAbilSlice(58)         // CA -> N1
+	chargeFrames[action.ActionAttack] = 53          // CA -> N1
+	chargeFrames[action.ActionSkill] = 52           // CA -> E
+	chargeFrames[action.ActionBurst] = 52           // CA -> Q
+	chargeFrames[action.ActionDash] = chargeHitmark // CA -> D
+	chargeFrames[action.ActionJump] = chargeHitmark // CA -> J
+	chargeFrames[action.ActionSwap] = 51            // CA -> Swap
+}
+
+func (c *char) ChargeAttack(p map[string]int) (action.Info, error) {
```
Matoba004, line 26: Durin can still perform N4C, which he shouldn't be able to.

internal/characters/durin/asc.go
```diff
@@ -0,0 +1,148 @@
...
+func (c *char) a4Dmg() float64 {
+	if c.Base.Ascension < 4 {
+		return 1.0
+	}
+
+	if c.a4stacks <= 0 {
+		return 1.0
+	}
+
+	if !c.StatusIsActive(blackKey) && !c.StatusIsActive(whiteKey) {
```
jwalk-511: Was this tested to be necessary? Wiki claims that Durin's default state is white mode, even if he doesn't use skill at all or his skill expires.
Talent description for burst also states `When Durin is not in a special state as a result of Essential Transmutation, he will use the Elemental Burst Principle of Purity: As the Light Shifts.`

jwalk-511 (changes requested):

internal/characters/durin/asc.go
Charlie-Zheng: This is for when burst expires. It's only relevant if user uses a very long travel time for the Black version I believe.

Edit: I see it's not checking the burst keys. I will fix that

internal/characters/durin/charge.go
Charlie-Zheng: It's a problem common to all sword chars (except furina). Should be fixed by a system level change

PR #2628, opened 2026-06-04
Add "fork" as an alias for Symphonist of Scents

jack5mikemotown:


srliao:
To keep shortcuts manageable, the "rule" is only two shortcuts per character/weapon/artifact. See: https://discord.com/channels/845087716541595668/884485336657444894/1220840757959589898

jack5mikemotown:
there is currently only one shortcut, "symphonist", the other one "symphonistofscents" is the weapon name, other weapons, for example, tfd, everlasting moonglow, mistsplitter, uraku, all have 2 shortcuts in addition to the weapon name

srliao:
the "rule" was put in place after the shortcuts and nicknames got out of hand with the older stuff

maybe this weapon can use another shortcut but I don't think `fork` is one of them

PR #2635, opened 2026-06-12
Added Athame Artis weapon

Nahbrowhy:
## Summary

Adds implementation for Athame Artis.

### Weapon PR checklist

- [x] New weapon package
- [x] Run pipeline with added config (generates weapon curve, .generated.json files)
- [x] Weapon key
- [x] Update documentation


internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+			return burstCDMod
+		},
+	})
+
+	// TODO:
+	// Passive specifies "when an Elemental Burst hits an opponent".
+	// Currently implemented using OnEnemyDamage as a conservative
+	// approximation until hit-vs-damage behavior is verified.
+
+	c.Events.Subscribe(event.OnEnemyDamage, func(args ...any) {
```
Charlie-Zheng: They both do the same thing. ATK% will snapshot before OnEnemyDamage or OnEnemyHit, so this comment is unnecessary

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+		atk := args[1].(*info.AttackEvent)
+
```
Charlie-Zheng, line 91: Add a check that the target is an enemy
```go
			t, ok := args[0].(*enemy.Enemy)
			if !ok {
				return
			}
```

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+		if atk.Info.AttackTag != attacks.AttackTagElementalBurst {
+			return
+		}
+
+		// Can trigger while off-field.
+
+		procSelfATK := selfATK
+		procTeamATK := teamATK
+
+		// TODO verify exact Hexerei condition
```
Charlie-Zheng: What exactly needs to be done for this?

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+		if c.Player.GetHexereiCount() >= 2 {
+			procSelfATK *= 1.75
+			procTeamATK *= 1.75
+		}
+
+		selfVal := make([]float64, attributes.EndStatType)
+		selfVal[attributes.ATKP] = procSelfATK
```
Charlie-Zheng: Create this outside of the callback so it's not constantly being allocated

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+		char.AddStatMod(character.StatMod{
+			Base:         modifier.NewBaseWithHitlag("athame-self-atk", 3*60),
+			AffectedStat: attributes.ATKP,
+			Amount: func() []float64 {
+				return selfVal
+			},
+		})
+
+		teamVal := make([]float64, attributes.EndStatType)
+		teamVal[attributes.ATKP] = procTeamATK
```
Charlie-Zheng: create this outside of the callback so it's not constantly being allocated

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		if atk.Info.AttackTag != attacks.AttackTagElementalBurst {
+			return
+		}
+
+		// Can trigger while off-field.
+
+		procSelfATK := selfATK
+		procTeamATK := teamATK
```
Charlie-Zheng: This is unnecessary

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (w *Weapon) SetIndex(idx int) { w.Index = idx }
+func (w *Weapon) Init() error      { return nil }
+
+func NewWeapon(
+	c *core.Core,
+	char *character.CharWrapper,
+	p info.WeaponProfile,
+) (info.Weapon, error) {
```
Charlie-Zheng: For consistency with other weapons, put this on one line

Charlie-Zheng (changes requested):

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,108 @@
...
+
+		if active != char.Index() {
+			c.Player.Chars()[active].AddStatMod(character.StatMod{
+				Base:         modifier.NewBaseWithHitlag("athameartis-team-atk", 3*60),
+				AffectedStat: attributes.ATKP,
+				Amount: func() []float64 {
+					return teamVal
+				},
+			})
+		}
```
Charlie-Zheng: Remove this buff from the callback. On initialization, add a buff to every character besides the holder with duration -1. Inside that buff, check if the holder has the `athame-self-atk` (make this a constant) and give the team buff if the holder has the self atk% buff. Also rename the `athame-self-atk` buff to just `athame-buff` or use the name given by the weapon `blade-of-the-daylight-hours`

Charlie-Zheng (changes requested):

Charlie-Zheng:
Also for the team buff, it does stack with the self buff when there are multiple holders.

For the team buff, it always uses the buff from the higher refine. So R1 Athame Artis and R5 Athame Artis holders will always use the R5's buff value for team buff, where applicable

internal/weapons/sword/athameartis/athameartis.go
Charlie-Zheng: Upon further research, how the on fielder buff works is that when the self buff is applied, it starts a task that grants the on fielder the ATK% buff for 2s/3s (Hex/non Hex) every 1s.

Nahbrowhy:
I've addressed the review comments and pushed the requested changes. The only remaining part is the team buff implementation, which I'm still verifying based on the latest behavior discussion and will update in a follow-up commit.

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+			return burstCDMod
+		},
+	})
+
+	selfVal := make([]float64, attributes.EndStatType)
+	selfVal[attributes.ATKP] = selfATK
+
+	selfValHex := make([]float64, attributes.EndStatType)
+	selfValHex[attributes.ATKP] = selfATK * 1.75
```
Charlie-Zheng: You don't need two of these. You can either set the value in Init(), since at that point all the characters have been initialized, or you can set the value in the StatMod

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,121 @@
...
+		Amount: func() []float64 {
+			if !char.StatusIsActive(bladeOfDaylightHours) {
+				return nil
+			}
+
+			if c.Player.GetHexereiCount() >= 2 {
+				return selfValHex
+			}
+
+			return selfVal
```
Charlie-Zheng: As mentioned earlier, you can do this
```go
if c.Player.GetHexereiCount() >= 2 {
    selfVal[attributes.ATKP] = selfATK * 1.75
} else {
    selfVal[attributes.ATKP] = selfATK
}
```

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,121 @@
...
+
+		active.AddStatMod(character.StatMod{
+			Base:         modifier.NewBaseWithHitlag(teamBuffKey, duration),
+			AffectedStat: attributes.ATKP,
+			Amount: func() []float64 {
+				return buff
+			},
+		})
+	}
+	c.Events.Subscribe(event.OnEnemyDamage, func(args ...any) {
```
Charlie-Zheng: Change this to OnEnemyHit, add a todo:
```
// TODO: ATK% buff should affect hit that triggered it. Currently gcsim snapshots before OnEnemyHit event so it isn't buffing the triggering hit.

File: internal/weapons/sword/athameartis/athameartis.go [hunk 12]
```diff
@@ -0,0 +1,121 @@
...
+
+			if c.Player.GetHexereiCount() >= 2 {
+				return selfValHex
+			}
+
+			return selfVal
+		},
+	})
+
+	applyTeamBuff := func() {
```
Charlie-Zheng: This needs to reapply every 1s to the current active char. I see your comment that you still need to add it, but I'm leaving this here as a reminder

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,121 @@
...
+		active := c.Player.ActiveChar()
+
+		duration := 3 * 60
+		buff := teamVal
+
+		if c.Player.GetHexereiCount() >= 2 {
+			duration = 2 * 60
+			buff = teamValHex
```
Charlie-Zheng: Same as the self buff, you can
```go
if c.Player.GetHexereiCount() >= 2 {
    teamVal[attributes.ATKP] = teamATK * 1.75
} else {
    teamVal[attributes.ATKP] = teamATK
}
```

internal/weapons/sword/athameartis/athameartis.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		active.AddStatMod(character.StatMod{
```
Charlie-Zheng: Add a TODO:

```
// TODO: When there are Athame Artis of different refines on the team, the team atk% buff should prioritize higher buff value

Charlie-Zheng (changes requested):

internal/weapons/sword/athameartis/athameartis.go
Charlie-Zheng: I think I know how it works but it's a bit complicated to implement, I will push an update to your branch when I figure it out. For now you don't need to worry about different refines on the team.

Charlie-Zheng (dismissed):

Charlie-Zheng (approved):

PR #2637, opened 2026-06-13
Add Nicole

Charlie-Zheng:
- [x] New character package
- [x] Character key
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation


ErrorRaffyline0:
plsss xianyun plunge

internal/characters/nicole/cons.go
```diff
@@ -0,0 +1,237 @@
...
+		return
+	}
+
+	c.Core.Events.Subscribe(event.OnEnemyHit, func(args ...any) {
+		atk := args[1].(*info.AttackEvent)
+		char := c.Core.Player.ByIndex(atk.Info.ActorIndex)
+		if !char.StatModIsActive(a1Key) {
+			return
+		}
+		atk.Info.IgnoreDefPercent = min(atk.Info.IgnoreDefPercent+0.4, 0.9)
```
jwalk-511: Should be min of 100%, if not higher
https://x.com/dhcwsp/status/2056181715287027761
https://discord.com/channels/763583452762734592/1422622248380141698/1517594720694698084

Though this was done on the creator experience server, so ideally someone can show a replicated version from prod

internal/characters/nicole/asc.go
```diff
@@ -0,0 +1,124 @@
...
+	char.AddStatMod(character.StatMod{
+		Base:         modifier.NewBase(a1Key, dur),
+		AffectedStat: attributes.ATK,
+		Amount: func() []float64 {
+			return c.a1Buff
+		},
+	})
+	c.c2OnUpgrade(char)
+	c.c4OnUpgrade(char)
+	c.c6OnUpgrade(char)
```
jwalk-511, line 108: is this supposed to be called for any character's upgrade or just when nicole's self-buff upgrades?
Wording on the constellation is `When Nicole's Grace of Kenosis is upgraded to Guidance of Theosis, all nearby [party](https://genshin-impact.fandom.com/wiki/Party) members' Grace of Kenosis will also be upgraded to Guidance of Theosis.`

e.g. if nicole uses skill, swaps to another char, and waits 3 seconds for the other char to activate without ever having elemental damage to upgrade herself, does c6 still proc?

Charlie-Zheng: If you look inside the c6OnUpgrade code it checks if it was nicole that upgraded

internal/characters/nicole/cons.go
Charlie-Zheng: I think I will remove it from here. It makes more sense for the max to set by the damage calculation section

jwalk-511 (approved):

PR #2638, opened 2026-06-15
Removed nightsoul consume event from nightsoul clear points

Matoba004:
Removed s.c.Events.Emit(event.OnNightsoulConsume from ClearPoints() in nightsoul.go.

Abilities that consumed all nightsoul points were changed from ClearPoints() to ConsumePoints(c.nightsoulState.Points()).

Purpose of this PR is to stop Mavuika from getting fighting spirit when characters cancel their nightsoul state and don't spend the points.

### DIFF
--- a/internal/characters/citlali/skill.go
+++ b/internal/characters/citlali/skill.go
@@ -78,7 +78,7 @@ func (c *char) Skill(_ map[string]int) (action.Info, error) {
 
 	if c.Base.Cons >= 6 {
 		currentPoints := c.nightsoulState.Points()
-		c.nightsoulState.ClearPoints()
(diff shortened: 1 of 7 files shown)


Charlie-Zheng:
Did you run dbsnapshot to compare main vs this branch?

Charlie-Zheng:
This sim seems to gain like 500 ish DPS, is that expected? https://gcsim.app/db/b6jGj7rHPdKr
This one loses some 500 ish DPS, is that expected? https://gcsim.app/db/MQc9TwDcB7Dq
This one gains like 5k or something DPS https://gcsim.app/db/MjWHC78RkWCF

PR #2639, opened 2026-06-15
Add Ifa

Matoba004:
Character PR checklist:

- [x] New character package
- [x] Character key
- [ ] Update `mode_gcsim.js` with shortcuts for syntax highlighting
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [ ] Update documentation


internal/characters/ifa/config.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 0 # 1-Hit DMG|{param0:F1P}
+    attack_2:
+      - 1 # 2-Hit DMG|{param1:F1P}
+    attack_3:
+      - 2 # 3-Hit DMG|{param2:F1P}
+    charge:
+      - 3 # Charged Attack DMG|{param3:F1P}
+    plunge_collision:
+      - 5 # Plunge DMG|{param5:F1P}
+    plunge_low:
```
Charlie-Zheng: use snakeCase except for the _1 or _2 at the end of attack

internal/characters/ifa/skill.go
```diff
@@ -0,0 +1,202 @@
...
+	if c.skillParticleICD {
+		return
+	}
+	c.skillParticleICD = true
+
+	c.Core.QueueParticle(c.Base.Key.String(), 4, attributes.Anemo, c.ParticleDelay)
+
+	if c.Core.Rand.Float64() < 0.3 {
+		c.Core.QueueParticle(c.Base.Key.String(), 1, attributes.Anemo, c.ParticleDelay)
+	}
```
Charlie-Zheng: Generally we just do `particles := 4.0` and then if the condition is true, `particles = 5.0`, then add all the particles in one call

internal/characters/ifa/skill.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Checks the current number of nightsoul points and exits nightsoul if there aren't enough. Returns the status of NS after the check
+func (c *char) checkNS() {
+	if c.nightsoulState.Points() < 0.001 {
+		c.exitNightsoul()
+	}
+}
+
+func (c *char) enterNightsoul() {
```
Charlie-Zheng, line 65: add `c.skillParticleICD = false`

internal/characters/ifa/ifa.go
```diff
@@ -0,0 +1,99 @@
...
+	c.Core.Events.Subscribe(event.OnNightsoulAdd, func(args ...any) {
+		amount := args[1].(float64)
+		if amount < 0.0000001 {
+			return
+		}
+		idx := args[0].(int)
+
+		c.teamNightsoulPoints[idx] += amount
+	}, "ifa-add-team-nightsoul")
+}
```
Charlie-Zheng: This is not needed. You can do `char.Condition([]string{"nightsoul", "points"})` to directly obtain the number of points a nightsoul character has (given they implement the Condition interface like they are supposed to).

Charlie-Zheng:
```go
func (c *char) getTeamNightsoul() float64 {
	sum := 0.0
	for _, char := range c.Core.Player.Chars() {
		points, err := char.Condition([]string{"nightsoul", "points"})
		if err != nil {
			continue
		}
		p, ok := points.(float64)
		if !ok {
			continue
		}
		sum += p
	}
	return sum
}
```

I suspect this probably only updates every X frames but you can also just call this every time you need it

internal/characters/ifa/burst.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, e := range enemies {
+		t, ok := e.(info.TargetWithAura)
+
+		if !ok {
+			continue
+		}
+
+		ele := c.targetElement(t)
+
+		if collision, _ := t.AttackWillLand(p); collision && ele != attributes.NoElement {
```
Charlie-Zheng: What is this `collision, _ := t.AttackWillLand(p); collision` check for?

Might be late but you can reference Heizou burst for implementation. Ifa burst seems to be almost an exact copy

internal/characters/ifa/charge.go
```diff
@@ -0,0 +1,130 @@
...
+		travel = 0
+	}
+
+	pos := c.Core.Combat.PrimaryTarget().Pos()
+	c.QueueCharTask(func() {
+		c.Core.QueueAttack(ai, combat.NewCircleHitOnTarget(
+			pos,
+			nil,
+			chargeRadius,
+		),
```
Charlie-Zheng, line 65: Is this what the `go fmt` gives for the format here? If so, I'd recommend moving the attack pattern onto it's own line.

internal/characters/ifa/skill.go
```diff
@@ -0,0 +1,202 @@
...
+	return action.Info{
+		Frames:          frames.NewAbilFunc(skillCancelPlungeFrames),
+		AnimationLength: skillCancelPlungeFrames[action.InvalidAction],
+		CanQueueAfter:   skillCancelPlungeFrames[action.ActionDash], // earliest cancel
+		State:           action.PlungeAttackState,
+	}, nil
+}
+
+func (c *char) healCB(a info.AttackCB) {
+	em := c.Stat(attributes.EM)
```
Charlie-Zheng, line 194: Is this supposed for heal once for every target hit?

internal/characters/ifa/asc.go
```diff
@@ -0,0 +1,73 @@
...
+					return 0
+				}
+
+				m := 0.0
+
+				for _, nightsoulAmount := range c.teamNightsoulPoints {
+					m += nightsoulAmount
+				}
+
+				m = 60 + max((m-60), 0)*c.c2Mult()
```
Charlie-Zheng: this looks kind of weird here? Just pass m into c.c2Mult() and calculate the total bonus points there.
```
m += c.c2BonusPoints(m)
```

internal/characters/ifa/asc.go
```diff
@@ -0,0 +1,73 @@
...
+	if c.Base.Ascension < 4 {
+		return
+	}
+
+	m := make([]float64, attributes.EndStatType)
+	m[attributes.EM] = 80
+
+	c.Core.Events.Subscribe(event.OnNightsoulBurst, func(args ...any) {
+		c.AddStatMod(character.StatMod{
+			Base:         modifier.NewBase("ifa-a4", 10*60),
```
Charlie-Zheng: Should be NewBaseWithHitlag

internal/characters/ifa/cons.go
```diff
@@ -0,0 +1,104 @@
...
+func (c *char) c4OnBurst() {
+	if c.Base.Cons < 4 {
+		return
+	}
+
+	m := make([]float64, attributes.EndStatType)
+	m[attributes.EM] = 100
+
+	c.AddStatMod(character.StatMod{
+		Base:         modifier.NewBase("ifa-c2", 15*60),
```
Charlie-Zheng: Should be NewBaseWithHitlag. The key should be `ifa-c4`

internal/characters/ifa/plunge.go
```diff
@@ -0,0 +1,89 @@
...
+	lowPlungeFrames[action.ActionAttack] = 50
+	lowPlungeFrames[action.ActionCharge] = 50
+	lowPlungeFrames[action.ActionSkill] = lowPlungeHitmark // assuming it's the same as burst
+	lowPlungeFrames[action.ActionBurst] = 50
+	lowPlungeFrames[action.ActionDash] = 61 - 19
+	lowPlungeFrames[action.ActionWalk] = 69
+	lowPlungeFrames[action.ActionSwap] = 57
+}
+
+func (c *char) LowPlungeAttack(p map[string]int) (action.Info, error) {
```
Charlie-Zheng, line 32: You checked that it's always low plunge and not high plunge unless they jump (and we don't support jump so we ignore that)?

Charlie-Zheng (changes requested):

internal/characters/ifa/plunge.go
Matoba004: Yes, when you don't jump before plunging you're certain to get a low plunge.

Matoba004: (except for when enemy hitbox pushes you up of course)

Matoba004:
I found a couple more issues with how I implemented Ifa skill bullets and frames in skill. I'll get to fixing them on Monday.

internal/characters/ifa/data_gen.textproto
```diff
@@ -0,0 +1,151 @@
+id: 10000113
```
shizukayuki: delete, not in use anymore.

ui/packages/localization/src/locales/names.generated.json
```diff
@@ -0,0 +1,5749 @@
+{
```
shizukayuki: delete, not in use anymore.

internal/characters/ifa/config.yml
```diff
@@ -0,0 +1,34 @@
...
+  - plunge_collision: attack(Plunge DMG)
+  - plunge_low: attack(Low/High Plunge DMG) # {}/{}
+  - plunge_high: attack(Low/High Plunge DMG|1) # {}/{}
+
+  - skill_dmg: skill(Tonicshot DMG)
+  - skill_heal: skill(Tonicshot Healing On Hit) # {} Elemental Mastery+{}
+  - skill_heal_flat: skill(Tonicshot Healing On Hit|1) # {} Elemental Mastery+{}
+
+  - burst_dmg: burst(Skill DMG)
+  - burst_mark: burst(Sedation Mark DMG)
```
Charlie-Zheng: These should be snakeCase

internal/characters/ifa/burst.go
```diff
@@ -0,0 +1,127 @@
...
+
+			return burstFramesGrounded[next]
+		},
+		AnimationLength: burstFramesGrounded[action.InvalidAction],
+		CanQueueAfter:   burstFramesNS[action.ActionDash], // earliest cancel
+		State:           action.BurstState,
+	}, nil
+}
+
+func (c *char) sedationMarkCB(a info.AttackCB) {
```
Charlie-Zheng: Currently this will do the check after aura gets consumed by the anemo swirl (I think?). Is this correct to how his burst works in game?

internal/characters/ifa/asc.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, char := range c.Core.Player.Chars() {
+		char.AddReactBonusMod(character.ReactBonusMod{
+			Base: modifier.NewBase("ifa-a1", -1),
+			Amount: func(ai info.AttackInfo) float64 {
+				if !c.nightsoulState.HasBlessing() {
+					return 0
+				}
+
+				m := c.getTeamNightsoul()
```
Charlie-Zheng: nit: use `ns` for the var name or `stacks` instead of `m`. We usually use `m` for the stat arrays

Charlie-Zheng (changes requested):

internal/characters/ifa/burst.go
Matoba004: in dm the check is done by onHittingOther which i presume happens before the reaction

Matoba004:
```
c.Core.Tasks.Add(func() {
	for _, t := range c.Core.Combat.EnemiesWithinArea(ap, nil) {
		c.sedationMark(t)
	}
}, burstHitmark)

c.Core.QueueAttack(
	ai,
	ap,
	burstHitmark,
	burstHitmark,
)
```
is it fine to resolve it like this? do tasks ever get swapped around in task queue?

Charlie-Zheng: It should be guaranteed that the tasks execute in order (assuming they are in the same queue).

Charlie-Zheng (approved):

PR #2649, opened 2026-06-22
Fixed mona c6 not consuming stacks

Matoba004:


Charlie-Zheng:
Please update to latest main. I think this mona hex burst change is already merged.

Matoba004:
will do

Charlie-Zheng (approved):

PR #2652, opened 2026-06-23
Lunar Crystallize Fixes 2.0

Charlie-Zheng:
Fixes LCr not triggering Scroll 4pc
Fixes LCr spawning moondrift not counting towards the hit number or contributor
Fixes LCr not enabling geo resonance

PR #2661, opened 2026-06-25
Updated geo resonance behavior to match 6.3 changes from datamine

Matoba004:
- Added 1s ticker hitlag extended by the active character
- Added onSwap event when the ticker should be forced to tick
- Updated moondrift distance

Currently conflicts with #2652 which adds moondrift checking but not the ticker

### DIFF
--- a/pkg/simulation/setup.go
+++ b/pkg/simulation/setup.go
@@ -254,6 +254,18 @@ func SetupResonance(s *core.Core) {
 				updateGeoResonance(s, geoResTickerSrc, &geoResTickerSrc)()
 			}, "geo-res-on-swap")
 
+			s.Events.Subscribe(event.OnShielded, func(args ...any) {
+				geoResTickerSrc = s.F
+
+				updateGeoResonance(s, geoResTickerSrc, &geoResTickerSrc)
+			}, "geo-res-on-shielded")
+
+			s.Events.Subscribe(event.OnShieldBreak, func(args ...any) {
+				geoResTickerSrc = s.F
+
+				updateGeoResonance(s, geoResTickerSrc, &geoResTickerSrc)
+			}, "geo-res-on-shield-break")
+
 		case attributes.Anemo:
 			s.Player.AddStamPercentMod("anemo-res-stam", -1, func(a action.Action) (float64, bool) {
 				return -0.15, false
--- a/pkg/simulation/setup.go
+++ b/pkg/simulation/setup.go
@@ -20,8 +20,6 @@ import (
 	"github.com/genshinsim/gcsim/pkg/modifier"
 )
 
-var geoResTickerSrc int
-
 const (
 	nightsoulBurstICDStatus = "nightsoul-burst-icd"
 	geoResKey               = "geo-res"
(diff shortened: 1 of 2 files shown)


pkg/simulation/setup.go
```diff
@@ -19,7 +20,12 @@ import (
 	"github.com/genshinsim/gcsim/pkg/modifier"
 )
 
-const nightsoulBurstICDStatus = "nightsoul-burst-icd"
+var geoResTickerSrc int
```
Charlie-Zheng: This var cannot be global. This would make it shared by every iteration. You can use a Tag/Flag for this

Charlie-Zheng (changes requested):

pkg/simulation/setup.go
```diff
@@ -308,6 +328,36 @@ func SetupResonance(s *core.Core) {
...
+			if playerPos.Distance(moondrift.Pos()) < 16 {
+				isGeoResActive = true
+			}
+		}
+
+		if isGeoResActive {
+			s.Flags.Custom[geoResKey] = 1
+		} else {
+			s.Flags.Custom[geoResKey] = 0
+		}
```
Charlie-Zheng, line 362: Can this be done as a status on Core that you add or delete instead (with duration -1)

Charlie-Zheng (changes requested):

pkg/simulation/setup.go
Matoba004: probably, is this how you check for if status is active in core? s.Status.Duration(geoResKey) != 0

pkg/simulation/setup.go
Matoba004: do i then pass it as a pointer in the updateGeoResonance?

pkg/simulation/setup.go
Charlie-Zheng: Looking at the code for Core.Status, it doesn't support infinite duration statuses, so we do need to use the Flags instead of a status.

Charlie-Zheng:
From dbcompare:
[pr2661.txt](https://github.com/user-attachments/files/29362708/pr2661.txt)

Most diffs look small enough. 


However, there are some navia sims which are consistently losing damage, I assume because the delay from absorbing the Crystallize Shard and when the ticker picks it up is causing Navia's hold E to not be buffed by the 15%
https://gcsim.app/db/8jLwzcqkmzGw

Can you get some verification that the buff doesn't apply immediately on shield?

Matoba004:
There are still some unexplained differences.

https://gcsim.app/db/JFRCCFzkBHJC

Matoba004:
Should work as intended now.

Charlie-Zheng:
Reran dbcompare, only sim that has changes is https://gcsim.app/db/wNHmwLJH6RKJ losing 1k DPS. Not sure why this would lose DPS compared to main? (Note need to rebase to get latest main changes)

Matoba004:
The change is from geo resonance with lunarcrystallize totems not working from off field which is intended.

Matoba004:
Relevant footage:

https://www.youtube.com/watch?v=PGgkJFoMmVg

The damage loss from 1.7k to 1.2k is caused by c2 40% hp buff expiring.

Charlie-Zheng (approved):

PR #2663, opened 2026-06-26
implement .tags for targets

imring:


pkg/conditional/enemy.go
```diff
@@ -85,6 +85,21 @@ func evalElement(c *core.Core, fields []string) (float64, error) {
 	return float64(dur), nil
 }
 
+func evalTags(c *core.Core, fields []string) (int, error) {
+	// .tags.e1.name
```
Charlie-Zheng: this should be .tags.t1.name

PR #2673, opened 2026-07-07
Fix Columbina A4

Charlie-Zheng:


pkg/reactable/lunarcharged.go
```diff
@@ -234,8 +229,12 @@ func (r *Reactable) nextLCTick(src int) func() {
 				continue
 			}
 
-			e.AddStatus(lcIcdKey, icd, true)
+			e.AddStatus(lcIcdKey, 2*60, true)
+
 			enemy.DoLCAttack()
+			if chance, ok := r.core.Flags.Custom[LcrExtraHitOverride]; ok && r.core.Rand.Float64() < chance {
```
jwalk-511, line 235: `LcrExtraHitOverride` should be `LcExtraHitOverrideKey`

I also think the LC custom flag should drop the `key` suffix to be consistent with LCr.

PR #2678, opened 2026-07-16
Fix dds-4pc event condition

yihuajack:
Disenchantment in the Deep Shadow 4pc event should be OnLunarReactionAttack rather than OnLunarChargedReactionAttack


Matoba004 (commented):
This serves to add crit stats to lunar contributors with this set. See Blizzard strayer for another example. Otherwise the crit rate adding is handled through attackMod.

PR #2682, opened 2026-07-21
feat(optimization): replace greedy descent with ascent from zero

bytecode-larper:
# Substat Optimizer: Greedy Ascent from Zero

## Problem

The current substat optimizer uses a **greedy descent** strategy:

1. Fill every relevant stat to its per-artifact maximum
2. Compute DPS gradients for removing a chunk (−20, −5, −2, −1) from each stat
3. Remove from the stat that loses the least DPS
4. Repeat until the total substat budget is met

Because it starts from maximum values, early large‑chunk removals (−20) skip over the marginal value of the last few subs in a stat before the next gradient batch is computed.

## Solution

Replace the descent with a **greedy ascent from zero**:

1. Start every relevant stat at **0** allocated substats
2. At each step add **+1** to the stat with the highest marginal DPS gain
3. Repeat until the substat budget is exhausted

Starting from zero avoids the plateau problem: once a stat reaches its useful cap (e.g. 100% CR) the next +1 gradient is ~0 and the algorithm naturally switches to under‑allocated stats. The +1 step size means no allocations are skipped.

### Changes vs the remote (main) algorithm

| Aspect | Remote (main) | This branch |
|---|---|---|
| **Strategy** | Descent from max | Ascent from zero |
| **Step size** | −20/−5/−2/−1 chunks (jumps over optimal) | +1 single-step (no skipping) |
| **Inner loop** | `for totalSubs >= 25` loops re-evaluating the same stale gradients | Removed — one decision per iteration with fresh gradient |
| **Gradient selection** | `allocateSomeSubstatGradientsForChar` sorted by gradient but only removes one stat | Direct argmax over gradients |
| **Seeds** | `time.Now().UnixNano()` per stat batch — different noise per stat | `gradientSeed` counter — same noise across all stats in a batch (more stable comparisons) |
| **Lazy recompute** | None — full batch recompute every loop iteration | Single-stat recompute for the stat that changed (reduces sim calls by ~50%) |
| **Flat stats** | HP, ATK, DEF always included | Included in initial set; gradient-threshold filter drops them when not useful |
| **Irrelevant %-stats** | All 9 stats evaluated every iteration | Gradient-threshold filter drops sub-1% stats after initial batch (auto-adapts per character) |

Additional infrastructure changes:
- **Deterministic seeds** (`gradientSeed`) in `details.go` + `substats.go`
- **`calculateSingleSubstatGradient`** helper in `details.go` for lazy recompute
- **Refactored descent** in `opt_damage.go`: removed inner while-loop, simplified to single-stat removal

## Benchmark Results

Configs pulled from the public DB at `simpact.app/api/db`, covering melt, freeze, vape, aggravate, and hypercarry teams. All times are wall‑clock on a Ryzen 5 5600X.

Gradient evaluations use **25 iterations per sim call** (unchanged from main). Each result is the mean of the 25 seeds.

### Light budget (20 substats, 10 per stat)

| Config | Team | Descent | Ascent | Δ |
|---|---|---|---|---|
| Skirk | Furina / Yelan / Escoffier | 136,558 | **138,994** | **+1.75%** |
| Mualani | Mona / Sucrose / Mavuika | 148,960 | **148,983** | **+0.02%** |
| Mavuika | Citlali / Sucrose / Nicole | 117,970 | **120,244** | **+1.89%** |
| Raiden | Xiangling / Kazuha / Bennett | 48,904 | **48,935** | **+0.06%** |
| Kokomi | Ayaka / Kazuha / Rosaria | 59,080 | **61,308** | **+3.63%** |
| Varesa | Fischl / Nahida / Xilonen | 84,467 | **89,832** | **+5.97%** |
| Kinich | Emilie / Xiangling / Bennett | 80,744 | **88,922** | **+9.20%** |
| Flins | Ineffa / Furina / Sucrose | 105,256 | **116,657** | **+9.77%** |
| Chasca | Furina / Xiangling / Bennett | 70,558 | **72,595** | **+2.81%** |
| Arlecchino | Furina / Yelan / Xianyun | 78,389 | **81,553** | **+3.88%** |

**Ascent wins 10 / 10 comparisons.** Weighted average gain: **+3.84%**.

### Heavy budget (40 substats, 15 per stat)

| Config | Team | Descent | Ascent | Δ |
|---|---|---|---|---|
| Skirk | Furina / Yelan / Escoffier | 194,361 | **197,422** | **+1.55%** |
| Mualani | Mona / Sucrose / Mavuika | 201,682 | **202,458** | **+0.38%** |
| Mavuika | Citlali / Sucrose / Nicole | 154,640 | **164,783** | **+6.16%** |
| Raiden | Xiangling / Kazuha / Bennett | **66,818** | 66,411 | −0.61% |
| Kokomi | Ayaka / Kazuha / Rosaria | 79,140 | **80,240** | **+1.37%** |
| Varesa | Fischl / Nahida / Xilonen | 120,127 | **124,975** | **+3.88%** |
| Kinich | Emilie / Xiangling / Bennett | 111,041 | **118,139** | **+6.01%** |
| Flins | Ineffa / Furina / Sucrose | 143,178 | **158,568** | **+9.71%** |
| Chasca | Furina / Xiangling / Bennett | 95,081 | **98,295** | **+3.27%** |
| Arlecchino | Furina / Yelan / Xianyun | 115,463 | **117,933** | **+2.09%** |

**Ascent wins 9 / 10 comparisons** (Raiden is the sole loss, within noise at −0.61%). Weighted average gain: **+3.59%**.

### Speed

| Budget | Descent | Ascent |
|---|---|---|
| Light (20/10) | 7.6s | 10.6s |
| Heavy (40/15) | 7.6s | 15.3s |

The ascent is **1.4–2× slower** than the descent because each of the 20–40 budget steps requires a separate sim evaluation (lazy recompute of the changed stat only). Worst case (heavy budget) is 15.3s per config — still within interactive range for a CLI tool.

### Favonius CR Allocation

CR substat allocation for Fav holders (8 characters across 7 configs) is identical between both algorithms at both budgets. The ascent's DPS advantage comes from better allocation of non-CR substats, not from starving Fav holders.

## Testing

Benchmarked against 10 configs pulled from the public DB, test / benchmark suite not part of this PR.

### DIFF
--- a/pkg/optimization/details.go
+++ b/pkg/optimization/details.go
@@ -2,8 +2,6 @@ package optimization
 
 import (
 	"context"
-	"sort"
-	"time"
 
 	"github.com/genshinsim/gcsim/pkg/core/attributes"
 	"github.com/genshinsim/gcsim/pkg/core/info"
@@ -41,43 +39,11 @@ type SubstatOptimizerDetails struct {
 	indivSubstatLiquidCap   int
 	totalLiquidSubstats     int
 	optimizer               *SubstatOptimizer
+	gradientSeed            int64
 }
 
-func (stats *SubstatOptimizerDetails) allocateSomeSubstatGradientsForChar(
-	idxChar int,
-	_ info.CharacterProfile,
-	substatGradient []float64,
-	relevantSubstats []attributes.Stat,
-	amount int,
-) []string {
-	var opDebug []string
-	sorted := newSlice(substatGradient...)
-	sort.Sort(sort.Reverse(sorted))
-
-	for _, idxSubstat := range sorted.idx {
-		substat := relevantSubstats[idxSubstat]
-
-		if amount > 0 {
-			if stats.charSubstatFinal[idxChar][substat] < stats.charSubstatLimits[idxChar][substat] {
-				stats.charSubstatFinal[idxChar][substat] += amount
-				stats.charProfilesCopy[idxChar].Stats[substat] += float64(amount) * stats.substatValues[substat] * stats.charSubstatRarityMod[idxChar]
-				return opDebug
-			}
-		}
-
-		if stats.charSubstatFinal[idxChar][substat] > 0 {
-			amount = clamp[int](-stats.charSubstatFinal[idxChar][substat], amount, amount)
-			stats.charSubstatFinal[idxChar][substat] += amount
-			stats.charProfilesCopy[idxChar].Stats[substat] += float64(amount) * stats.substatValues[substat] * stats.charSubstatRarityMod[idxChar]
-			return opDebug
-		}
-	}
-
-	// TODO: No relevant substat can be allocated/deallocated, alloc/dealloc some random other substat??
-	opDebug = append(opDebug, "Couldn't alloc/dealloc anything?????")
-	return opDebug
-}
-
+// #1: deterministic seed — uses stats.gradientSeed which increments each call,
+// but is constant within a batch so all N stats share the same noise.
 func (stats *SubstatOptimizerDetails) calculateSubstatGradientsForChar(
 	idxChar int,
 	relevantSubstats []attributes.Stat,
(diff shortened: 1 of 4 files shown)


Charlie-Zheng:
Some characters do actually use flat stats, such as kokomi when she isn't using burst scales only ATK and Flat Atk. Gradient ascent without giving CR can cause some sims to issues due to Favonius conditions. Also processing characters in reverse order does nothing because character order is completely up to the config writer.

Also please split the PRs up. Isolate the PR to only the gradient ascent here. If you want to have performance comparisons between ascent vs descent, don't change the number of iterations per call.

bytecode-larper:
Thanks for the review, @Charlie-Zheng.

**Flat stats**: Fair point — I've added HP, ATK, DEF back to the initial relevant substats set. The gradient-threshold filter drops them when they're not useful anyway, so the overhead is negligible and correctness is preserved for edge cases like Kokomi.

**Fav CR**: I'll test it and see if there's a regression.

**Reverse order**: I kept it because there's no downside, doing them from first to last was arbitrary to begin with, and this covers cases like support buffs being settled for the main DPS, unless I'm missing something.

**Splitting the PR**: Will do.

bytecode-larper:
I also tested the same chunking of substats I think it was still better than original algo but worse than going 1 by 1 and the speed isn't that much worse so that's what is contained here

bytecode-larper:
@Charlie-Zheng any issues?

bytecode-larper:
mb gang

soloxcx:
I checked one of your results, the Flins+Ineffa+Furina+Sucrose improvement from 105k to 116k, and your PR is currently giving Flins/Ineffa/Sucrose 30 liquid substats instead of 20.

current https://gcsim.app/db/RPPt69LRBMwM#
after PR https://gcsim.app/sh/KwGkhPw6mhkr

in general your results showing the descent optimizer being off by 10% or even 5% would strike me as unusual so I would double check all those results.

bytecode-larper:
liquid bug was present in benchmarks; after fix the interesting gap is big-budget CR handling; want to try a few approaches before changing the default following discord feedback; closing for now;

PR #2689, opened 2026-07-22
Change scroll set to use pipeline data

shizukayuki:


Charlie-Zheng:
This doesn't seem to actually be any easier than typing the numbers by hand?

shizukayuki:
i agree it is not easier but fixing that require using binout for names.
weapons might be a better fit for this.

PR #2706, opened 2026-07-31
add missing character import baseline

Basilaa:
## What changed

- add baseline character packages for 13 missing community characters
- register community character keys and simulation imports
- add a manifest plus discovery, generation, and validation scripts
- add shared basic-import scaffolding and a tracked follow-up list

## Why

The repository did not yet have importable package skeletons for these newer characters. This establishes a compile-safe baseline so frame data, multipliers, mechanics, constellations, and reactions can be filled incrementally.

## Impact

The new packages compile and are discoverable by the simulator, but they are intentionally incomplete and should not yet be treated as finished character implementations.

## Validation

- `conda run -n genshin go test ./...` — passed
- `conda run -n genshin python scripts/validate_missing_characters.py` — reports remaining Linnea work: four local config labels are absent and the generated multiplier array is not 15 levels

## Follow-up

- resolve the Linnea config-label and multiplier-array validation findings
- fill authoritative frame, hit, energy, reaction, weapon, artifact, and character-mechanic data


Charlie-Zheng:
We don't accept work with pre-release stuff

PR #2712, opened 2026-08-05
Add Prune

Khoi0612:
Character PR checklist:

- [x] New character package
- [x] Run pipeline with added config (generates character curve, talent stats, `.dm.json` files)
- [x] Character key
- [x] Normal Attack
- [x] Skill
- [x] Burst
- [x] A1
- [x] A4
- [x] C1
- [x] C2
- [x] C3
- [x] C4
- [x] C5
- [x] C6
- [x] Hitlag
- [x] ICD
- [x] StrikeType
- [x] Hitboxes
- [x] Attack durability
- [x] Particles
- [x] Frames
- [x] Update documentation


internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+			c.AddStatus(a1SwirlCheckICDKey, 72, false)
+
+			// queue passive
+			ai := info.AttackInfo{
+				ActorIndex: c.Index(),
+				Abil:       "Verdict and Punishment (A1)",
+				AttackTag:  attacks.AttackTagElementalBurst,
+				ICDTag:     attacks.ICDTagNone,
+				ICDGroup:   attacks.ICDGroupDefault,
+				StrikeType: attacks.StrikeTypeBlunt,
```
jwalk-511, line 63: blunt damage should include the poise done (20 from wiki, but double-check dm)

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+				0,
+				45,
+				c.makeA4CB,
+				c.makeC1CB,
+				c.c4Ricochet(ele, attacks.AttackTagElementalBurst),
+				c.makeC2CB,
+			)
+
+			// swirl burst has different energy drain delay frame
+			c.burstEnergyDrainDelay = 19
```
jwalk-511: this won't be set until after a swirl occurs (subscribed to onswirl), while https://github.com/genshinsim/gcsim/pull/2712/changes#diff-4d383d21e1c53347bde053d47ef6225470eda1ab779ef6027dc56ff225d1f2dfR74 is called as soon as the burst ability is used.

But checking the viability of having a variable burst energy drain delay at all-
A swirl burst would hit at earliest 34 frames after burst is called, so how can the burstEnergyDrainDelay be modified based on an event that happens after it should already be finished? https://github.com/genshinsim/gcsim/pull/2712/changes#diff-4d383d21e1c53347bde053d47ef6225470eda1ab779ef6027dc56ff225d1f2dfR65

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+
+	c.Core.Events.Subscribe(event.OnSwirlCryo, swirlfunc(attributes.Cryo), "prune-burst-cryo")
+	c.Core.Events.Subscribe(event.OnSwirlElectro, swirlfunc(attributes.Electro), "prune-burst-electro")
+	c.Core.Events.Subscribe(event.OnSwirlHydro, swirlfunc(attributes.Hydro), "prune-burst-hydro")
+	c.Core.Events.Subscribe(event.OnSwirlPyro, swirlfunc(attributes.Pyro), "prune-burst-pyro")
```
jwalk-511: remember to come back after stellar swirl is added and add a subscription for it

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (c *char) makeA4CB(a info.AttackCB) {
+	if c.Base.Ascension < 4 || a.Target.Type() != info.TargettableEnemy {
+		return
+	}
+	c.a4Buff = make([]float64, attributes.EndStatType)
+	atk := c.SelectStat(true, attributes.BaseATK, attributes.ATKP, attributes.ATK).TotalATK()
```
jwalk-511, line 105: requesting a double-check that she actually snapshots attack and isn't fully dynamic.
Most references I can find on discord say she is dynamic, but I'll trust your reading of the dm over that if confirmed

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+	c.a4Buff[attributes.DmgP] = math.Min(math.Max(atk-2000, 0)*0.00025, 0.5)
+
+	for i, char := range c.Core.Player.Chars() {
+		if i == c.Index() {
+			continue // nothing for prune, need testing
+		}
+		char.AddStatus(a4Key, 5*60, true)
+		char.AddAttackMod(character.AttackMod{
+			Base: modifier.NewBaseWithHitlag("prune-a4", 5*60), // 5 s
```
jwalk-511: you can name this base `a4Key` instead of `"prune-a4"` since a status is just a base with no other use. Then you can remove the extra addStatus at line 103

internal/characters/prune/attack.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (c *char) Attack(p map[string]int) (action.Info, error) {
+	ai := info.AttackInfo{
+		ActorIndex:         c.Index(),
+		Abil:               fmt.Sprintf("Normal %v", c.NormalCounter),
+		AttackTag:          attacks.AttackTagNormal,
+		ICDTag:             attacks.ICDTagNormalAttack,
+		ICDGroup:           attacks.ICDGroupDefault,
+		StrikeType:         attacks.StrikeTypeDefault,
```
jwalk-511: wiki claims prune attacks are blunt
https://genshin-impact.fandom.com/wiki/Badaboom!_Hexbuster_Hammer#Advanced_Properties

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,117 @@
...
+	duration := 813
+	if c.Base.Cons >= 6 {
+		duration = 1053
+	}
+
+	c.AddStatus(burstKey, duration, false)
+
+	if c.Base.Cons >= 2 {
+		c.c2Buff[attributes.ATKP] = 0.10
+		c.AddStatMod(character.StatMod{
```
jwalk-511: """
Prune gently swings the Witchlure Bell, dealing Anemo DMG to nearby opponents. After this, the Bell will enter the Hunter-Seeker mode.
"""

Is Prune in Hunter-Seeker mode as soon as burst is used, or not until after the hitmark?
Or, does the burst initial hit benefit from C2 or only the dot?

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,117 @@
...
+		if target == nil {
+			return
+		}
+
+		c.Core.QueueAttack(
+			ai,
+			combat.NewCircleHitOnTarget(target, nil, 1.6),
+			0,
+			0,
+			c.makeC2CB,
```
jwalk-511, line 98: Only a converted hammer hit should trigger c2, not the initial bell hit

internal/characters/prune/charge.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (c *char) ChargeAttack(p map[string]int) (action.Info, error) {
+	ai := info.AttackInfo{
+		ActorIndex:         c.Index(),
+		Abil:               "Charge Attack",
+		AttackTag:          attacks.AttackTagExtra,
+		ICDTag:             attacks.ICDTagNone,
+		ICDGroup:           attacks.ICDGroupDefault,
+		StrikeType:         attacks.StrikeTypeDefault,
```
jwalk-511: wiki says charge is blunt

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,174 @@
...
+			if atk.Info.ActorIndex != c.Index() {
+				return
+			}
+
+			// reaction must have been triggered by the initial skill hit
+			if atk.Info.Abil != "Ring-A-Ding-Ding! Hexhunter Chime DMG" {
+				return
+			}
+
+			// A1 has a shared 1.2s ICD across all four swirl elements
```
jwalk-511: Comment should be Skill Swirl check, not A1

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,174 @@
...
+
+			// skill recast has different cancel frame
+			skillFrames[action.ActionSkill] = 28
+		}
+	}
+
+	c.Core.Events.Subscribe(event.OnSwirlCryo, swirlfunc(attributes.Cryo), "prune-skill-cryo")
+	c.Core.Events.Subscribe(event.OnSwirlElectro, swirlfunc(attributes.Electro), "prune-skill-electro")
+	c.Core.Events.Subscribe(event.OnSwirlHydro, swirlfunc(attributes.Hydro), "prune-skill-hydro")
+	c.Core.Events.Subscribe(event.OnSwirlPyro, swirlfunc(attributes.Pyro), "prune-skill-pyro")
```
jwalk-511, line 112: will also need stellar swirl here

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (c *char) skillConvert() (action.Info, error) {
+	ai := info.AttackInfo{
+		ActorIndex:         c.Index(),
+		Abil:               "Clang Clang! Witch-tribution Comes! DMG",
+		AttackTag:          attacks.AttackTagElementalArt,
+		ICDTag:             attacks.ICDTagNone,
+		ICDGroup:           attacks.ICDGroupDefault,
+		StrikeType:         attacks.StrikeTypeDefault,
```
jwalk-511: https://genshin-impact.fandom.com/wiki/Ring-A-Ding-Ding!_Hexhunter_Chime#Advanced_Properties
Converted skill is labelled as blunt

internal/characters/prune/prune.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func NewChar(s *core.Core, w *character.CharWrapper, p info.CharacterProfile) error {
+	c := char{}
+	c.Character = tmpl.NewWithWrapper(s, w)
+
+	c.EnergyMax = 70
+	c.NormalHitNum = normalHitNum
+	c.SkillCon = 3
+	c.BurstCon = 5
```
jwalk-511: burst and skill con should be swapped

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+
+			c.Core.QueueAttack(
+				ai,
+				combat.NewCircleHitOnTarget(target, nil, 2.3),
+				0,
+				45,
+				c.makeA4CB,
+				c.makeC1CB,
+				c.c4Ricochet(ele, attacks.AttackTagElementalBurst),
+				c.makeC2CB,
```
jwalk-511, line 78: similar question to the initial hit comment on burst; do you know that the +5% is only supposed to be added after the hit occurs, or is it supposed to benefit the hit that triggered it too?

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,205 @@
...
+			// A1 has a shared 1.2s ICD across all four swirl elements
+			if c.StatusIsActive(a1SwirlCheckICDKey) {
+				return
+			}
+			c.AddStatus(a1SwirlCheckICDKey, 72, false)
+
+			// queue passive
+			ai := info.AttackInfo{
+				ActorIndex: c.Index(),
+				Abil:       "Verdict and Punishment (A1)",
```
jwalk-511: personal preference- this should be called "Banehunter Oathhammer (A1)", to make it more clear that this triggers c2. But it should be ok to leave like this if preferred

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,173 @@
...
+	if c.Base.Cons < 2 || a.Target.Type() != info.TargettableEnemy || !c.StatusIsActive(burstKey) {
+		return
+	}
+
+	c.c2Buff[attributes.ATKP] = math.Min(c.c2Buff[attributes.ATKP]+0.05, 0.40)
+}
+
+func (c *char) c4Ricochet(ele attributes.Element, tag attacks.AttackTag) info.AttackCBFunc {
+	if c.Base.Cons < 4 {
+		return nil
```
jwalk-511: this might cause an error? To be safe, you can put this inside the return func after line 45. I'm not sure what happens if you pass `nil` as a callback func.

If you've already tested this, please ignore

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return func(a info.AttackCB) {
+		// has to be on the enemy hit by the attack that call this
+		if a.Target.Type() != info.TargettableEnemy {
+			return
+		}
+
+		// C1 has 0.1s ICD
```
jwalk-511: comment should say c4, not c1

Also, is this icd shared between skill and burst bounces?

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		// queue the attack
+		ai := info.AttackInfo{
+			ActorIndex: c.Index(),
+			Abil:       "Banehunter Oathhammer Ricochet (C4)",
+			AttackTag:  tag,
+			ICDTag:     attacks.ICDTagNone,
+			ICDGroup:   attacks.ICDGroupDefault,
+			StrikeType: attacks.StrikeTypeBlunt,
```
jwalk-511, line 117: blunt attack needs poise damage

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,173 @@
...
+			return
+		}
+
+		// triggerer must be affected by tolling rally
+		triggerer := chars[atk.Info.ActorIndex]
+		if !triggerer.StatusIsActive(a4Key) {
+			return
+		}
+
+		c.AddStatus(c6WindowKey, 5*60, false)
```
jwalk-511: hitlag check here; I'd assume it is hitlag affected

jwalk-511 (changes requested):
did not review frames

internal/characters/prune/asc.go
Khoi0612: I believe from [here ](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15180)to [here ](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15392)means that when the hammer hit, she will calculate the buff, store it in `_ABILITY_Prune_AddRatio`, remove any previous modifier, and reattached the modiffier with the newly calcualted buff, all of which will repeat the next time it hit.

internal/characters/prune/asc.go
Khoi0612: I think [this](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L12707) means that the poise dmg is 20.

internal/characters/prune/cons.go
Khoi0612: Couldn't find one on the wiki, so just put [this ](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L17971)in the dm as reference.

internal/characters/prune/burst.go
Khoi0612: I have [tested](https://docs.google.com/spreadsheets/d/1Mbx_FRboChq7JeG7uDJo6kdvsxk856SSRXACndSiR_Q/edit?usp=sharing) and apparently the burst intial hit doesn't get affected by the 10% from C2.

internal/characters/prune/burst.go
Khoi0612: This is wrong from what I've [tested](https://docs.google.com/spreadsheets/d/1Mbx_FRboChq7JeG7uDJo6kdvsxk856SSRXACndSiR_Q/edit?usp=sharing), both the converted hammer and the bell dot increment the C2 buff. Also [this](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L14749).

internal/characters/prune/asc.go
Khoi0612: It's added [after](https://docs.google.com/spreadsheets/d/1Mbx_FRboChq7JeG7uDJo6kdvsxk856SSRXACndSiR_Q/edit?usp=sharing) the hit occur, so the attack that trigger it won't be benefited from it.

internal/characters/prune/cons.go
Khoi0612: Yes because in the dm the icd is checked in the c4 handler itself, and both e2 and a1 are tagged as `Elemental_Hammer`, so they go through the same c4 handler and therefore share the same icd

internal/characters/prune/cons.go
Khoi0612: It is hitlag affected, just put [this](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15898) here for reference.

internal/characters/prune/cons.go
Khoi0612: I've tested by running a c0 prune team and found no error. But I'll move it in the return function just in case.

internal/characters/prune/burst.go
jwalk-511: Ok I just can't read, I totally missed "and when the **Witchlure Bell and** the Banehunter Oathhammer that "

jwalk-511 (approved):

internal/characters/prune/charge.go
```diff
@@ -0,0 +1,58 @@
...
+		StrikeType:         attacks.StrikeTypeBlunt,
+		PoiseDMG:           120,
+		Element:            attributes.Anemo,
+		Durability:         25,
+		Mult:               charge[c.TalentLvlAttack()],
+		HitlagFactor:       0.01,
+		HitlagHaltFrames:   0.03 * 60,
+		CanBeDefenseHalted: true,
+	}
+
```
Charlie-Zheng, line 49: Need to account for the ~10f CA windup in the impl. Need find specific scenarios where the ~10f windup shouldn't be added

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+	burstFrames[action.ActionDash] = 74
+	burstFrames[action.ActionJump] = 73
+	burstFrames[action.ActionSwap] = 71
+}
+
+func (c *char) Burst(p map[string]int) (action.Info, error) {
+	duration := 813
+	if c.Base.Cons >= 6 {
+		duration = 1053
+	}
```
Charlie-Zheng: Make a func in cons.go

```
func (c *char) c6BurstBonusDur() int {
    if c.Base.Cons >= 6 {
        return 0
    }
    return 4*60
}
```

and change this to
```suggestion
    duration := 813 + c.c6BurstBonusDur()
```

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+
+	c.AddStatus(burstKey, duration, false)
```
Charlie-Zheng: Are you sure this status should be added immediately on casting the burst?

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+	if c.Base.Cons >= 2 {
+		c.c2Buff[attributes.ATKP] = 0.10
+		c.AddStatMod(character.StatMod{
+			Base:         modifier.NewBase("prune-c2", duration),
+			AffectedStat: attributes.ATKP,
+			Amount: func() []float64 {
+				return c.c2Buff
+			},
+		})
+	}
```
Charlie-Zheng: Move to a func in cons.go

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+		StrikeType: attacks.StrikeTypeDefault,
+		Element:    attributes.Anemo,
+		Durability: 25,
+		Mult:       burst[c.TalentLvlBurst()],
+	}
+
+	c.Core.QueueAttack(
+		ai,
+		combat.NewCircleHitOnTarget(c.Core.Combat.Player(), nil, 6.5),
+		0,
```
Charlie-Zheng: Are you sure she snapshots on cast immediately? Typically melee hits don't snapshot at all (so for gcsim, snapshot on the hitmark frame)

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+			Amount: func() []float64 {
+				return c.c2Buff
+			},
+		})
+	}
+
+	c.burstSrc = c.Core.F
+	for i := 137; i < duration; i += 117 {
+		c.Core.Tasks.Add(c.burstTick(c.burstSrc), i)
+	}
```
Charlie-Zheng: It's better to just queue follow up attacks in the c.burstTick() function itself
```suggestion
	c.Core.Tasks.Add(c.burstTick(c.burstSrc), 137)
```

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		c.Core.QueueAttack(
+			ai,
+			combat.NewCircleHitOnTarget(target, nil, 1.6),
+			0,
+			0,
+			c.makeC2CB,
+		)
+	}
```
Charlie-Zheng, line 101: Add queuing the next hit in this function


```suggestion
		)
		c.Core.Tasks.Add(c.burstTick(src), 117)
	}
```

internal/characters/prune/burst.go
```diff
@@ -0,0 +1,115 @@
...
+		ai := info.AttackInfo{
+			ActorIndex:   c.Index(),
+			Abil:         "Witchlure Bell",
+			AttackTag:    attacks.AttackTagElementalBurst,
+			ICDTag:       attacks.ICDTagNone,
+			ICDGroup:     attacks.ICDGroupDefault,
+			Element:      attributes.Anemo,
+			Durability:   25,
+			Mult:         burstDot[c.TalentLvlBurst()],
+			HitlagFactor: 0.01,
```
Charlie-Zheng: HitlagFactor doesn't do anything when the attack doesn't have hitlag, so just delete this

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,176 @@
...
+				return
+			}
+			c.AddStatus(skillSwirlCheckICDKey, 12, false)
+
+			// allow skill to be recasted
+			c.skillConvertEle = ele
+			c.AddStatus(skillRecastWindowKey, 364, true)
+
+			// skill recast has different cancel frame
+			skillFrames[action.ActionSkill] = 28
```
Charlie-Zheng: Do not modify this during runs, it leads to a race condition because every thread is modifying the same variable constantly. Change the `Frames:` func to return 28 if skillRecastWindowKey is active

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,176 @@
...
+	skillConvertFrames[action.ActionSwap] = 65
+}
+
+func (c *char) Skill(p map[string]int) (action.Info, error) {
+	// if this is second press and had swirled, do the convert dmg
+	if c.StatusIsActive(skillRecastWindowKey) {
+		return c.skillConvert()
+	}
+
+	skillFrames[action.ActionSkill] = 41
```
Charlie-Zheng: See the issue on line 109. Causes race condition due to multiple threads modifying the same variable

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,176 @@
...
+			// reaction must have been triggered by the initial skill hit
+			if atk.Info.Abil != "Ring-A-Ding-Ding! Hexhunter Chime DMG" {
+				return
+			}
+
+			// skill has a shared 1.2s ICD across all four swirl elements
+			if c.StatusIsActive(skillSwirlCheckICDKey) {
+				return
+			}
+			c.AddStatus(skillSwirlCheckICDKey, 12, false)
```
Charlie-Zheng: Change to this 0.2 * 60 
Also update the comment/the duration to be consistent, currently it's 1.2s in comment and 0.2s in code

internal/characters/prune/skill.go
```diff
@@ -0,0 +1,176 @@
...
+		Mult:               skillConvert[c.TalentLvlSkill()],
+		HitlagFactor:       0.01,
+		HitlagHaltFrames:   0.05 * 60,
+		CanBeDefenseHalted: true,
+	}
+
+	c.Core.QueueAttack(
+		ai,
+		combat.NewCircleHitOnTarget(c.Core.Combat.Player(), info.Point{Y: 1.5}, 2.5),
+		0,
```
Charlie-Zheng: Are you sure this snapshots on cast? Typically melee attacks snapshot on hit

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+			// reaction must have been triggered by the burst dot
+			if atk.Info.Abil != "Witchlure Bell" {
+				return
+			}
+
+			// A1 has a shared 1.2s ICD across all four swirl elements
+			if c.StatusIsActive(a1SwirlCheckICDKey) {
+				return
+			}
+			c.AddStatus(a1SwirlCheckICDKey, 72, false)
```
Charlie-Zheng: Change to 1.2 * 60

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+				PoiseDMG:   20,
+				Element:    ele,
+				Durability: 0,
+				Mult:       1.5,
+			}
+
+			c.Core.QueueAttack(
+				ai,
+				combat.NewCircleHitOnTarget(target, nil, 2.3),
+				0,
```
Charlie-Zheng: Does this snapshot? (Default assumption is no snapshot for melee attacks, and snapshot on projectile spawn for projectiles)

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+				c.c4Ricochet(ele, attacks.AttackTagElementalBurst),
+				c.makeC2CB,
+			)
+		}
+	}
+
+	c.Core.Events.Subscribe(event.OnSwirlCryo, swirlfunc(attributes.Cryo), "prune-burst-cryo")
+	c.Core.Events.Subscribe(event.OnSwirlElectro, swirlfunc(attributes.Electro), "prune-burst-electro")
+	c.Core.Events.Subscribe(event.OnSwirlHydro, swirlfunc(attributes.Hydro), "prune-burst-hydro")
+	c.Core.Events.Subscribe(event.OnSwirlPyro, swirlfunc(attributes.Pyro), "prune-burst-pyro")
```
Charlie-Zheng: Subscriptions should be `prune-a1-`

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+	// TODO: Add subscriptions for stellar-swirl when it's implemented
+}
+
+func (c *char) makeA4CB(a info.AttackCB) {
+	if c.Base.Ascension < 4 || a.Target.Type() != info.TargettableEnemy {
+		return
+	}
```
Charlie-Zheng:
```suggestion
	if c.Base.Ascension < 4 {
		return
	}
	
	if a.Target.Type() != info.TargettableEnemy {
		return
	}
```

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+	c.a4Buff = make([]float64, attributes.EndStatType)
```
Charlie-Zheng, line 94: Allocate this array in a `c.a4Init()` function, instead of every time the CB is triggered

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+	atk := c.SelectStat(true, attributes.BaseATK, attributes.ATKP, attributes.ATK).TotalATK()
+	c.a4Buff[attributes.DmgP] = math.Min(math.Max(atk-2000, 0)*0.00025, 0.5)
+
+	for i, char := range c.Core.Player.Chars() {
+		if i == c.Index() {
+			continue // nothing for prune, need testing
```
Charlie-Zheng: It doesn't need testing, the talent says 

> that has undergone Elemental Conversion, __other__ nearby party members will gain the Tolling Rally effect for 5s

which would exclude Prune

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		char.AddAttackMod(character.AttackMod{
+			Base: modifier.NewBaseWithHitlag(a4Key, 5*60), // 5 s
+			Amount: func(
+				atk *info.AttackEvent,
+				target info.Target,
+			) []float64 {
+				if atk.Info.AttackTag != attacks.AttackTagNormal && atk.Info.AttackTag != attacks.AttackTagExtra && atk.Info.AttackTag != attacks.AttackTagPlunge && atk.Info.AttackTag != attacks.AttackTagElementalArt && atk.Info.AttackTag != attacks.AttackTagElementalBurst {
+					return nil
+				}
```
Charlie-Zheng: Use a switch case for this, default to return nil
```suggestion
		switch atk.Info.AttackTag {
		case attacks.AttackTagElementalBurst:
		case attacks.AttackTagElementalArt:
		case attacks.AttackTagElementalArtHold:
		case attacks.AttackTagNormal:
		case attacks.AttackTagExtra:
		case attacks.AttackTagPlunge:
		default:
			return nil
		}
  		return c.a4Buff
```

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+			if len(args) < 2 {
+				return
+			}
+
+			// reaction target must be an enemy
+			if _, ok := args[0].(*enemy.Enemy); !ok {
+				return
+			}
+
+			// reaction must have been triggered by a character attack
```
Charlie-Zheng: This comment isn't needed; delete it

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+			atk, ok := args[1].(*info.AttackEvent)
+			if !ok {
+				return
+			}
+
+			triggererIndex := atk.Info.ActorIndex
+			if triggererIndex < 0 || triggererIndex >= len(chars) {
+				return
+			}
```
Charlie-Zheng: This check isn't necessary; delete it

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+				AffectedStat: attributes.ATKP,
+				Amount: func() []float64 {
+					return c.hexTeamBuff
+				},
+			})
+
+			c.Core.Log.NewEvent("prune hex team buff triggered", glog.LogCharacterEvent, c.Index()).
+				Write("triggerer", triggererIndex).
+				Write("atk percent", c.hexTeamBuff[attributes.ATKP]).
+				Write("expiry", c.Core.F+5*60)
```
Charlie-Zheng: This isn't needed, the information contained is part of the statmod already

internal/characters/prune/asc.go
```diff
@@ -0,0 +1,203 @@
...
+				AffectedStat: attributes.ATKP,
+				Amount: func() []float64 {
+					return c.hexSelfBuff
+				},
+			})
+
+			c.Core.Log.NewEvent("prune hex self buff triggered", glog.LogCharacterEvent, c.Index()).
+				Write("triggerer", triggererIndex).
+				Write("atk percent", c.hexSelfBuff[attributes.ATKP]).
+				Write("expiry", c.Core.F+5*60)
```
Charlie-Zheng: This isn't needed, the information contained is part of the statmod already

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+	c4ICDKey    = "prune-c4-icd"
+	c6WindowKey = "prune-c6-window"
+)
+
+func (c *char) makeC1CB(a info.AttackCB) {
+	if c.Base.Cons < 1 || a.Target.Type() != info.TargettableEnemy || c.StatusIsActive(c1ICDKey) {
+		return
+	}
+
+	c.AddStatus(c1ICDKey, 108, false)
```
Charlie-Zheng: Use 1.8*60

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+const (
+	c1ICDKey    = "prune-c1-icd"
+	c4ICDKey    = "prune-c4-icd"
+	c6WindowKey = "prune-c6-window"
+)
+
+func (c *char) makeC1CB(a info.AttackCB) {
+	if c.Base.Cons < 1 || a.Target.Type() != info.TargettableEnemy || c.StatusIsActive(c1ICDKey) {
+		return
+	}
```
Charlie-Zheng:
```suggestion
	if c.Base.Cons < 1 {
		return
	}
	
	if a.Target.Type() != info.TargettableEnemy {
		return
	}
	
	if c.StatusIsActive(c1ICDKey) {
		return
	}
```

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+
+	c.AddStatus(c1ICDKey, 108, false)
+	c.AddEnergy("prune-c1", 2)
+}
+
+func (c *char) makeC2CB(a info.AttackCB) {
+	if c.Base.Cons < 2 || a.Target.Type() != info.TargettableEnemy || !c.StatusIsActive(burstKey) {
+		return
+	}
```
Charlie-Zheng:
```suggestion
	if c.Base.Cons < 2 {
		return
	}
	
	if a.Target.Type() != info.TargettableEnemy {
		return
	}
	
	if !c.StatusIsActive(burstKey) {
		return
	}
```

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+			PoiseDMG:   15,
+			Element:    ele,
+			Durability: 0,
+			Mult:       0.8,
+		}
+
+		c.Core.QueueAttack(
+			ai,
+			combat.NewCircleHitOnTarget(target, nil, 1.5),
+			0,
```
Charlie-Zheng: Does this use the same snapshot as the original oathhammer?

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+		AffectedStat: attributes.ATK,
+		Amount: func() []float64 {
+			if !c.StatusIsActive(c6WindowKey) {
+				return nil
+			}
+			return c.c6Buff
+		},
+	})
+
+	// give the currently active character 350 ATK
```
Charlie-Zheng: Was this tested to be an "instant" on field buff, or a linger buff that's reapplied every X frames?

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+		triggerer := chars[atk.Info.ActorIndex]
+		if !triggerer.StatusIsActive(a4Key) {
+			return
+		}
+
+		c.AddStatus(c6WindowKey, 5*60, true)
+
+		c.Core.Log.NewEvent("prune c6 triggered", glog.LogCharacterEvent, c.Index()).
+			Write("atk", c.c6Buff[attributes.ATK]).
+			Write("expiry", c.Core.F+5*60)
```
Charlie-Zheng: This event isn't needed

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,174 @@
...
+			if !c.StatusIsActive(c6WindowKey) {
+				return nil
+			}
+			return c.c6Buff
+		},
+	})
+
+	// give the currently active character 350 ATK
+	for _, ch := range chars {
+		char := ch
```
Charlie-Zheng:
```suggestion
	for _, char := range chars {
```

internal/characters/prune/prune.go
```diff
@@ -0,0 +1,72 @@
...
+
+	w.Character = &c
+
+	return nil
+}
+
+func (c *char) Init() error {
+	if c.Base.Cons >= 2 {
+		c.c2Buff = make([]float64, attributes.EndStatType)
+	}
```
Charlie-Zheng: Move to cons.go `func (c *char) c2Init() {`

Charlie-Zheng (changes requested):

internal/characters/prune/burst.go
Khoi0612: It shouldn't based on dm: [burst cast at 0](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L14953), [initial hit at normalized 0.47](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L14989), and [status start at 0.75](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15095). I'll delay the status add and c2 intial buff by 57f, as the [status duration](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L10463) in dm is `(12 + 0.6) * 60 = 756f` whereas the recorded frame for burst duration is 813f, which is 57f difference. The same can also be seen with [c6 duration](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L10607).

internal/characters/prune/cons.go
Khoi0612: I assume it use the snapshot when the original oathammer hit, but I don't know how to verify it

internal/characters/prune/cons.go
Khoi0612: In the DM, the buff last for `5 + 0.5 = 5.5s` for some reason (both [self buff](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15943) and [team buff handler](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L15998)). It checks every [0.5s](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L16134) if the character trigger reaction is the active character who is not prune and has tolling rally, then [apply](github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L16089) the buff or [remove](https://github.com/iam-akuzihs/binout/blob/2da39f60a21ff03924de4f4c865a9b9ad29c69b2/v7.0.0-be8d439be0/BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Prune.json#L16070) it accordingly.

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,224 @@
...
+	c.AddStatus(c1ICDKey, 1.8*60, false)
+	c.AddEnergy("prune-c1", 2)
+}
+
+func (c *char) c2(duration int) {
+	if c.Base.Cons < 2 {
+		return
+	}
+
+	c.c2Buff[attributes.ATKP] = 0.10
```
Charlie-Zheng, line 53: nit: this can be set once in c.c2Init() and then we don't need to set it here

internal/characters/prune/cons.go
```diff
@@ -0,0 +1,224 @@
...
+	})
+}
+
+func (c *char) c2Init() {
+	if c.Base.Cons < 2 {
+		return
+	}
+
+	c.c2Buff = make([]float64, attributes.EndStatType)
+}
```
Charlie-Zheng: nit: move it before c.c2()

internal/characters/prune/cons.go
Charlie-Zheng: If you want to use the same snapshot, you can use the following

```go
	c.Core.QueueAttackWithSnap(
		ai,
		a.AttackEvent.Snapshot,
		combat.NewCircleHitOnTarget(target, nil, 1.5),
		63,
		c.makeA4CB,
		c.makeC1CB,
		c.makeC2CB,
	)
```

Charlie-Zheng (changes requested):

Charlie-Zheng (dismissed):

internal/characters/prune/cons.go
Charlie-Zheng: Sorry I misunderstood how Prune C2 works, your original impl was correct, I'll update it back

Charlie-Zheng (dismissed):

Charlie-Zheng (dismissed):

Charlie-Zheng (changes requested):
It's missing N1 -> W, N2 -> W, and N3 -> W frames

Charlie-Zheng (dismissed):

Charlie-Zheng (approved):

PR #2741, opened 2026-09-03
Validate cons, talent and refine ranges, add ignore_skill_cooldown

### LINKED ISSUE #2621: Refines underflow on UI without underflowing in sim
https://gcsim.app/sh/jMm7b9pHcgPz

On the UI Sucrose appears to have R5 TTDS but in the config she actually has `refine=4294967301`

Image

The actual computations proceed as if TTDS is refine=4294967301

AJTDaedalus:
Fixes #2621

`cons`, `talent` and `refine` are parsed with no bounds check. They are serialized as int32, so an out of range value wraps when the config reaches the UI. The config in the issue uses `refine=4294967301`, which is `5 + 2^32`, so the weapon displays as R5 while the sim runs the passive at the full value.

Bounding refine also closes a bypass that is used on purpose. An out of range refine on a Sacrificial weapon makes its reset chance always proc and drives its ICD negative, which removes skill cooldown entirely. There is no supported way to do that today, so this adds `ignore_skill_cooldown` next to the existing `ignore_burst_energy`.

| Added | Detail |
|---|---|
| Range check on `cons`, `talent`, `refine` | 0-6, 1-10, 1-5. Rejected at parse time with the token position, matching the existing base/max level check |
| `ignore_skill_cooldown` option | Skill can be cast while on cooldown |
| Skipped in `SetCD`, not `ActionReady` | Charges stay at max and the queue stays empty, so `Cooldown`, `ResetActionCooldown` and `ReduceActionCooldown` stay consistent. Gating `ActionReady` alone panics on `AvailableCDCharge < 0` at the second cast |
| DANGER warning on the results page | Same treatment as `ignore_burst_energy` |
| Docs and editor keyword | Config reference table and `ui_editor.js.templ` |
| No schema version bump | The field is additive, so old results stay valid |

| Test | Result |
|---|---|
| `TestCharFieldRanges` | pass. Each invalid value paired with a valid one differing only in the number |
| `TestIgnoreSkillCooldownOption` | pass. Option parses to true, false and defaults off |
| `TestIgnoreSkillCooldown` | pass. Five casts, no panic, cooldown stays 0 |
| `TestIgnoreSkillCooldownMultiCharge` | pass. Klee keeps 2 charges, no charges-above-max panic |
| `TestSkillCooldownAppliesWhenFlagOff` | pass. Cooldown still applies with the flag off |
| `go test ./internal/... ./pkg/gcs/... ./pkg/reactable/...` | pass, 11 packages |
| `golangci-lint run` | 0 issues |

Notes and open questions:

- Nothing in the repo violates the new bounds, and no first party producer can emit an out of range value. Team builder defaults, Enka import and GOOD import are all in range. Existing share links still render, since the viewer only validates when the config is modified.
- Sacrificial does not proc when `ignore_skill_cooldown` is on. It already early returns when the skill is off cooldown, so it never resets a cooldown that was never set.
- Ability windows that are not cooldowns, such as Fischl's Oz recast, are still gated. Say the word if that should change.
- Hard parse error rather than clamping. Clamping would run a value the config does not say, which is the same disagreement this issue is about. Easy to switch if you prefer clamping.
- English locale only. The other six fall back rather than get a machine translation.


shizukayuki:
this doesn't fix the mentioned issue and `ignore_skill_cooldown` won't be merged as is.

PR #2745, opened 2026-09-05
Fix Varka's CD mechanisms

Charlie-Zheng:


Charlie-Zheng:
Ran dbsnapshot and confirmed that the changes in cooldown.go don't cause any diffs

PR #2840, opened 2026-09-13
fix(workers): resolve preview-card assets in-process; skip caching fallback renders

### LINKED ISSUE #2832: feat(workers): fetch preview-card assets in the Worker; skip caching renders that fall back
## Agent Brief

**Category:** enhancement
**Follow-up to:** #2830

**Summary:** Move the preview card's asset fetching out of Satori and into the Worker. Instead of letting Satori fetch each portrait/weapon/artifact `` over the network at render time, the Worker pre-fetches the assets, inlines them as `data:` URIs passed into `SatoriPreviewCard`, and — the immediate win — **skips the `caches.default` store when any asset was missing and a fallback had to be used**, so a card isn't frozen for 60 days with a broken/placeholder image.

**Current behavior (after #2830):**
- `handleOgPreview` (`ui/packages/workers/src/preview/handleOgPreview.tsx`) renders `<SatoriPreviewCard>` and stores the resulting PNG in `caches.default` with a 60-day TTL.
- `SatoriPreviewCard` builds image URLs internally from a string `assetBase` prop: `${assetBase}/avatar/<name>.png`, `${assetBase}/weapons/<weapon>.png`, `${assetBase}/artifacts/<set>_flower.png`, and the empty-slot `${assetBase}/misc/nahida.png` (`ui/packages/components/src/Cards/SatoriPreviewCard/Portraits.tsx`).
- **Satori itself** fetches every such `` at render time (it accepts absolute `http(s)` URLs and `data:` URIs; anything else throws). Those URLs resolve to `/api/assets/*`, served by `handleAssets` proxying `env.ASSETS_ENDPOINT` (assets live on a dedicated asset endpoint/server, not in the Worker).
- The Worker has **no visibility** into whether an asset resolved. If an asset is genuinely unavailable (e.g. a brand-new character whose portrait hasn't been uploaded yet), the render still produces a PNG — and that PNG is cached for 60 days, so the broken card sticks until the TTL expires.
- The element backgrounds are already inlined as base64 `data:` URIs (`backgrounds.ts`), so the "inline assets into the card" pattern is already proven here — just not for the portrait imagery.

**Desired behavior:**
- The Worker enumerates the assets a given `SimulationResult` needs, fetches them (checking `resp.ok`), and passes the bytes into the card as `data:` URIs. Satori makes zero outbound image fetches.
- When an asset fetch is **not** OK, the Worker substitutes a fallback (e.g. the existing `misc/nahida.png` placeholder, or a solid slot) **and does not write the render to `caches.default`** — so once the missing asset lands, the next request re-renders and caches the correct card.
- When all assets resolve, behavior is unchanged: render once, store in `caches.default` for the full TTL.

**Key details / gotchas:**
- **Seam:** bytes can't flow through today's string `assetBase`. Add a resolver seam to `SatoriPreviewCard` — a `resolveAsset(path: string) => string` prop that returns a `data:` URI (or URL), defaulting to `` (p) => `${assetBase}/${p}` `` so Storybook / bare `satori()` callers keep working unchanged. Keep the path-building (avatar/weapon/artifact/nahida) inside the card; the Worker only overrides the URL→bytes step. This is the single card-side change and must not alter layout.
- **Absolute-URL requirement:** Satori rejects non-absolute ``; `data:` URIs satisfy it. Base64 inflates ~33% and all images sit in memory during the render — fine for the handful of small PNGs per card (≤4 per portrait × 4 portraits).
- **Fetch in parallel:** resolve all assets with `Promise.all` so the added Worker fetches don't serialize the render.
- **Dedup:** the same asset can appear across portraits; fetch each unique path once.
- **Only the cache store is conditional:** a fallback-containing render should still be **served** (200, `image/png`) — it just must not be cached. Non-fallback renders cache as today.

**Route/handler changes (worker):**
- In `handleOgPreview`: enumerate + fetch assets, build the `resolveAsset` map/function, pass it to `<SatoriPreviewCard>`, and gate the `caches.default.put` on "no fallback used".

**Acceptance criteria:**
- [ ] Satori performs no network image fetches during render; all card imagery is supplied by the Worker as `data:` URIs — verified under `wrangler dev` (e.g. no `/api/assets/*` fetches observed during a render).
- [ ] A card whose every asset resolves renders identically to today and is stored in `caches.default`.
- [ ] A card with at least one unavailable asset renders with a fallback, returns 200 `image/png`, and is **not** written to `caches.default` (a subsequent request re-renders).
- [ ] `SatoriPreviewCard` gains only an injectable asset-resolver seam; its layout and default (`assetBase`-based) behavior are unchanged, and existing non-Worker callers (Storybook / `satori()`) still render.
- [ ] Assets are fetched in parallel and de-duplicated.
- [ ] The Worker builds; no dead references.

**Out of scope:**
- Serving assets from R2. Assets currently live on a dedicated endpoint (`ASSETS_ENDPOINT`); this issue only moves the *fetch* into the Worker. Once that's in place, sourcing those bytes from an R2 binding instead of an HTTP fetch is a possible **future** bonus — track it separately, not here.
- Changing `SatoriPreviewCard`'s layout or `DEFAULT_ASSET_BASE`.
- Reworking `handleAssets` / the `/api/assets/*` proxy.

srliao:
The live OG preview render (`/api/preview/*`) pre-fetches each card image in the Worker and inlines it as a `data:` URI so Satori makes zero outbound image fetches. Post-#2842, `/api/assets/*` is served in-process, so this routes the pre-fetch through the Worker's **own** resolver instead of an outbound HTTP call — and fixes the caching bug that froze broken cards for 60 days.

## New
- Shared `resolveAsset(subpath, env, ctx)`, extracted from the `/api/assets/*` handler: the R2 → source-host → placeholder resolution, emitting `X-Gcsim-Asset`. Both the public asset route and the preview pre-fetcher now use it.

## Changed
- Preview asset pre-fetch resolves in-process via `resolveAsset` — no outbound `ASSETS_ENDPOINT` fetch, no duplicated resolution logic.
- "Skip caching this render" is now driven by `X-Gcsim-Asset: fallback`, not `resp.ok`. A genuine miss is served as HTTP 200 `misc/default.png`, so the old `resp.ok` check treated a miss as success and cached the broken card for the full 60-day TTL; it now renders, serves `200 image/png`, and is **not** cached, so it re-renders once the asset lands.
- Missing assets show the Worker's real `misc/default.png` placeholder instead of a transparent gap.

## Removed
- The 1×1 transparent placeholder constant (misses now show `misc/default.png`).
- The unused `ASSETS_ENDPOINT` binding.

---

The `/api/assets/*` handler's external behavior (routes, `X-Gcsim-Asset` values, edge caching, R2 write-through, TTLs) is unchanged — the resolver extraction is a pure relocation. All acceptance criteria in the brief are met.

Closes #2832

🤖 Generated with [Claude Code](https://claude.com/claude-code)


srliao:
> *This was generated by AI during triage.*

## Agent Brief (supersedes the original #2832 brief)

**Category:** enhancement
**Re-triaged:** `upstream/main` has moved since #2832 was first written. #2842 landed R2-backed asset serving **inside this Worker**: `/api/assets/*` is now served in-process by the Worker's own asset handler, which resolves `avatar|weapons|artifacts/<key>.png` through R2 → source host (mihoyo) → placeholder, and static files straight from R2. The original brief's "serving assets from R2 is out of scope / assets live on a dedicated `ASSETS_ENDPOINT`" premise is now **stale**.

**Summary:** Fix how this PR fetches card assets. Route the preview render's asset fetching through the Worker's **own** in-process asset resolution instead of an outbound `ASSETS_ENDPOINT` HTTP call, and decide "skip caching this render" from the resolver's miss signal rather than `resp.ok`.

**Current behavior (of the diff):**
The PR pre-fetches each portrait/weapon/artifact image, inlines it as a `data:` URI, feeds it to `SatoriPreviewCard` via a `resolveAsset` seam, and skips `caches.default` when a fetch "failed" — the right shape. But the fetch step is wrong for the post-#2842 world:

- It fetches `${env.ASSETS_ENDPOINT}/api/assets/<path>` — an **outbound** call to the old external asset server — duplicating the resolution logic (R2 lookup, source-host fallthrough, R2 write-through, placeholder) that the Worker's own asset handler now performs in-process.
- It gates the "did this asset resolve?" decision on `resp.ok`. But the Worker's asset handler returns **HTTP 200** for a genuinely-missing asset: it serves the `misc/default.png` placeholder with `Cache-Control: no-cache` and the header **`X-Gcsim-Asset: fallback`**. So `resp.ok` is `true` on a miss → the placeholder bytes are inlined, `usedFallback` stays `false`, and the broken card **is cached for the full 60-day TTL** — the exact failure #2832 set out to prevent. This is the concrete defect behind the "avatar/bg didn't fully load" reports: the render already `await`s every asset before invoking Satori (there is no draw-before-fetch race for the inlined imagery), but a missing asset silently becomes the placeholder *and* gets frozen in cache.

**Desired behavior:**
- The preview render resolves each asset through the **Worker's own** asset resolution, in-process — no `ASSETS_ENDPOINT`, no second copy of the R2/source-host/placeholder logic.
- A render is written to `caches.default` **only when every asset resolved to a real asset**. If any asset came back as the placeholder (resolver miss), the render is still produced and served (`200 image/png`) but **not cached**, so the next request re-renders once the asset lands.
- Missing assets show the Worker's real `misc/default.png` placeholder (what `/api/assets/*` itself serves), not a transparent gap.

**Key interfaces / seams:**
- **Extract a shared asset resolver.** The Worker's asset handler today is effectively `edge-cache wrapper` + `resolve(subpath) → Response`, where the inner resolve step already emits the `X-Gcsim-Asset` status header (`hit` / `origin` / `fallback`, plus `edge-hit` on a cache hit). Pull that inner resolve step out into a shared function that both callers use:
  - the public `/api/assets/*` HTTP handler keeps its edge-cache wrapper around it — **its external behavior (routes, response headers including `X-Gcsim-Asset`, edge caching, R2 write-through, TTLs) must be unchanged**; this is a behavior-preserving refactor.
  - the preview asset pre-fetcher calls the shared resolver directly, reads the `X-Gcsim-Asset` status, inlines the returned bytes as a `data:` URI, and sets its `usedFallback` flag when the status is `fallback`.
- **`X-Gcsim-Asset` is the cache-decision signal** (confirmed intended). Because a fallback is served `no-cache` it never enters the edge cache, so a cache hit always reports `edge-hit`, never `fallback` — `status === "fallback"` reliably means "a real asset was missing → do not cache the generated card."
- The `resolveAsset` seam on `SatoriPreviewCard` and the path-enumeration helper stay as the PR has them (path-building lives in the card; the enumerator and the card's rendered `` set must stay in lockstep so no path the card renders is left for Satori to fetch live).
- The 1×1 transparent placeholder constant the PR added is no longer needed — a miss now yields the Worker's `misc/default.png` bytes. Remove it (or justify keeping it).
- `ASSETS_ENDPOINT` becomes unused by the preview path; drop the binding if nothing else references it.

**Acceptance criteria:**
- [ ] The preview render makes **no** outbound `ASSETS_ENDPOINT` fetch; asset bytes come from the Worker's own in-process resolver. Verifiable under `wrangler dev` (no `/api/assets/*` or external asset fetch observed during a render).
- [ ] A card whose every asset resolves renders identically to before and is stored in `caches.default`.
- [ ] A card with ≥1 genuinely-missing asset renders with the Worker's `misc/default.png` placeholder, returns `200 image/png`, and is **not** written to `caches.default` (a subsequent request re-renders); detection is driven by `X-Gcsim-Asset: fallback`, not `resp.ok`.
- [ ] The `/api/assets/*` handler's external behavior is unchanged after the resolver extraction — existing asset serving (R2 hit, source-host fallthrough + R2 write-through, static files, traveler special icons, placeholder headers, edge caching) still works.
- [ ] Assets are fetched in parallel and de-duplicated (unchanged from the PR).
- [ ] The transparent-placeholder constant is removed (misses show `misc/default.png`).
- [ ] Worker builds; no dead references (including `ASSETS_ENDPOINT` if fully removed).

**Out of scope:**
- Changing `SatoriPreviewCard`'s layout, `DEFAULT_ASSET_BASE`, or the supersample/scale behavior.
- Changing the asset handler's *resolution policy* — source hosts, R2 write-through, TTLs, placeholder choice. The refactor only relocates that logic into a shared function; it must not alter it.
- The `?format=svg` debug path's existing behavior.

PR #3006, opened 2026-09-18
Fixed Mona omen extending on attack that breaks the bubble

Matoba004:


Charlie-Zheng:
If you want to do it this way, why can't you just add a OnEnemyHit event subscription for extending Omen instead? Extend on Mona's NA damage. Or add the extension to before the `triggerBubbleBurst` in the burstHook

internal/characters/mona/asc.go
```diff
@@ -108,19 +107,14 @@ func (c *char) astralGlowGainCB(a info.AttackCB) {
 	c.AddStatus(astralGlowKey, 8*60, true)
 }
 
-func (c *char) omenRefreshCB(a info.AttackCB) {
+func (c *char) omenRefreshCB(t info.Enemy) {
```
Charlie-Zheng: If it's not a callback anymore, remove it from the name. Use like `triggerOmenRefresh` or something

internal/characters/mona/burst.go
```diff
@@ -122,11 +122,23 @@ func (c *char) burstHook() {
...
 			return
 		}
+
+		atk := args[1].(*info.AttackEvent)
+		if atk.Info.ActorIndex == c.Index() {
+			switch atk.Info.AttackTag {
+			case attacks.AttackTagNormal:
+			case attacks.AttackTagExtra:
+			default:
+				return
```
Charlie-Zheng: you don't want to return here, that would skip popping the bubble

Charlie-Zheng (changes requested):

Charlie-Zheng (approved):

